Issue #703, opened 2020-11-27
Not all text is translatable


Woseseltops:
Quote from Jaime Burriel:

> I also found some pages that have fields not translated or hardocded, like "Available datasets page" table headers and the first line of the search fields tooltips: 'In many text fields, you can use patterns as follows:'

vanlummelhuizen:
That should be easy to change because all occurrences of that text are in templates (see https://github.com/Signbank/Global-signbank/search?q=%22patterns+as+follows%22&type=code). Just make it a {% trans ... %} thing.

While you are (or I am) at it, perhaps also try to bring it down to one occurrence since they seem exact copies of each other.

Woseseltops:
I think this nicely fits my 'add new interface language' branch for #696 !

Woseseltops:
> While you are (or I am) at it, perhaps also try to bring it down to one occurrence since they seem exact copies of each other.

Orders executed.  They were not 100% identical, solved with a parameter.

vanlummelhuizen:
Thank you. Dismissed ;)

susanodd:
I'm looking to see what issues still need attending to.
This one hasn't been posted to for a while.
Related to this issue, there are some translation files that are apparently missing on lightning.
See the screenshots in #746. (Yes, it's about capitalisation, but while checking things I noticed some things were translated on my local server, but not on Global. Translations that used to show up.)
https://github.com/Signbank/Global-signbank/issues/746#issuecomment-1120913834

susanodd:
There are quite a few of these non-trans texts in the Feedback section.

susanodd:
@Woseseltops how extensive should the translations be, for example something like "Error" ?
Occasionally, there are some very long paragraphs.

susanodd:
The Feedback text has been enclosed in trans directives.
This has been pushed to branch `title_caps`.
Additionally test for the existence of a previously uploaded video Feedback has been added.
This checks that the file also exists in the location:
 https://github.com/Signbank/Global-signbank/blob/cbfe53d2b452a88068a7649e6e436f70ed906228/signbank/settings/server_specific/default.py#L11

susanodd:
I read through the wiki section on translation.
I have resolved most of the fuzzy text in the NL po file.
The four remaining are all "patterns" for plurals.

Woseseltops:
> @Woseseltops how extensive should the translations be, for example something like "Error" ?

Can you give me an example of where you are in doubt, @susanodd ?

[...]

Issue #696, opened 2020-11-23
Create another two interface translations (Hebrew & Arabic)


ocrasborn:
This is of course not 'blocking' in the sense of us not being able to work anymore, but @Woseseltops or @vanlummelhuizen , it'd be good if you could do this before continuing with other things. I'm not quite sure what is required, but we'll need to have another interface element to allow for selecting the new language, the admin needs to display new fields for page translation e.g., and a .po file needs to be generated for Hebrew.
What all is not covered by either 1) pasting hebrew texts in the appropriate box in the Pages on Admin, or 2) the .po file? Is everything included in the .po file by new, or are some things still hard-coded?

susanodd:
The pull-down language setting (with the flags), that's hard coded. And a few other places that explicitly look at the interface language.

susanodd:
Some of the places where that interface language code is used is in determining e.g., what field choice language to show.
And similarly for handshapes.
(I searched in the code on "zh-hans" to see where it's explicitly tested for.)

Woseseltops:
The current flag icons are from an icon set that is no longer available, and thus can no longer be extended. Proposed new icons (UK one already being used by the Catalans): 

![image](https://user-images.githubusercontent.com/1909317/100349379-a27dbe80-2fe8-11eb-9af5-425bff670e75.png)

Woseseltops:
Created a branch for this, added a .po: https://github.com/Signbank/Global-signbank/blob/506156d37dfc3b6f5b8cb40d3d2ce78913fe78cf/conf/locale/he/LC_MESSAGES/django.po

Woseseltops:
To do:
* Adding the new interface language to the settings
* Integrating it in the menu dropdown (@susanodd already located where)

If I'm not mistaken, model-translation should handle the rest of it from here automagically (like an extra field to translate pages, etc). The only place where it does not work is field choices : https://github.com/Signbank/Global-signbank/issues/658

ocrasborn:
Hi @Woseseltops , thanks for taking this up! Could you adapt the permissions so that I can access the .po file?

ocrasborn:
Sorry @Woseseltops , my bad! typo in the rsync command. All good.

Woseseltops:
By chance, I discover today (a day later), that this is not true:

> If I'm not mistaken, model-translation should handle the rest of it from here automagically (like an extra field to translate pages, etc) ...

See #702 .

ocrasborn:
[...]

PR #723, opened 2021-02-05
New interface languages and language picker

Woseseltops:
In this pull request:
* #696 Added two new interface languages: Hebrew and Arabic 
* #696 Wired the language picker to the interface languages in the settings (no longer dynamic) 
* #696 Removed the flags and animations from the language picker
* #703 Unified tooltip html to single location, made it translatable

Points of doubt:
* I added this setting, which populates how the languages are called in the language picker:

``
#Short (abbreviated) version of how language users call their language, to be used in the language picker
INTERFACE_LANGUAGE_SHORT_NAMES = ['EN','NL','官话','עִברִית','عربى'] 
``

A prettier solution, I think, would be to add this to the language objects in the Django admin, as this is where we specify how languages are called in other languages anyway. However, because this would require an extra database migration, which is inconvenient as there are already migrations going on at at least one other branch, I've postponed it for now.
* Adding new interface languages also means that all models that have fields that are influenced by django-model-translation should get extended... I forgot how this worked again, and I guess this is also a database migration?

### DIFF
--- a/conf/locale/ar/LC_MESSAGES/django.po
+++ b/conf/locale/ar/LC_MESSAGES/django.po
@@ -1,0 +1,3121 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+#
+#, fuzzy
+msgid ""
+msgstr ""
+"Project-Id-Version: PACKAGE VERSION\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2020-11-27 13:54+0100\n"
+"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language-Team: LANGUAGE <LL@li.org>\n"
+"Language: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+
+#: signbank/attachments/templates/list.html:33
+msgid "File attachments:"
+msgstr ""
+
+#: signbank/attachments/templates/list.html:36
+msgid "File link"
+msgstr ""
+
+#: signbank/attachments/templates/list.html:36 signbank/pages/admin.py:14
+#: signbank/pages/models.py:10
+msgid "URL"
+msgstr ""
+
+#: signbank/dictionary/admin.py:57
+msgid "number of senses"
+msgstr ""
+
+#: signbank/dictionary/admin.py:70
+msgid "No Senses"
+msgstr ""
+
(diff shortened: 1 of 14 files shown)


vanlummelhuizen:
Thanks @Woseseltops. I have one issue: if you add languages to the LANGUAGES setting, a makemigrations would add those languages to translated model fields. I tried it out and got

```
    - Add field name_ar to language
    - Add field name_he to language
```

The main question here is: do we need to also translate language names to those new languages? If yes, we need to make migrations and use them. If not, we could add a MODELTRANSLATION_LANGUAGES setting that does not have the new languages. 

Making a migration at this point does not pose any problems if you merge from master first. That means you could also add another field to Language.

Woseseltops:
Okay so the answer to my point...

> Adding new interface languages also means that all models that have fields that are influenced by django-model-translation should get extended... I forgot how this worked again, and I guess this is also a database migration?

... is to do a `makemigrations`, which indeed results in a migration and is thus something we should avoid.

> Making a migration at this point does not pose any problems if you merge from master first. That means you could also add another field to Language.

Do you mean you have no migration files in your separate branch, and I can just make migrations as long as I put them on the master branch as fast as possible? Because then maybe we could avoid what I am talking about below:

> The main question here is: do we need to also translate language names to those new languages? If yes, we need to make migrations and use them. If not, we could add a MODELTRANSLATION_LANGUAGES setting that does not have the new languages.

To make things more concrete, this about cases where you for example select Arabic as interface language, and then look at the NGT dataset which has Dutch as a translation language; Signbank will needs to know the Arabic translation of the name 'Dutch', and will try get this via field `name_ar`, which is automatically generated by django_model_translation. Without doing a migration, I expect errors like `No column name_ar for table Dictionary_language` or something like that. 

Are you suggesting we modify the template so that it only shows a translated language name if we say so in a setting, and if not show the name of the language in the 'main' interface language (which is English)?

vanlummelhuizen:
Sorry for the late reply.

> Do you mean you have no migration files in your separate branch, and I can just make migrations as long as I put them on the master branch as fast as possible?

In branch [fieldchoice_field_translations_658_661](https://github.com/Signbank/Global-signbank/tree/fieldchoice_field_translations_658_661) I  do have migrations, but since I do not know when it is finished I believe I cannot block migrations on other branches. Furthermore, I now have some experience in resolving migration conflicts because I have merged two branches that both contained new migrations. As I said earlier somewhere, merging my branch with master is going to be hard, so having a few extra migrations to deal with is ok.

> Are you suggesting we modify the template so that it only shows a translated language name if we say so in a setting, and if not show the name of the language in the 'main' interface language (which is English)?

I am not sure this is a template thing, but I think this is the essence. We are talking about two language uses:

1. Interface language (.mo and .po files)
2. data language governed by django-modeltranslation (now only Language.name, later also FieldChoice.name)

Both of them are now based on the LANGUAGE setting. If we separate them, we need a fallback translation language (English), but we avoid making a migration.

Woseseltops:
Outcome face to face conversation: MODELTRANSLATION_LANGUAGES is an existing option offered by django-modeltranslation which should stop `No column name_ar for table Dictionary_language` from happening. That means no migrations. 

To do:
* Add an extra commit to test the setting.
* Merge.

Issue #697, opened 2020-11-25
FieldChoice with field 'Handshape' and machine_value 2 has many duplicates


vanlummelhuizen:
@ocrasborn @Woseseltops @susanodd Do know why?

ocrasborn:
Don't know how it came about, but remember that it was there, yes. I thought we had already polished it away. Is there perhaps an older issue about it, @vanlummelhuizen ? If it was re-introduced, we should dive into it. If not, then we should check the data, make them consistent, and remove one of the values.

susanodd:
There's a huge amount of them now!!!

![handshapes-machine-value-2](https://user-images.githubusercontent.com/24272544/110940129-b00c4980-8336-11eb-8a46-4a7c888859da.png)

susanodd:
I'm wondering if something is messed up in Admin.
I noticed on the other issue (field_color in admin) that Django calls the get_form multiple times when working on field choices.
I don't understand enough about how the Django automated part of Admin works to know why it is doing this.
It's also using VersionAdmin and putting things in a history.

susanodd:
I'll try deleting all the extra ones on my server and see if they appear again.

susanodd:
I've deleted them all, except one. That leaves the original fieldchoice duplicate machine values. In Handshapes, this is already resolved as there is only one with machine value2.

![handshapes-field-choices-2](https://user-images.githubusercontent.com/24272544/110942032-80ab0c00-8339-11eb-87c3-e282e1ffa958.png)

![handshapes-machine-value](https://user-images.githubusercontent.com/24272544/110942046-86085680-8339-11eb-9d77-e740e4376c66.png)

susanodd:
@ocrasborn @vanlummelhuizen, shall I do this on global? There are 50 duplicates of A > T

vanlummelhuizen:
@susanodd I  am OK with you deleting duplicates on global/live. We always have a backup. @ocrasborn agreed?

ocrasborn:
Agreed @vanlummelhuizen @susanodd

susanodd:
I'm working on this now. In the admin. I am getting intermittent Cannot operate on a closed database errors quite often.
I can do about 3 deletes, then this message appears. Then I wait a while, read issues in github, go back to Dictionary then Field Choices, then start deleting again.

[...]

PR #736, opened 2021-03-31
Issue #697: Fixed Field Choice Test.

susanodd:
Revised tests.py to account for removed duplicate field choices.


vanlummelhuizen:
Thanks @susanodd It is live now (although tests.py does not do anything for the user).

Issue #735, opened 2021-03-30
Other Media files and admin


susanodd:
Improvements to the interface to make management of other media easier for the user.
Followup to #440 

- Gloss/Morpheme Display checks existence of Other Media files on server, user feedback
- Gloss/Morpheme Edit checks filename and file type during Upload, user feedback

susanodd:
This has been pushed to github.

ocrasborn:
Hey @susanodd , I just noticed that for the CSL_Shanghai dataset, all 'main' movies have disappeared, but that these very movies are now available under Other Media. Is this related to this issue? Or something else you guys are working on?

susanodd:
I'm not sure. @vanlummelhuizen has just merged the branch. I don't know if it was live when you posted that or not?

The other media is stored under the gloss id (folder) under (protected_media) othermedia.
The videos are stored under (protected_media) glossvideo under the dataset acronym (folder).
Nothing was moved.

It could be that broken paths are now working. There may have been a path prefix missing?
(I'll have to search to figure it out.)
There is also a folder called import_videos. But that is connected to a batch import. @Woseseltops was working on that. #717

vanlummelhuizen:
> I'm not sure. @vanlummelhuizen has just merged the branch. I don't know if it was live when you posted that or not?

@susanodd I merged your work and put it live after @ocrasborn 's comment.

ocrasborn:
@susanodd @vanlummelhuizen : I looked on the server under glossvideo/, and see that in the "old" (?) CSL_Shanghai/, there's only empty subdirectories, while there's a new folder CSL_Shangh/ that contains all the videos. I could simply clean things up there manually, but I'd best leave that to you, and more importantly: how could this have come about?!

susanodd:
Yes, it's best to check whether there are paths to the old locations inside glosses.
Was the dataset renamed?

susanodd:
I'm looking at the GlossVideo objects (on my local server) to see what the paths look like.
All of the paths with "CSL" in them refer to folder glossvideo/CSL_Shanghai/ ....
I tried looking at one of these glosses (6675), but there are no "Other Media" files for it (nor on the live server).
Have you got a gloss ID for where this happens?

susanodd:
Would it fix it to just rename CSL_Shangh ? It should be CSL_Shanghai 

Maybe #717 did something?

[...]

PR #738, opened 2021-04-01
Issue #735: Other media files Gloss/Morpheme view/upload.

susanodd:
Modifications to Other Media section of Gloss Detail View.
Checks for existence of file and greys out download if something is wrong.
Upload quotes the input filename if necessary and uses that instead of the original filename on the user's computer.
Upload checks that the file type is okay.

### DIFF
--- a/signbank/dictionary/admin.py
+++ b/signbank/dictionary/admin.py
@@ -48,6 +48,7 @@ class RelationInline(admin.TabularInline):
 
 class OtherMediaInline(admin.TabularInline):
     model = OtherMedia
+    readonly_fields = ['path']
     extra = 1
 
 from django.utils.translation import ugettext_lazy as _
--- a/signbank/dictionary/adminviews.py
+++ b/signbank/dictionary/adminviews.py
@@ -1446,7 +1446,7 @@ class GlossDetailView(DetailView):
         other_media_type_choice_list = FieldChoice.objects.filter(field__iexact='OthermediaType')
 
         for other_media in gl.othermedia_set.all():
-            (media_okay, path, other_media_filename) = other_media.get_othermedia_path(gl.id, check_existence=True)
+            media_okay, path, other_media_filename = other_media.get_othermedia_path(gl.id, check_existence=True)
 
             human_value_media_type = machine_value_to_translated_human_value(other_media.type,other_media_type_choice_list,self.request.LANGUAGE_CODE)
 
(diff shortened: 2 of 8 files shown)


susanodd:
Here's a screenshot showing the Other Media.
The greyed out one is not on my local server. It's also disabled.
If you hover, you can see the href of both of the media files, but clicking on the greyed out one does nothing.

other-media-greyed-out-disabled

susanodd:
Here's an example of trying to upload a "csv" file where the extension has been altered (by me) to ".png".
This generates an error.

select-fake-png-file

error-message-fake-media

susanodd:
I searched the database and only found a test pdf file that had been uploaded.
I was concerned that there were so many python files and text files. (These are not on my machine.)
They were at the time uploaded to static. The original paths are in the OtherMedia objects.
That's why everything with bad paths gets greyed out now.

I added the video/* and image/* filter so that the browser limits the choice.
Is it possible to leave this until users request otherwise?

susanodd:
I guess they're all OSError.

susanodd:
I had problems with the os calls sometimes because of special characters in filenames.
I have a couple of sample files on my computer with Farsi characters in the filenames.
So conceivably, every single "os" call could throw an exception?
That's why the code takes apart the original filename to check the format.
The R->L filenames of the original file were causing problems.

susanodd:
Probably some of these need to be in the media folder for tests. Although it's deliberately making things that are supposed to throw exceptions.

susanodd:
I tried out adding an accept for application/pdf files.
It works okay to upload them.

I created a pdf by saving a CSV file as PDF inside of LibreOffice (on Ubuntu).
I don't have any software to create PDF files other than that.

The source file opens okay, so nothing is wrong with the actual file. And if I save the uploaded file it is also okay.
But Firefox doesn't start up a viewer for it, it has to be downloaded.

susanodd:
I completed all of @Woseseltops requests.

Woseseltops (approved):
Awesome @susanodd , thank you for processing my whining ;) . Hereby my seal of approval.

@vanlummelhuizen , you agree we don't need two approvals every time?

vanlummelhuizen:
> you agree we don't need two approvals every time?

@Woseseltops  I agree.

susanodd:
@vanlummelhuizen can you put this live?

vanlummelhuizen:
@susanodd I will. If I cannot find the time today, I will do it first thing tomorrow.

vanlummelhuizen:
@susanodd Done!

Issue #739, opened 2021-04-01
Display refinement to Admin Gloss


susanodd:
There are some large text areas inside of the Gloss Edit of Admin.
I noticed this when checking the Other Media display there to see what paths are shown.
This issue is meant to refine the display there. Some of the code is legacy.

susanodd:
Okay, this is not exactly related to this issue, but it is something weird with a test gloss (on the live server).

Here, the url has a circular path. (@Woseseltops @vanlummelhuizen ):

![weird-url](https://user-images.githubusercontent.com/24272544/114181321-a8de5880-9941-11eb-9f8b-db06a89719cf.png)

susanodd:
I removed all the duplicate stuff at the end of the path and reloaded it. Now it's disappeared from the url.

https://signbank.cls.ru.nl/dictionary/gloss/8822/dictionary/gloss/dictionary/gloss/dictionary/gloss/dictionary/gloss/37052

https://signbank.cls.ru.nl/dictionary/gloss/8822/

susanodd:
THIS ISSUE

Here is a screenshot of the Glosses section of Admin on the live server.
Observe that it looks empty.

![admin-glosses-top-signbank](https://user-images.githubusercontent.com/24272544/114182691-34a4b480-9943-11eb-8c48-e9ccf7aa8ce1.png)

susanodd:
THIS ISSUE STEP TWO

The above prompted me to investigate this on my local server. I modified the display of glosses to also show the ID, the Lemma (the same as above), followed by the is_public toggle.

Here is the same "top" rows, followed by the rows where Lemma's start to be shown.

![admin-glosses-top-local](https://user-images.githubusercontent.com/24272544/114182872-759cc900-9943-11eb-944e-2f18d84cf4f0.png)


![admin- glosses-mid-local](https://user-images.githubusercontent.com/24272544/114182894-7df50400-9943-11eb-8b5a-9bbbb552e222.png)

susanodd:
In the database, there's a rather large number of totally empty glosses. They don't have a Lemma so they're not in a dataset. The database is all but empty for those rows. It's not possible to access them except for in the Admin.

I'm working on making the Glosses Admin less cumbersome. I revised it to show the ID because all of those empty glosses cause problems trying to search for anything.

ocrasborn:
Ah, that'll be very helpful @susanodd ! thanks

susanodd:
[...]

Issue #682, opened 2020-09-10
Rename dataset field for 'use by others'


ocrasborn:
As I'm almost always logged in, I only see now that anonymous users can only see the datasets they can apply for access to (right now: NGT, NGT-ns, EmblemsNL). Only when they have an account and are logged in, they see all datasets, irrespective of whether they can apply for access to them. Thus, they don't even get to the field that was intended for them in the dataset metadata page: "Conditions of use by others". (A misnomer, In fact, hence #683.) In this way, we don't advertise the datasets we have, nor the system as a possible location for new datasets by new users.
Therefore: please make sure that anonymous users can see all dataset pages, whether or not they can currently apply for access when they register, and whether or not they have public glosses in them. The latter is not really related to begin with, but the former does seem to be, there's a setting in the dataset admin that makes the three 'Dutch' datasets above visible: "Is public".
In other words (I'm writing this as I'm finding out what the situation is): we should keep using the existing setting "Is public" for sharing in the sense of allowing users to apply for access, but make the metadata for all datasets available to anonymous users at all times.

susanodd:
Do you want the anonymous user to be able to see how many glosses are in the datasets?

susanodd:
At the moment, the column to apply for access is only available for logged in users.
Make this column visible?
There actually isn't any place to click on anything in the Available Datasets list, only to see the Detail page, but not to try to gain access.

ocrasborn:
Hi @susanodd , both suggestions are worth following up on, please go ahead and implement (quoting you to be sure we're on the same page):
* Maybe for anonymous users there can be a Request Access button on the Dataset Detail page?
Then they also need to create an account.
* Or maybe just a message telling them they need to create an account if the user viewing the details is anonymous.

susanodd:
I'm working on this issue now.
I've made it so there's a button to request access if the user is anonymous.
One glitch, the current registration form only offers the choice of is_public datasets.
So this means that if the user has requested access to a dataset that is not public, it won't be offered.
At the moment, there are only 4 public datasets.
Easy solution is to only show this button for public datasets.

susanodd:
![dataset-detail-ngt-anonymous](https://user-images.githubusercontent.com/24272544/114021950-8da41800-9871-11eb-9288-0bdd0349ba93.png)

![registration-preselect-ngt](https://user-images.githubusercontent.com/24272544/114021974-9399f900-9871-11eb-8b94-1e5cefd7d5f3.png)

[...]

PR #745, opened 2021-04-29
Admin issues 739 682

susanodd:
This branch includes a simple change (#682) which adds a button to the Dataset Detail page for an anonymous user to request access. This button goes to the Registration page, where that Dataset is preselected in the choice of datasets.

The other issue in this branch is #739.
This includes numerous changes to Admin (!) that restrict some of the functionality that was causing extreme slowness (in Admin).
This is primarily in Gloss administration, as shown in #739.
Some additional Inline classes have been filled in because other models were also extremely slow.

LemmaIdgloss now has administration. But this is limited to only adding translations if they are missing for the dataset.

If the dataset is missing on a lemma, this can not (yet) be "fixed" (i.e., the Dataset field, if it is not set, can not be set. This is to prevent an Admin from inadvertently putting a lemma (and glosses) into a dataset. I tried implementing a "guess my dataset" feature based on the translation languages, but this proves too ad hoc..... If this is desired, I will try to come up with something. I noticed there were a number of "loose" objects, e.g., glosses without datasets, that was the original intent of this issue, to make  objects visible that were in the database, but otherwise not accessible.....)

If the dataset is known, it's easier to set translations now (see screenshots in #739). The interface has become more user friendly.

Most of the "select action to apply to selected objects" have been hidden. This is to prevent accidental deletion of objects that are linked to elsewhere, such as Keyword objects. This is to prevent broken links. However, individual objects can still be deleted.

Since the admin is used mainly by @ocrasborn, @Woseseltops, @vanlummelhuizen, and @susanodd, the author of this issue is of the opinion that it's easiest to just "put the new Admin live" and try it out.
It takes way too much time to make screenshots of every change to the interface. Howevewr, some screenshots of LemmaIdgloss and LemmaIdglossTranslation have been added to the issue.

### DIFF
--- a/media/css/custom_admin.css
+++ b/media/css/custom_admin.css
@@ -1,0 +1,18 @@
+.inline-group .tabular tr.has_original td {
+    padding-top: 0px;
+}
+
+.original {
+    display: none;
+}
+
+.delete {
+    text-indent: 10px;
+}
+
+input[type="checkbox"] {
+    outline-style: solid;
+    outline-color: black;
+    outline-width: 1px;
+}
+
--- a/signbank/dictionary/admin.py
+++ b/signbank/dictionary/admin.py
@@ -1,8 +1,10 @@
 from colorfield.fields import ColorWidget
 from django import forms
+from django.forms import TextInput, Textarea, CharField
 from django.contrib import admin
 from django.contrib.auth.admin import UserAdmin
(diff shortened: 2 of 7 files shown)


Woseseltops:
I have planned to review this May 6.

Woseseltops (approved):
Couldn't find anything to whine about

Issue #743, opened 2021-04-19
'Internal Server Error' upon downloading minimal pair CSV


ocrasborn:
When I generate a list of minimal pairs from menu Analysis > Minimal Pairs, this takes some time. Downloading a CSV then takes so long for NGT that I get an 'internal server error'; for Kata Kolok it does download after... 15-30 seconds perhaps. Is there some time-out setting that can easily be changed?

Woseseltops:
If @susanodd didn't look into by then, I will look into this Thursday 29. 

My first steps would be: 
* Figure out where in the code this export is created.
* Figure out what is taking so long.

susanodd:
@Woseseltops, I finished the improvements to Admin and have pushed them to branch admin_issues_739_682

I can look at this issue too now. But you are welcome to work on it!
I recall we had some similar issue a while back, but with importing big files.

Woseseltops:
Thanks @susanodd I indeed already started this morning!

What I figured out is that the problem is in function minimalpairs_focusgloss, in tools.py. This function takes 0.25-0.5 seconds or so, and is called thousands of times in the situation @ocrasborn describes. It takes so long because it takes a gloss and then compares to a (potentially) large collection of other glosses.

I see two approaches to solving this:

The first one is to 'make the weight more bearable', and improve the user experience. We could create an 'export task' with its own ID and detail page that the user can go back to to see if the file is already done and available for download. We could even have the application give updates on its progress, as we know really well how many glosses are already done and how many should be done in total.

The second one is the optimize the whole process, which comes down to decrease the number of glosses to compare to. If I understand things correctly, @susanodd has already done some optimization inside the minimalpairs_objects() function on the gloss object. An optimization I was thinking about is something along the lines of this:

1. Select one field, and divide all glosses into a group where the value is the field is the same as the focus gloss (group A), and a group where it is different (group B). Filtering glosses like this should be fast. I expect group A to be small and group B to be large.
2. All glosses in group B that also has another value for another field can be excluded. So go over each field and filter group B so you only keep the ones that have a similar value. The result should be a really small group.
3. Then really compare the focus gloss to the remaining glosses (group A and small group B) one by one.

What do you think @susanodd ?

susanodd:
The first approach seems more straightforward.

I know the current computation sums up for each other gloss how many fields differ from the focus gloss. It also takes into account whether a field is empty.
The computation is done by a database query constructed in this method:
[...]

Issue #741, opened 2021-04-13
Replace h by H in the name of the tab "Revision history"


ocrasborn:
cf. capitalisation of the tabs Frequency View and Relations View, e.g.
(Yes, how small can an issue be!)

And while your at it, it might pay off to go through the English interface for 15 minutes to spot other inconsistencies (Menu item names?)

susanodd:
This is done. It's in a queue of updates now.

susanodd:
This has been pushed to master branch.

susanodd:
This is live on Global. (Already for quite some time.)

susanodd:
I revised all the English page titles in Admin.

susanodd:
This issue has been subsumed by #746.

susanodd:
The capitalisation has been included in #703.

Issue #733, opened 2021-03-25
Create a view to list and review unused lemmas


Woseseltops:
I reviewed the code but could not find this functionality already. I suggest a list with a 'delete' button behind each item.

Requested by Hannah Lutzenberger

ocrasborn:
Would it be an idea to integrate this in #529, that is, to add a column to an import file that does not match a field in the database but that could be used to mark Lemma ID Glosses for deletion?
In the CSV export of Lemma ID Glosses, it would be helpful then to add a column that lists (comma-separated) the Annotation ID Glosses that *are* linked to a lemma; with that, users could easily search for the unlinked ones.

susanodd:
If you are on the Lemma List view, you see a column with Number of glosses.
If there are no glosses, there is a trash bin in the Update column

hannahlutzenberger:
susanodd, this is what I have been using, but it is quite time-consuming to delete unused Lemmas in this dataset (1636 lemmas on 33 pages). What I was interested in is whether there is a more efficient way to this, i.e. bunch deleting the lemmas with trash bin symbol (or even better, an automatically deleting unused lemmas).

susanodd:
Okay, thanks for the feedback. That is way too many to delete by hand.
Originally, the lemma's were created by default because the field idgloss used to be in the Gloss model. So those got turned into  lemma's behind the scenes. The lemma's also are where the (pointer to the) Dataset resides.
I will look to see what is going on. It looks like things got made obsolete in some way.

hannahlutzenberger:
yes, initially they were created automatically but I am using them to further group sign variants together in the Kata Kolok dataset. This means that I may create new ones and edit existing ones to collect a few variants. It may not be an issue in the NGT dataset  but for me it would be quite helpful! thank you!

susanodd:
I added numerous improvements to Admin LemmaIdgloss and LemmaIdglossTranslation in #739. Those aren't online yet. There are a couple of screenshot examples of Kata Kolok. Although those are focussing on administration of the Lemma objects, not on whether they are used.

I'm working on this issue now.

Woseseltops:
Curious what you are building @susanodd . To give you some inspiration, maybe
* an overview of all unused lemmas with checkboxes before them that are checked by default
* a button 'delete all selected'

susanodd:
So far, I'm adding an additional "form your query" parameter, No Glosses.
Then all the results will have no glosses.
Okay, a column would work.

[...]

Issue #727, opened 2021-02-12
Use field value colours in pull-down menus in the search form


ocrasborn:
So please make it cf. https://signbank.cls.ru.nl/settings/glosslist_colors/ (created for #693 ), but ensure that the enhanced keyboard navigation (#140 ) is kept intact.

susanodd:
I've got this working!!!


![gloss-list-colors-select2-pulldown](https://user-images.githubusercontent.com/24272544/108994451-1a1cc180-769c-11eb-9834-4c6fd9ba2807.png)

susanodd:
Once selected, they're just normal as before.

ocrasborn:
Well done @susanodd ! Now you mention it, I realise we hadn't discussed this yet, but it was my intention that indeed the colours should only be visible when the menu is open, not after a value has been selected and edit mode is off.

susanodd:
Here is the new solution that does not modify the select2 javascript.
It looks a bit cheesy.

![select2-span-height30px](https://user-images.githubusercontent.com/24272544/110102953-a66a6b00-7da5-11eb-8286-c0bddba0049a.png)

susanodd:
This new solution is on github.
Kindly put theoddbranch live @vanlummelhuizen .

Woseseltops:
You are now overriding the select 2 dependency instead of changing it, right @susanodd ? 

Furthermore, I noticed there are still a number of ``<!-- commented lines --> `` in the commits, I guess it's cleaner to remove these?

susanodd:
I originally modified the select2 javascript source code.
The "solution" is now is one of the parameters that are given when the select2 is created.
(It doesn't override anything, it's an optional parameter.)

It's basically the same solution, except it had to be coded in a different way.
I left the commented out console logs in the code, because @vanlummelhuizen didn't like the original solution.
The new solution looks a bit like magic, but I already knew how the select2 javascript works.

I assumed @vanlummelhuizen was going to try out the code on his own server, and it seemed easier to just leave the console logs there so he could examine it.
I'll remove the comments if you want.

ocrasborn:
It works like a charm, but now we still have the odd situation that to see the colours you need to go to the Settings menu instead of to the Signs menu, where users normally select Search. Can't we just discard the Preview Gloss Colors Search page, and instead use the colours in the regular search page, @susanodd ? Or is there another function of the Preview page that I'm missing, and do we simply need to link to that page also from the menu item Signs > Search?

[...]

PR #747, opened 2021-05-12
Lemmas without glosses 733 and minimalpairs to CSV 743

susanodd:
This branch has the following:

- allows deletion of lemma's that have no glosses, along with their related translation objects. This functionality is offered on the Lemma List page. There is a confirm dialog. The deletion only applies to lemma's without glosses that are in the queried results.

- adjusts the order during export of minimal pairs to CSV to account for large files. The minimal pairs are computed before generation of the file is started. Previously, the file was created and then written row by row as the minimal pairs were computed. While the computation time is still slow, 15 minutes for NGT, it does not timeout/crash any more, at least on my local server.

-  includes a bugfix with the links in the results scroll bar

- adds magic to the python virtual environment requirements

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,3 +16,4 @@ sqlparse==0.2.4
 unicodecsv==0.14.1
 python-dateutil==2.8.0
 django-colorfield==0.3.2
+python-magic==0.4.22
--- a/signbank/dictionary/adminviews.py
+++ b/signbank/dictionary/adminviews.py
@@ -5828,7 +5828,9 @@ class LemmaListView(ListView):
         qs = self.get_queryset()
 
         if len(get) == 0:
-            return qs.annotate(num_gloss=Count('gloss'))
+            results = qs.annotate(num_gloss=Count('gloss'))
+            num_gloss_zero_matches = results.filter(num_gloss=0).count()
+            return (results, num_gloss_zero_matches)
 
         only_show_no_glosses = False
         only_show_has_glosses = False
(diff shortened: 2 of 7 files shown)


susanodd:
@vanlummelhuizen @Woseseltops any idea when you'll have time to review?

Woseseltops:
Sorry for the wait @susanodd ! Will try today or else tomorrow.

Woseseltops:
My remarks:

* I got a bit confused by the fact that these seem to be unrelated issues. Or is there some relation, @susanodd ? If unrelated, not the end of the world of course, but I would say it makes more sense to only group issues that are about related additions/fixes, and create smaller pull requests for the rest. Besides clarity for the reviewer, an advantage would be that one pull request (and thus feature) can already be deployed while the other is revised.

* For #733 the code seems fine (I actually learned about annotating datasets, thanks!), but I have done a few small extra functional suggestions in the issue.

* For #735 I think you forgot to include the main commit ;) . I have asked a question out of curiosity there.

* For #743, some more UI/UX is required, but that shouldn't stop us from putting this live already.

susanodd:
Bullet 2 (#733) has been completed! Good idea to add the count to be deleted!

Bullet 1 is mainly because I work on a number of issues and wait until the previous merge request has been completed before creating a new branch for the new issues. (I don't like to work on an older version because I get used to working with all the changes on my branch. It all gets tested more that way, too.)
It would be easier if I could keep spanning new branches from the branch I'm still working on.

Bullet 3 (#735) is already on the server. That's how @ocrasborn got the error message about Magic, because I'd forgot to put that in the requirements file on the previous branch where this issue was fixed. The requirements file is included here.

susanodd:
@Woseseltops regarding creating branches for issues...
Since requesting review on this branch, I've completed two issues involving colours in pull-down menus (#726 and #727), morpheme tags (#748), and the H in revision history (#741).
The H doesn't have anything to do with the others. And the morpheme tags issue was encountered while working on the colour pull-downs. The Morpheme View has also been modified a bit qua formatting.
My changes are all still on this branch, because I'm waiting for it to be merged before creating a new branch for these issues which are all ready to go.

Woseseltops:
Deployed!

PR #750, opened 2021-05-31
Colours added to Search Forms and Edit Views

susanodd:
This branch adds colours to the pull-down menus of:

- search forms in the Gloss and Morpheme List Views
- edit mode of Gloss and Morpheme Detail Views

In the Search form, this is done by way of a filter applied to the Select2 multiple select fields javascript. Extra colour information has been provided in the context parameters. This code was first prototyped as the Settings->Preview Gloss Colors Search page. It's been extended to the Morpheme Search page.

In the Edit form, this is done via extra functionality in the custom editable code for the various input types, together with custom Ajax calls. The code also accommodates cursor input during editing. This code was first protptyped in the Gloss Detail Preview page. It's been extended for morphemes.

Some formatting of the Morpheme Detail View has been revised since it had not been edited in a while.


Woseseltops (commented):
My review:
* In signbank/dictionary/adminviews.py, is there a special reason not to do the imports at the top?
* You mention AJAX, but I don't really understand (or see in the code) its relation to the dropdown colors, could you explain that to me?

susanodd:
Regarding the imports, no there isn't a reason for not putting the import statements at the top. Previoulsy, @vanlummelhuizen has a different programming style, where the imports should have a smaller scope if they're only used in a smaller piece of code. If you prefer them at the top, I'll put them there instead.

The AJAX is handled in the editable and gloss_edit javascript files. In the previous related branch with colours, you had feedback asking whether the existing library for editable could be used as has been done for select2.
So I put that comment in the merge request to highlight that the AJAX calls are different. There are e.g. additional parameters to accommodate the colours. Would you like this written out?

vanlummelhuizen:
> Regarding the imports, no there isn't a reason for not putting the import statements at the top. Previoulsy, @vanlummelhuizen has a different programming style, where the imports should have a smaller scope if they're only used in a smaller piece of code. If you prefer them at the top, I'll put them there instead.

To make clear: I also prefer imports at the top. If you got the impression I rather put them elsewhere it probably is either a mistake on my  side or I tried to solve a circular import problem.

susanodd:
It was something related to the guardian permissions. I'll move them to the top.

susanodd:
It's only regarding imports in this branch, right?
There are actually a number of imports not at the top not introduced in this branch.
I removed the four import json's in adminviews.py and put it one time at the top.

susanodd:
About the ajax calls inside jquery.jeditable.mini_colors.is,
This has some coding in order to determine what the "table row id" was of the thing the user is editing.
In the original table of phonology fields, the rows are identified by the field name, which is later used in order to make an ajax call to update that field.
For the colours, an additional layer of "stuff" has been added in order to display the (dynamically generated) pull-down lists with colours. These are generated in this javascript code. They are attached to a (dynamically generated) button with a (dynamically generated) form that when you click on this, you see the list. (In the code this is now a list instead of a select, in order to add the colours, That's also how select2 works.)
Because of this extra "layer" of (dynamically generated) html code, the actual "field name" (which is found in the table row identifiers) has to be retrieved from a different parent in order to know what ajax call should be made as well as store the previous value, and to be able to undo it if the user cancels.

I'm describing what the code does.
It's actually not that easy to read. It was developed over a long time with lots of console.logs to determine how to obtain this information and provide it to ajax.
Additionally the colours are fed in via parameters that get used when creating the pull-downs in this code.
All of this is why the code was originally used only in the Settings menu.

susanodd:
Any more feedback on this pull request?

vanlummelhuizen:
@susanodd 

The imports in

https://github.com/Signbank/Global-signbank/blob/392608433181de599beac650134f85fea7a6116c/signbank/dictionary/adminviews.py#L351

and

https://github.com/Signbank/Global-signbank/blob/392608433181de599beac650134f85fea7a6116c/signbank/dictionary/adminviews.py#L2027

 are they there (not at the top, in a for loop) for a reason?

susanodd:
Those were already like that before this branch,

susanodd:
Probably nobody noticed before. I'll move them.

vanlummelhuizen:
@susanodd When testing your work on Firefox or Chrome om Ubuntu 20.04 I noticed there are some things that are quite different from how it is now. See screenshot below.

1. The small forms (with drop-down menu,  OK and Cancel button) do not disappear if you click a value of another field.
2. The part that drops down is positioned left to the value instead of above or below.
3. The part that drops down expands the table row instead of being in front of the table.

Is all that intentional?

Screenshot 2021-06-17 at 13-00-33 Sign for wesseltest6

susanodd:
Yes, that is what it looks like.

susanodd:
It's a bit weird that they are still showing, though. Did you not click on OK?

susanodd:
The default is Cancel so you don't actually change things. But it's true it's remembering (still displaying) what you chose until you OK it.

susanodd:
@vanlummelhuizen what are you suggesting with bullet 1?

It's basically the behaviour "click someplace else without clicking on OK or Cancel".
But which is it then? The user was busy editing the field and chosen something. Then went on to do something else without clicking OK or Cancel. The behaviour is ambiguous.

The code also takes care of using Return to keep going to the next field and not doing anything to the existing value.
That was in the Issue about Keyboard Navigation to speed up editing of Phonology. So the Colours need to also accommodate that issue of using Return.
Return advances to the next field,.

You complained in the past that there were too many Ajax calls.
It's always been the case that the Ajax calls are done per field. They are not done when you close Edit.
We don't have any facility for Undo. So Cancel works like an Undo and does not do anything. But it leaves the Preview of the thing you selected. If you actually hit Cancel, you will see the Original value again.

I gather you do not like this interface.
But it's the same as it was in the Gloss Preview under the Settings.

Supoose the User clicks away (as you have done by opening another field)
What is the behaviour then?
Has the User Cancelled or Committed ?


I'm not sure what you intend with the other bullet points.
Regarding the layout. I was so far unable to "position" the dynamic things any other way.
I introduced extra classes to tag the dynamic things (which keep disappearing) in order to add some formatting.
It conflicts with some of the Bootstrap things defined for other structures.
The Select2 Library actually puts all the stuff it generates dynamically at the end of the document rather than inline.
I tried doing that as well, but the Field has to be identified in the outer context in order to put the "new value" in the right place in the table.
The way Select2 works, it generates numerous extra identifiers with codes on them and then links these up to all the javascript at the end of the file.
Another problem with that approach is that the layout starts inheriting layout of "main" because that is the parent then.

Can the particular layout positioning be a new issue?

Woseseltops:
Hey @susanodd , I think you are reading too much into the comment by @vanlummelhuizen ;). I think he just means: hey, it notice this is different than it was before, bug or feature? 

To keep things simple, let's focus on the first bullet point first. Right now, I see in the screenshot multiple little 'okay' buttons, whereas I get only 1 at the same time in the live version. Do you know why this is?

susanodd:
Yes, more than one OK buttons are visible because the user opened more field lists without clicking on OK or Cancel on the other ones.

susanodd:
This is bit simmilar to the colour of the buttons in the Lemma List, what the default should be.

For the Gloss Edit, using the keystroke navigation, hitting Return advances to the next field, without doing anything to the current field. So the interpretation of Return is do nothing.
So now, actually the default should become Cancel if a user has opened a list and navigated within the list, but not done an OK.

At present, the code does nothing with this because the user has not clicked on OK and not clicked on Cancel. That is why the fields stay showing Cancel and OK. There has not been an Ajax call done yet.
The old version of the code did the Ajax call already.
But this caused problems with keyboard navigation because doing Return should do nothing and merely go to the next field.

susanodd:
I pushed the `morpheme_edit_color.js` file, which didn't exist before. It is needed in Morpheme Edit View and I forgot it before.

I added a function to hide an open field form if the user opens a different field but didn't close the previous one.

This is done by a new routine `hide_other_forms` in gloss_edit.js.

The scope of "going out of focus" is beyond the form the user was inside of, and beyond the scope of the table row of the particular field. This function is outside of the editable code.

Woseseltops:
Okay I guess the answer to my question here...

> To keep things simple, let's focus on the first bullet point first. Right now, I see in the screenshot multiple little 'okay' buttons, whereas I get only 1 at the same time in the live version. Do you know why this is? In other words: what changed so you can now have multiple 'okay' buttons simultaneously?

... is that this was on purpose, so it better aligns with keyboard controls.

Onto the next bullet:

> The part that drops down is positioned left to the value instead of above or below.

Any special reason for this?

susanodd:
The formatting is a bit difficult. For the Glosses, there is a (mainly invisible) extra column for when the gloss has attrbutes. (Very few glosses have extra stuff above them with plus signs. But there is extra code in an extra column of the panel that is mostly empty, but the code is still there.) That causes alignment problems because there is something that might be there.
I had to make it so the stuff would "stay together" because it was ending up breaking up the "field choice list form", because it is leaving space at the right for something. (That isn't there most of the time.)
I was not able to make this be elegant. (The Morpheme View puts an empty column there because it was making the thing really wide otherwise.)

I fixed it so there are no longer multiple okay buttons. The code erases other OK buttons before it opens a new field.
This is done in gloss_edit_colors.js
https://github.com/Signbank/Global-signbank/blob/f10ee70b9bf5d2517c57ca7697d6eb217eabad86/media/js/gloss_edit_color.js#L499-L517

Woseseltops:
Sorry I didn't manage to look into this today, got this planned for next Thursday!

Woseseltops:
So @susanodd would you say the reason for

> The part that drops down is positioned left to the value instead of above or below.

is that this is not on purpose, but you could not get it right? In that case, we can maybe put it live, and I can take a look?

Woseseltops:
And then finally the last question:

> The part that drops down expands the table row instead of being in front of the table.

Any special reason for this?

susanodd:
Yes, the last two questions are more or less formatting problems.
The javascript code needs to construct things dynamically. I added at least one class name to induce formatting.
But because the code is nested inside other things on the page, it ends up inheriting formatiing sometimes, as well as from Bootstrap. I couldn't get the formatting to work all the time.
I suspect that some Bootstrap or inherited formatting needs to be overridden.
It would be great if you could put it live and take a look.

susanodd:
This has been revised so can be merged with the master.

susanodd:
https://github.com/Signbank/Global-signbank/pull/750#issuecomment-880615348



> And then finally the last question:
> 
> > The part that drops down expands the table row instead of being in front of the table.
> 
> Any special reason for this?


This is primarily to do with the width of the column changing when the new pull-down list is shown.

There is a word wrap in the column cell (in the original template). Depending on the width of the individual choices in the list, the content of the cell (that includes the pull-down) is sometimes wrapped so the buttons appear underneath it (in the original template).

This was causing weirdness in the new template because the buttons that get "wrapped" (because the choices are wide) end up interfering with the display of the new pull-down list. Hence it was chosen to force the buttons in the cell to be placed next to each other in the new template, instead of wrapping.

susanodd:
It's not possible for me to make a screenshot of the original pull-down when open. The browser interferes with this and the list keeps being closed in the screenshot.

susanodd:
I fixed this!!!!!
I put in quite a few console logs in the javascript file to figure out what the context is on the various parts of the drop-down.
I moved the code that creates the click button for the drop-down of a particular row to the "inside" of the dynamic form element.
I just figured out how to hide that button when the drop-down is displayed. It's working now and looks more like the original.

I need to check the code a bit before pushing it to make sure the whole thing works.

susanodd:
@Woseseltops  here is the improved drop-down with colours:

improved_drop_down_footsteps

Woseseltops:
I started reviewing this code today. I see lots of comments everywhere, nice!

Some things which can possibly be improved:
* I notice you added a whole new file called `morpheme_edit_color` here, which I guess is edited from `gloss_edit_color`. For the sake of the DRY principle, would you think it's possible to merge the two into one?
* Instead of this

```python
                field_kind = fieldname_to_kind(field)
                # take care of different representations of empty text in database
                if field_kind == 'text' and (machine_value == '-' or machine_value == ' ' or machine_value == '------' or machine_value == '' or machine_value is None):
                    # print('gloss detail text field ', field, ' machine value set to empty: ', machine_value)
                    human_value = ''
```
perhaps more efficient:
```python
                # take care of different representations of empty text in database
                if fieldname_to_kind(field) == 'text' and machine_value in ['-',' ','------','',None]:
                    human_value = ''
                else:
                    human_value = machine_value
```

susanodd:
It looks like the use of a separate file for morphemes was motivated by a variable: gloss_phonology versus morpheme_phonology, used in the respective js files.
The morpheme_phonology is obtained from fields in the setting MORPHEME_DISPLAY_FIELDS
whereas that of gloss_phonology is computed in the template context. For glosses, there are extra fields for HANDSHAPE_ETYMOLOGY_FIELDS and HANDEDNESS_ARTICULATION_FIELDS. These are handled differently in the template.

Once upon a time @vanlummelhuizen wanted the js not to include lots of code that wasn't relevant. There were even versions of templates that checked permissions before including javascript.
(But that led to other problems.)

I have newer versions of the code that haven't been pushed yet. I revised the update of the weakdrop and weakprop fields. (We use None, True, and False as three different values. The interface was inadvertently setting the field to the wrong value if you just hit Return without selecting a value from the list. Instead of using the original value it was setting it to None. I fixed this in the Ajax calls and update.py. So now return without selecting just leaves the original value. This is relevant because this is the "auto advance to next phonology field" when the user hits return.)

Now that you pointed out the extra morpheme_edit file, I need to check it's still working properly, although it doesn't include these fields.

susanodd:
This is good to go.

Woseseltops:
While I still suspect there might be a cleaner solution for the morpheme related Javascript, let's not let that delay this pull request even further :)

PR #758, opened 2021-06-23
Issue #743: Efficiency revisions to minimal pairs computation.

susanodd:
This branch makes major revisions to the computation of minimal pairs to improve efficiency.


Woseseltops:
I have a question about this before I accept; I'll post it in the issue itself so we don't lose it.

Woseseltops:
Took some time for me to understand what exactly @susanodd had optimized (see discussion in #743 ), but it turned out the bottleneck was in figuring out what exactly was the minimal part in earlier detected minimal pairs :).

Issue #394, opened 2017-12-22
ASL SB: The image disappears when you alter the Lemma Gloss field


ocrasborn:
* Is that a freaky effect or what?! When I change it back, it appears again. Remember that file names are now based partly on the Lemma ID Gloss field --> My diagnosis: the picture file name does not appear to be changed when the Lemma ID Gloss field is changed.
* For the movie, no such problem (I think, but then the movie didn't display in the first place, and it seems like you're working on this as I type, @Woseseltops !)
* But, possibly related: When I edit a lemma gloss in my new test sign, the edit field that appears shows a lot of leading and trailing spaces around the gloss. Not so for the Annotation ID Gloss field.

Woseseltops:
The extra problem caused by trailing spaces is what made this problem really messy confusing for me; when I fixed that (simply a matter of removing them from the template, fortunately), I could finally see what was happening... which turned out to be exactly what you describe: the picture name does not change when you edit the lemma field. Turned out to be related to file extensions, which were not remembered correctly. I think I fixed it now!

PS
We also had both problems (trailing space, picture names not changing) in the NGT Signbank, so I will also fixed this on the main branch.

Woseseltops:
Report from Julie:

> Image sometimes disappears when changing lemma ID but not always. It used to be all the time but now it's unpredictable.

> Image disappears when changing lemma - continuing issue - when I edit ASB entries for suggested signs, if I change the lemma, the uploaded image disappears. Changing the annotation ID gloss doesn’t affect anything though.

And because Julie only does high quality bug reports, there is a video of the problem as well: https://www.dropbox.com/s/alw0her9c4t70nb/FileMissingAfterChangingLemma.mov?dl=0__;!!HJOPV4FYYWzcc1jazlU!ozRBLNZUDAz2udQnCMAEwAL-emYDrSAnwIQ4Q3163pBv3qzRtt7I3vnZq9Lpu8pT5Q$

susanodd:
The video helps a lot.
Related to this I think: #763. I added a comment about a setting in the video/models.py file.
In the video, it looks like the small "d" typo at the start could be causing the implementation to look for or create or move files to a new folder with the first two characters.
The video shows the initial characters Annotation as "~D".

susanodd:
I tried doing this on my local server.
I have two glosses matching DINO for ASL (on Global there is also a dataset ASL, so I'm trying it there with the Global code.)
[...]

PR #772, opened 2021-09-23
Move files on disk setting

susanodd:
This setting needed to be false for some reason.
When true it was giving errors during testing for the new corpus creation functionality.

### DIFF
--- a/signbank/video/models.py
+++ b/signbank/video/models.py
@@ -377,7 +377,7 @@ class GlossVideo(models.Model):
         # code has been introduced elsewhere to make sure paths are the correct encoding
         return self.videofile.name
 
-    def move_video(self, move_files_on_disk=True):
+    def move_video(self, move_files_on_disk=False):
         """
         Calculates the new path, moves the video file to the new path and updates the videofile field
         :return: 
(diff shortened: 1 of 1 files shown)


susanodd:
This might be related to #394.
Suppose you alter the Annotation or Lemma ID Gloss.
Before doing this, the video file was stored under the Dataset acronym folder, in another folder with the first two characters of the text.
By altering the Annotation or Lemma ID Gloss, this setting would seek to move the video into a folder that matches the new text.
I suspected something like this was happening, causing the videos or images to disappear. That's why I set it to False.

susanodd:
The video files, their paths, should be (being) stored in the database. At one point we started looking for things on disk rather than looking up the paths in the database.

susanodd:
Okay, here's a place in the original code where move_files_on_disk **IS** set to false.

https://github.com/Signbank/Global-signbank/blob/94dbe93ee7552b7d9cd19416ee29b86d5d18dd61/signbank/video/models.py#L422-L438

susanodd:
Okay, I seem to have fixed it for tests.
I added this table to the deleted tables list in create_test_db.py

video_glossvideo

Woseseltops:
Hey @susanodd , I vagely remember that this pull request could be closed... is my memory correct?

susanodd:
Yes, this branch is no longer relevant. You fixed the problem by installing a package that was missing.

Issue #764, opened 2021-07-22
Add another relation between signs


Woseseltops:
Requested by Julie:

> Add a new relation under "relations to signs" - our current "see also" is overused so I'd like to add one relationship in particular "handshape paradigm" which we will use to capture relationships between signs where they have the same phonological form but different handshapes (e.g., numeral incorporation sets, initialized sets eg GROUP, CLASS, FAMILY, etc).

Should be a small task.

@ocrasborn , do we want this for the Global Signbank as well?

susanodd:
The relation choices are defined by a constant in models.py

https://github.com/Signbank/Global-signbank/blob/94dbe93ee7552b7d9cd19416ee29b86d5d18dd61/signbank/dictionary/models.py#L1723-L1730

susanodd:
Related to this, I've made improvements to Gloss Relations View.

ocrasborn:
@susanodd , yes, let's add this for Global SB as well.

susanodd:
Okay, on the live server something got messed up with the display of minimal pairs in the Related Signs View. The contrasting field values are not displayed anymore.
This could have something to do with the space of time between this branch first being created and the numerous updates via other branches in the between time.

susanodd:
@Woseseltops  is there any reason why on MY SERVER the minimal pairs show up correctly, but on the live server they do not? (Compare the screenshots below)

https://signbank.cls.ru.nl/dictionary/gloss_relations/349

(screenshot)

![relations_view_349_live_server](https://user-images.githubusercontent.com/24272544/151545143-740cb44a-7c35-43a3-b59c-8d96c691e43b.png)


This is the same page on my server:

![relations_view_349](https://user-images.githubusercontent.com/24272544/151544909-5d792fec-9390-4378-a40d-55bb5b47c199.png)

susanodd:
@Woseseltops 
I'm on branch relations_between_signs for this issue.
I merged it with the most recent master.
It's only one commit ahead of the master, (code for adding the new relation).

I'm not able to fix the issue of the minimal pairs (screenshot) on the live server because it doesn't show up on my server!
Is there some file missing on the live server?

susanodd:
@Woseseltops , it's fixed now! Thanks!!

susanodd:
I added the code for displaying the new Handshape Paradigm relation to the Related Signs View.

Here's a silly example.


[...]

PR #773, opened 2021-09-27
Related Signs View: Fixed issues with pattern variants.

susanodd:
I fixed some issues with pattern variants showing up for some glosses.
I also removed the hard coded languages in the template.


susanodd:
There were some hard-coded languages in the Relations View.
While repairing that with an additional setting, I noticed the glitch in the original code was also used in Gloss Detail View, so I repaired that as well.

susanodd:
There was a glitch appearing with variants.
It's possible to store a variant relation explicitly.
It's also possible to let them be found using patterns.
The (new and improved) Gloss Frequency View also shows variants.
In testing that code, I looked for glosses with variants.
It became evident that the handling of variants was inconsistent.
I revised the code to be more robust. Modifications were made to the the "pattern" search.
The old code was exposing BEER-A to be a variant of BIER-A.
(Via the English Annotation, instead of only looking at the same language.)

Woseseltops:
I am a bit confused here. This pull request refers to an issue which is about simply adding another hardcoded relation type, but that commit is not part of this pull request... instead, I believe this is about things for which there are no issues:
* A small fix for pattern variants
* Moving things related to hardcoded translation languages to the view (instead of the template)

 @susanodd , is this correct?

About the last thing, you replaced all of the translated field choices with `get_FIELDCHOICE_display`, like `get_handedness_display`, which seems super elegant to me... but how does it work? Where can I find the code that takes as input a field choice and an interface language, and then outputs the correct human value?

Furthermore, reviewing the code, I noticed you are using `x` as variable name a few times in for loops... I guess these are a good candidate for some quick readability improvement?

susanodd:
The issue that adds a new relation type needs this issue to be committed because of the template Relations View.
I completely rewrote that code for that template. But it's not live yet.
The issue that adds a new relation type needs the newest template.
The "live" signbank template Relations View has hard coded languages in it.

susanodd:
The fix isn't that small. It repairs a bug.

susanodd:
This branch is ready to merge with the master.

susanodd:
@Woseseltops Branch relations_between_signs (for #764) is waiting on this branch to be merged.
This branch makes changes to the GlossRelationsDetailView template and removes the hard coded languages.

susanodd:
Brought up to date with most recent master.

Woseseltops:
Hey @susanodd , did another review of this code today. Overall I'm happy, but as nearly always I have some stuff to complain about :)

 My feedback:

```python
        if len(queries) > 1:
            query = queries.pop()
            for q in queries:
                query |= q
        else:
            try:
                query = queries[0]
            except:
                query = []
```

This part strikes me as confusing... It tries to transform something inside `queries` to `query`, but I don't understand what the end result is or why and why you use `pop` the first time, but not afterwards. I think we need at least (1) a comment explanation, (2) more descriptive variable names and (3) something to follow `except`. In other words, there must be a reason why you can't do it simply like this

```python

merged_single_query = []
for q in queries:
    merged_single_query |= queries
```

and the reader should be able understand right away.

I think the code directly below that can be cleaner:

```python
        other_relations_of_sign = self.other_relations()
        other_relation_objects = [x.target.id for x in other_relations_of_sign]

        if queries:
            pattern_variants = Gloss.objects.filter(query).exclude(id__in=other_relation_objects).distinct()
        else:
            pattern_variants = [ self ]
```

```python
        if queries:
           #Run the query, but exclude all glosses that have a relation to this gloss
            related_gloss_ids = [relation.target.id for relation in self.other_relations()]
            pattern_variants = Gloss.objects.filter(query).exclude(id__in=related_gloss_ids).distinct()
        else:
            pattern_variants = [ self ]
```

Other than that it's go my thumbs up :)

susanodd:
That's a good question about the queries syntax. (First comments above.) I will rewrite it. It may have been written as so because of types.
In other issues, I discovered during testing (a different issue) that I needed to explicitly create an empty query result rather than an empty list. I suspect it only happens during run-time.
In some of the templates, it's only checked whether something has a value, the variable or function result is non-empty.
That's probably how type errors would arise because the template is only checking whether the thing is non-empty and ignores it if not. At the very least, there should be three cases, >1, 1, 0. So it's very weird that I put the exception in there.
Probably it was generating a type error on None or [] instead of empty query.

For the second statement, I know I am paranoid about things not being evaluated right away.
That's the main reason I put things in separate assignments, to force (strict) evaluation rather than ending up with a giant database expression. I was under the impression that statements like Gloss.objects..... creates a database query.

susanodd:
It does not seem to work with the alternative deconstruction.
The Q elements are different than normal list elements.
the "|" operator needs a non-empty Q element to start.

susanodd:
It looks slightly different but more or less as you suggested, @Woseseltops !

Woseseltops:
Great, thanks! Could you please:
* Also add 1 sentence comment explaining why 

> That's the main reason I put things in separate assignments, to force (strict) evaluation rather than ending up with a giant database expression. I was under the impression that statements like Gloss.objects..... creates a database query.

You might very well be right! Unless you are 100% sure something won't work, my preference would be to try the pretty solution first, and only make it uglier when it turns out there's a bug or performance problem.

susanodd:
Okay, I see what's a bit weird. The merged query expression used to operate on the list variable "queries" which includes patterns to search for for variants. The list constructed now contains at least three patterns. Originally, this only contained one pattern so the subsequent code originally tested whether it was empty or not.
The pop is used to remove one of the Q patterns from this list. Then the rest are iterated over with the for loop.
Except after popping, a Q element has been removed. So the subsequent "if queries" test is looking at the queries expression with that popped element removed. Although this doesn't matter for the code, it's only concerned with whether there is a Q expression to use as a filter. Yes, there is actually a type problem in this code depending on which branch is followed!
Good idea to include a comment. The code probably needs a different constructor for the empty case.

Issue #756, opened 2021-06-16
Frequency functionality for datasets


susanodd:
At the moment, frequency is only available for the NGT dataset. Expand this functionality to other datasets.

This extends #354, which focused on display of frequency data, once the data is already available. That issue was opened in 2017, before datasets existed. It had a slightly different understanding of datasets, for example gender, regions, and ages.

susanodd:
@ocrasborn, what shall the naming convention be for corpora vs datasets?
At the moment, there is exactly one Corpus object:
### id,       name,        description,        speakers_are_cross_referenced
`1,        'NGT',          'Corpus NGT',      True`

The EAF files are Document objects that link to this Corpus via a corpus_id field.

On the Dataset Manager page, the user can upload EAF files for a dataset.
Shall the Dataset acronym be used as the Corpus name?

susanodd:
Here is a first go at a Dataset Corpus Frequency Overview.
This shows that of NGT.
I put the meta file and the EAF files (scroll panel) there that are uploaded on the Manage page.

Then the Corpus (it's just the name; all the other objects associated with it are linked to it).

Then the Speakers (in a scroll panel). These have been retrieved from the GlossFrequency objects that are stored for the Corpus, its Documents, and Speakers objects. [This is what is actually stored shown in this panel; it has not been retrieved from the metadata file.]

So the idea would be to display an overview and make handles available for other datasets that have EAF files and metadata.
Obviously, there needs to be a button to generate speakers and parse the EAF files (during the night).


![dataset-corpus-overview](https://user-images.githubusercontent.com/24272544/123259369-ad53ef80-d4f4-11eb-83a4-e868487dbf4c.png)

ocrasborn:
@susanodd , yes, using the Dataset acronym for corpus name is fine.

Note that for Signbank's purposes, we don't really care whether the uploaded EAF files constitute a specific corpus with a specific identifier; it could be they combine EAF files from multiple datasets, or adult and child data, or whatever. It's up to them to spell that out in the description of their dataset in Signbank. We leave that up to the dataset managers. They upload EAF files, and Signbank calculates frequencies of glosses in those files.

susanodd:
Okay, I'll add that functionality. The Document objects refer to a Corpus object.
I just checked, and that does not preclude that a particular EAF file may be uploaded to more than one dataset.

susanodd:
Would you like separate functionality for documents?
This to avoid uploading them multiple times.

[...]

Issue #354, opened 2017-07-13
Allow for frequencies from different data sets


ocrasborn:
Both for NGT and for other sign languages, we have multiple data sets from which frequencies can be extracted. Can we somehow make the no. of frequency fields (and searching for them) more dynamic, based on what a script generates (and of course we'll also need to adapt the script).

vanlummelhuizen:
Frequencies are now taken from the Corpus NGT. How are datasets identified? How will the CNGT datasets correspond with Signbank datasets?

ocrasborn:
Hadn't thought about it much yet. Perhaps by directory name at a certain level, making sure there's one corpus per repository? Multiple repositories could exist per language/data set, and the repositories would be under a directory that has  the language/data set name as the parent level.

ocrasborn:
We talked about this in our demo today:
* We would also need to provide a way for users to upload EAF files (and upload updates)
* We would need a standard format for metadata that links the file name of the EAF file to a region (CSV file; column 1 EAF file name minus extension; column 2 Region of Signer 1; column 2 Region of Signer 2; etc. if there are more signers). Signer identities can then be extracted from the Participant tag of the EAF files. In this way we also force annotators to use the Corpus NGT conventions for glossing multiple people: each on a different tier.
* We would need to explain to users that Signbank will extract gloss frequencies from all tiers that have the type 'gloss' (so force them to adopt to the Corpus NGT conventions in this sense; which I think is a good idea anyway)
* something more, @vanlummelhuizen ?

ocrasborn:
Addition to what we discussed: let's not be specific about what the metadata categories are (using just region), but let us extract the relevant feature from the metadata. If users find age groups more important than regions, let them add a feature for that. Instead of having a column "Region" and letting the values in that column determine the field names, use any text from a column Category to determine field names. So a table could look like this:

<table> <tr> <th>Session</th> <th>Category</th> </tr> <tr> <td>0001</td> <td>Region: Washington</td> </tr> <tr> <td>0002</td> <td>Region: Philly</td> </tr><tr> <td>0001</td> <td>Register: narrative</td> </tr> <tr> <td>0002</td> <td>Register: discussion</td> </tr></table>

Woseseltops:
A suggestion on how to do this. 

Back end storage
-----------------------
[...]

PR #774, opened 2021-09-27
Upgraded Frequency View Display

susanodd:
I upgraded the graphics display code from chartjs v. 2.9.3 to v. 3.5.1
This required additional support in the python code.
This code uses the GlossFrequency objects as well as Regions based on the Corpus.
This provides support for the new corpora functionality for datasets other than CNGT.


susanodd:
There is also a page:
dictionary/lemma_frequency/<gloss_id>/
This was already on the site. I also upgraded it to use chartjs v. 3.5.1

Woseseltops:
I'm currently reviewing this. To do:
* Read and answer questions in #354 
* Read and answer questions in #756 
* Review code in e573657

Woseseltops:
Hi @susanodd, I finally found the time to review this code. I have a bit of hard time to make sense of what I'm looking at... could you perhaps help me out here:
* Put in some more comments. This not only explains the code, but also puts in bite sized chunks (paragraphs, if you will ;) )
* Have more descriptive variable names (for example, in for loops)
* Remove commented out code that is no longer necessary OR at least put in an extra comment why it is there

For example, take this piece of code:

```python
    def data_datasets(self):
        # for use in displaying frequency for a gloss in chart format in the frequency templates
        # the frequency fields are put into a dictionary structure

        total_occuurences = 0
        data_datasets = []

        try:
            frequency_regions = self.lemma.dataset.frequency_regions()
        except (ObjectDoesNotExist, AttributeError):
            return (total_occuurences, data_datasets)

        gf_objects = GlossFrequency.objects.filter(gloss=self)

        signers_per_region = {}
        frequency_per_region = {}
        gfs_per_region = {}
        for r in frequency_regions:
            speaker_location = gf_objects.filter(speaker__location=r)
            gfs_per_region[r] = speaker_location
            signers_per_region[r] = speaker_location.count()
            frequency_per_region[r] = 0
            for sl in speaker_location:
                # print('r sl: ', r, sl.__dict__)
                frequency_per_region[r] += sl.frequency
        speakers_per_region = {}
        for r in frequency_regions:
            speakers_per_region[r] = []
            for gfo in gfs_per_region[r]:
                gfo_speaker = gfo.speaker.identifier
                if gfo_speaker not in speakers_per_region[r]:
                    speakers_per_region[r].append(gfo_speaker)
                # print('region: ', r, ' frequency: ', gfo.frequency, ' speaker: ', gfo.speaker.identifier)
        # print('speakers_per_region: ', speakers_per_region)
```

I could review is faster (and I believe it would also be easier to read/maintain) if it was like this:

```python

    def data_datasets(self):
        # for use in displaying frequency for a gloss in chart format in the frequency templates
        # the frequency fields are put into a dictionary structure

        #Prepare counters
        total_occuurences = 0
        data_datasets = []
        signers_per_region = {}
        frequency_per_region = {}
        gloss_frequency_per_region = {}
        speakers_per_region = {}

        #Find the frequency regions and frequency objects for for this dataset
        try:
            frequency_regions = self.lemma.dataset.frequency_regions()
        except (ObjectDoesNotExist, AttributeError):
            return (total_occuurences, data_datasets)

        frequency_objects = GlossFrequency.objects.filter(gloss=self)

        #For each region, collect all frequency information
        for region in frequency_regions:

            speaker_locations = frequency_objects.filter(speaker__location=region)
            gloss_frequency_per_region[region] = speaker_location
            signers_per_region[region] = speaker_location.count()
            frequency_per_region[region] = 0

            for location in speaker_locations:
                frequency_per_region[region] += location.frequency

        #For each region , collect all speaker information
        for region in frequency_regions:

            speakers_per_region[region] = []

            for frequency_obj in gloss_frequency_per_region[region]:
                speaker = frequency_obj.speaker.identifier
                if speaker not in speakers_per_region[region]:
                    speakers_per_region[region].append(speaker)
```

Let me know what you think about this.

Woseseltops:
Hi @susanodd, sorry the time between my responses here is so long. 

I reviewed your last commit, and both functions you improved are looking great! Looking at ``get_corpus_speakers`` I came up with some opportunities for making it even shorter, let me know what you think:

```python
def get_corpus_speakers(dataset_acronym):

    try:
        corpus = Corpus.objects.get(name=dataset_acronym)
    except ObjectDoesNotExist:
        return []

    # Get all frequencies mentioning this corpus
    frequencies = GlossFrequency.objects.filter(document__corpus=corpus)

    # Get all speaker identifiers
    speaker_identifiers = [speaker['speaker__identifier'] for speaker in frequencies.values('speaker__identifier').distinct()]

    # Return the list of speakers corresponding to these identifiers
    return Speaker.objects.filter(identifier__in=speaker_identifiers).order_by('identifier')
```

Furthermore, I feel `get_context_data` could use a similar improvement to what you did for `get_corpus_speakers` (like more comments than it has now, and perhaps even splitting it up), but is good enough for this pull request :) .

susanodd:
I've been working on the Corpus Overview template for a while. It's tangentially related to this issue as it provides the data that gets displayed here.

I put everything related to frequencies in a new frequency.py file.
So since that is a different branch, I need to make sure I don't end up doing things twice.

susanodd:
About the get_corpus_speakers. That has been moved to the new file frequency.py

I guess I need to modify it in both places so it will be able to merge later?

susanodd:
Okay, I've modified as you suggest. Also the same in the other branch in the new frequency.py file.

I need to get that branch pushed first since it's nearly done.

susanodd:
Okay the get_corpus_speakers code has been pushed.
It now corresponds to the same code as in the corpus_functionality branch, where it appears in that branch in frequency.py instead of in tools.py as it does in this branch. I made them identical to assist merging later.

susanodd:
Okay, yes, the various things computed in the context are quite confusing.
This has to do with whether or not "this gloss" is (should or should not be) included or not in the list of variants returned by the methods in the Gloss model.
https://github.com/Signbank/Global-signbank/blob/f2d5f61b4210cb171389096b6248cd9d5802a47f/signbank/dictionary/models.py#L1051-L1055
https://github.com/Signbank/Global-signbank/blob/f2d5f61b4210cb171389096b6248cd9d5802a47f/signbank/dictionary/models.py#L1064-L1070
https://github.com/Signbank/Global-signbank/blob/f2d5f61b4210cb171389096b6248cd9d5802a47f/signbank/dictionary/models.py#L1072-L1100

Because some other code also uses these methods, I tried to avoid changing other code not part of this issue.
That's why the code has a weird test whether something is greater than 1.
https://github.com/Signbank/Global-signbank/blob/f2d5f61b4210cb171389096b6248cd9d5802a47f/signbank/dictionary/adminviews.py#L3213
Because itself is included in the list.

The display in the template should only show "variants" if other glosses are variants.
But this gloss also needs to be shown in the list.

I'll see what I can change to make this more readable.

susanodd:
Hmm. The nomenclature is also very confusing. At some point we were differentiating between whether a variant was deduced as a pattern (suffix -A, -B, etc.) or because it was saved as a "variant" type relation under Relations with Other Glosses.
But both of these are included in the list of Variants. But the storage of variants should not be overlapping. (The inferred ones are not saved by the user.)
It looks like the code is a bit obtuse trying to check all these cases.

susanodd:
Regarding the get_context_data code of GlossFrequencyView in adminviews, I see I wrote the code for variants (which are shown in Gloss Frequency View) so as not to modify the methods it calls in models. That's why the code is a bit obtuse, as discussed above.

susanodd:
I made some important changes to the display. The percentage bars in one chart with variants the length of the bars was not filling up the width. This has been repaired.

In the get_context_data python code there is quite a lot of "ready" code that is all used in the chartjs display in the template.
Probably this needs to be moved (in the form of helper functions) to the frequency.py file (which is set up in another branch).

frequency_display_NGT_variants_open-A

Woseseltops:
Hmm @susanodd now I'm a bit confused. I was about to merge this pull request, but now you say you made additional changes for which I don't see commits... do I need to wait for these before I merge?

susanodd:
@Woseseltops would you be able to merge #779 first?
That branch creates the separate frequency.py file.
I want to put all the frequency related stuff from this issue into that file.

susanodd:
You could merge it without the newest changes.
I added about 20 lines of code to get_context_data, all for displaying chartjs. Then I realised it needs to be put elsewhere. All the code related to chartjs preparations can be in a new file: frequency_display.py. There is also functionality in models.py related to frequency display. The models code hasn't been modified by this issue. I tried to keep things inside the issues so they wouldn't affect other code.
This branch is older than branch corpus_functionality.

Woseseltops:
> @Woseseltops would you be able to merge #779 first?

Sure, just did it!

Let me know if this branch is ready for review again.

susanodd:
@Woseseltops I just got this branch merged with the master.
Still need to move chartjs specific code to an additional file.
It is however okay to merge it if you want. Or give feedback.
The extensive code in get_context_data of GlossFrequencyView all for display purposes is quite cumbersome.

susanodd:
@Woseseltops any suggestions in how to structure the code to facilitate maintenance?
E.g., one function for each graph? (This is not currently done. The code was originally trying to reduce the number of iterations, but that hasn't been very successful. So I'm wondering if focussing on what the result should be would help.)

As it is, each of the graphs shown in the GlossFrequencyView needs an entirely different structure for its data.
The code has evolved into various loops to collect the needed data in the correct dictionary structure for use in javascript.
But this isn't very straightforward.
At the moment, I'm moving the "straightline" code (in adminviews) to be bodies of various functions inside of a new frequency_display.py file.
This is a bit sneaky, but it moves most of the code out of the way to improve readability in the get_context_data function.

susanodd:
As an example, there are many loops like this:
`for i_key in ['Female', 'Male']:`
and like this:
`for i_key in ['< 25', '25 - 35', '36 - 65', '> 65']:`

But it's not really possible to combine all the loop bodies into a single loop body.

So far, I've made functions such as these:

`(age_range_labels, speaker_age_data) = collect_speaker_age_data(speakers_summary)`

`(variants_data, variants_data_quick_access, sorted_variant_keys, sorted_variants_with_keys) = collect_variants_data(variants)`

`(variants_age_range_distribution_data, age_range_labels) = collect_variants_age_range_data(sorted_variants_with_keys, age_range_labels)`

So just the signatures are visible in get_context_data.

susanodd:
@Woseseltops I pushed the changes. The code works fine. It's just a question of whether it's sufficiently readable.

susanodd:
The code has been made much more concise. Ready for merge.

Woseseltops:
Ah great, it's truly incredible how much faster I can mentally process your code now. Also, your restructuring of some code into functions in a separate file really works for me. I'll merge the pull request and put it online right now :)

One minor thing @susanodd : I notice you define SEX_CATEGORIES and AGE_CATEGORIES in >1 place... perhaps this is a good thing for a setting?

Issue #752, opened 2021-06-10
Change order of columns on Manage Datasets page


ocrasborn:
The new column to download a CSV file with the links was put between the two upload columns for the frequency functionality; please put those two back together again. I have no preference for where to put the download CSV column.

susanodd:
Okay, this has been done. It's awaiting other changes to be merged.

susanodd:
This is live.
The Hyperlink download button for CSV is before the two columns for EAF.
Perhaps the headers are not clear because the Metadata header has CSV in it?

ocrasborn:
Thanks @susanodd ; I think the headers are fine like this, explaining users what is what will be something we take up in the manual.

PR #775, opened 2021-09-27
Dataset Manager View: Changed order of columns.

susanodd:
This implements Issue #752.


Woseseltops (approved):
Sorry this simple change took so long to review.

Issue #781, opened 2021-11-18
Display available and selected datasets in the header for anonymous users


ocrasborn:
See #780 . For the same reason, we need to display available (public) and selected datasets for anonymous users in the header, like we do for logged-in users. Now, anonymous users see a red header with just the logos and name, but not the datasets.
* To not confuse these users, let's only display the datasets that are actually available to them.
* Note: these are not so much datasets that logged-in users can apply for access to (these get a checkmark under 'Is public' for that dataset in the admin), but datasets that have any glosses that have the checkmark 'In dictionary' under Publication Status set to true.

susanodd:
I've implemented this using a session variable, since anonymous users do not have an account. For users with an account, the selected datasets is stored in the database and retrieved when the user logs in again.

susanodd:
Regarding the display of datasets  for which there are public glosses, would this also be for logged in users?

susanodd:
Here is what this currently looks like for test user_dataset. This user only has access to two datasets.

![available_datasets_test_user_dataset](https://user-images.githubusercontent.com/24272544/146002397-f7d78f79-751d-4864-b599-4c1978ed03e5.png)

![select_datasets_test_user_dataset](https://user-images.githubusercontent.com/24272544/146002418-340beaf3-3418-424a-bf9c-bb05d2a727eb.png)

susanodd:
See #780 for screenshots of the menu bar with (public) datasets for anonymous users.

susanodd:
![anonymous_menu_EmblemsNL](https://user-images.githubusercontent.com/24272544/146010066-becabdd8-830b-442d-ab00-3805f366205a.png)

ocrasborn:
> Regarding the display of datasets for which there are public glosses, would this also be for logged in users?
No, logged in users should always see all datasets.

susanodd:
@Woseseltops I can't figure out how to put the Select Datasets in the Menu for Anonymous Users.
I gave group Guest permission to view this page, but that did not suceed in putting it into the header menu.

susanodd:
Okay, this is fixed now! i had to remove all the allowed groups, since everybody can view it now in the menu.

susanodd:
This is live now.

Issue #780, opened 2021-11-18
Make menu item 'Select datasets' available to anonymous users


ocrasborn:
Now that more dataset managers are making data available to the public, it becomes desirable to have anonymous users also select a specific dataset. Please make that menu item available to them.

susanodd:
![anonymous_user_select_dataset](https://user-images.githubusercontent.com/24272544/146001524-f87dc678-4791-4445-95fe-6d40592f0084.png)

susanodd:
@ocrasborn can you check on EmblemsNL. I followed the instructions above and made datasets which are public available to select for anonymous users. In addition, datasets which contain public glosses are available to select and shown in the header. (#781).
However, some datasets, e.g. EmplemsNL is accessible by others, but has no public glosses.
What should an anonymous user be allowed to do?

susanodd:
(This is all on my local server, not online.)

ocrasborn:
Ah, good point, @susanodd. Only datasets with public glosses should be selectable by anonymous users, not those datasets that are available to other researchers. Should have made that clearer before.
(In fact, we strive towards harmony here (datasets available to other researchers should have public glosses), but copyright reasons may prevent full public access to glosses.)

susanodd:
I've finally got this working for just the datasets with public glosses. (#781).
One thing that I'm not sure about is the situation where the user chooses "None" in the select view.
At the moment, this defaults to "all". This is visible in the header. Although the list shows nothing selected.
Once upon a time, this was desired because otherwise the "no datasets selected" becomes a problem elsewhere.
However it is maybe confusing.

For anonymous users, the header shows with nothing highlighted. (A session variable is used for anonymous users),
whereas for normal users, the default is to highlight the datasets they have access to (when nothing is selected).

Screenshots are below.

susanodd:
This shows test **user_dataset**, who only has permission for two datasets. Now all datasets are shown in the header.
But the user can only select from those that can be accessed. Here "None" has been selected. The header defaults to highlight those this user has access to (rather than "none"). The code has had this behaviour for several years.

![user_dataset_nothing_selected_header](https://user-images.githubusercontent.com/24272544/146546655-1a1f5ccf-8a3b-40b9-83fb-ed2153c6164c.png)

susanodd:
[...]

PR #783, opened 2021-12-17
Select datasets for anonymous users

susanodd:
This branch implements issues #780 and #781.
It sets up a session variable for the selected datasets for anonymous users.


Woseseltops (approved):
This looks like clean and easily readable code to me, great @susanodd ! The decisions you've made make sense to me.

Issue #762, opened 2021-07-22
Make two more fields multiselect


Woseseltops:
Currently you can select more than 1 value for the field *Regional Dialect*, Julie and @ocrasborn are thinking about also wanting this for the fields:
* Semantic field
* Derivation history (ASL only)

susanodd:
Currently the Semantic Field is a choice field, and the Dialect field is a ManyToMany field with the model Dialect.
https://github.com/Signbank/Global-signbank/blob/94dbe93ee7552b7d9cd19416ee29b86d5d18dd61/signbank/dictionary/models.py#L662
https://github.com/Signbank/Global-signbank/blob/94dbe93ee7552b7d9cd19416ee29b86d5d18dd61/signbank/dictionary/models.py#L493
So this issue entails creating a new model for Semantic Fields.

That's interesting. I just did a search on other ManyToMany fields.
The Gloss field signlanguage is also ManyToMany.
https://github.com/Signbank/Global-signbank/blob/94dbe93ee7552b7d9cd19416ee29b86d5d18dd61/signbank/dictionary/models.py#L473
But we don't seem to exploit that because a gloss only belongs to one dataset (via its lemma).
https://github.com/Signbank/Global-signbank/blob/94dbe93ee7552b7d9cd19416ee29b86d5d18dd61/signbank/dictionary/models.py#L470
https://github.com/Signbank/Global-signbank/blob/94dbe93ee7552b7d9cd19416ee29b86d5d18dd61/signbank/dictionary/models.py#L2337-L2338

susanodd:
I'm working on this issue now.
In order to implement multiple select, as discussed above, a new model SemanticField is needed.
Another field that is multiple select is Dialect.

At the moment, the semField of Gloss is a field choice category.
This choice list can be seen in Admin. It has the FieldChoice fields: machine_value, and the hard coded language fields for Dutch, English, and Chinese.
Recall that Handshapes also were once FieldChoice objects. When the Handshape model was created, new fields were added specific to this new model. The original FieldChoice fields machine_value and the hard coded Dutch, English and Chinese language fields were maintained to support compatibility.

@ocrasborn what extra fields would be useful in a SemanticField class?
In what sense are they multilingual?
What language(s) should be included in a semantic field object?
Would it be at all relevant to let Django do the translation part of this?
There are many more languages now than in the beginning.

I'll start out just putting the current choices for semField into a new class SemanticField.

[...]

PR #794, opened 2022-03-01
Semantic fields are multiple choice

susanodd:
This branch makes semantic fields multiple choice.
This is the first part of #762 .

New classes have been created to support this: SemanticField and SemanticFieldTranslation.
Only English is stored in SemanticField under the field "name".
Other translations can be created for other languages in SemanticFieldTranslation.
There is a url for showing all the semantic fields:
/dictionary/semanticfields/
Here the user can also see the translations.
Details are shown in a semantic field detail view. (Analogous to Handshapes.)

In the Gloss model, a new field has been added, semFieldShadow, which is a ManyToManyField with SemanticField.
(This is analogous to another multiple select field Dialect.)
Because semField is a FieldChoice, the functionality for semField in choice lists has been maintained.
The field choices for semField are used in the pull-down menus for Gloss Search.

The Admin supports the two new classes.
Bookkeeping is done to make sure when new semantic fields are created, they are present in both formalisms.
(This is how Handshapes are maintained, in their own class as well as in FieldChoice.)
Furthermore, testing has been added to check that searching on a semantic field, or on two semantic fields, results in a gloss being found.

This branch has a database migration.

Update 30.03.22

Derivation History is now also multiselect.
It has been implemented analogously to Semantic Field, with all the frazzle dazzle.
It also has a database migration.


Woseseltops (changes requested):
Overall: high quality code, as far a I can see. I have one larger thing and one smaller thing.

Larger thing: `configure_derivationhistory` creates HTML by manipulating strings. I think the Django template language was intended to avoid this... would it be possible to do the same thing with Django templates? If so, I think that would the preferred approach.

Smaller thing: `update_semanticfield` and `update_derivationhistory` contain an `except` without Exception

susanodd:
Eeeks! I see what you mean, that except statement is way overloaded.

susanodd:
The configure url is basically a patch. It should only be used once. And only needed by Global.
Although it looks like a patch. It only exists because the DerivationHistory and FieldChoice shouldn't be empty.
For the Handshapes and SemanticFields, there was already data in FieldChoice.
But now FieldChoice has no data for derivHist (except for ASL).
Is it desirable to make the configuration elegant?
I've kind of done it a brute force hack job kind of implementation. :|
At the very least it could stay empty I guess and just let the user see there are no choices?
Or I could make it elegant and let the Admin enter values during configure.
(With a proposed suggestion of those used now, with "Use these?")

susanodd:
Finished @Woseseltops !

Issue #761, opened 2021-07-15
Make the CSV import more user friendly


Woseseltops:
The CSV import already contains some functionality to detect any problems the user may have accidentally introduced, but there are still many other things that could be 'caught' by Signbank. I just tried to import a Catalan file, and these things would have helped me:

* Verify that the filename ends in .csv. If not, suggest that the user uploads a CSV. Current behavior: encoding error.
* If there encoding errors (because the CSV is not a UTF-8 file), explain how the user can user Notepad++ to convert. Current behavior: Django error page.
* Check for extra empty row, and automatically ignore them. Current behavior: complains about missing dataset information.

ocrasborn:
Re: bullet 2: on Mac I always use LibreOffice for this, Excel/Mac's CSV output is not accepted; I'm assuming that's also a file encoding issue. So if that's indeed the case, please add this suggestion for Mac users in the feedback.

susanodd:
This is also relevant for the metadata file of Issue #756.

@Woseseltops what is the naming convention for a CSV file that uses tabs?

susanodd:
I've revised the CSV import routine for issue #756 to use the Python Sniffer to determine the separation character. I'll do that for this one, too.

One caveat for skipping empty rows, I once had a LibreOffice file problem where LibreOffice was not able to detect the end of the file. It got into an infinite loop when trying to open the file. You could scroll down in the window of LibreOffice and there was no end, all the rows were empty but it couldn't find the end. With the spinning timer symbol that it was still loading.
This had something to do with importing files made on a different system, which either had special characters in the cells or somehow a different line break or missing an end of document symbol. We were using NULL values in the database at one time.
So I'm a bit weary of skipping empty rows rather than just stop when the first empty row is encountered.

The Python Sniffer routine, so far it works. Although I just realised it's making use of a readline() to inspect the header row. (To identify the separator. There are other code examples in the documentation that specify a number of bytes to read. But that actually didn't work because it was spanning multiple rows and couldn't identify a separator.)

What I've done is to include a <code> element in the template for creating a Corpus, with the header row for Metadata for Speakers for EAFs so the users can copy that to start with.
(@ocrasborn will need to approve something like that.)

susanodd:
I'm working on this issue now.

susanodd:
I've modified the file input field of the three upload csv templates to only accept various csv encodings 
(as recognised by html input). 
This makes it only show csv files in the file browser selection window.
[...]

PR #797, opened 2022-03-16
Csv user friendly

susanodd:
This branch adds user friendly import of csv files as discussed in #761.

Empty rows are ignored rather than complained about.

Non-UTF-8 encoded files are given an informative error message rather than crashing. (See screenshot in the issue.)

Files that do not end in csv are not allowed to be chosen by the browser input dialog.
If a user has changed the extension of a non-csv file to csv, it is caught in the python code and an error message displayed rather than crashing.

Delimiters can be chosen on the` Import CSV Update` and `Import CSV Lemma` templates.
This was previously only on the `Import CSV Create` template as it was assumed users would obtain an initial csv from exporting.

### DIFF
--- a/signbank/dictionary/templates/dictionary/import_csv_create.html
+++ b/signbank/dictionary/templates/dictionary/import_csv_create.html
@@ -34,7 +34,7 @@
 
 {% if stage == 0 %}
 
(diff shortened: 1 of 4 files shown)


Woseseltops:
Nothing to complain here

Issue #799, opened 2022-03-29
Video import: separate the log messages per dataset


ocrasborn:
When batch importing videos from writable/import_videos (or something of the like), the system looks in all subfolders for all datasets and does its thing. The log message that the user then gets (admin users, for now) combines log info for all datasets. Please separate the log per dataset with a clear header, keeping the different categories of log for each dataset.

The present situation can be a bit confusing, as I often add new videos for dataset X, while there are also some remaining problematic files for other datasets (where the filenames don't match a gloss in that dataset for the specified language in the subfolder). Cleaning those up would be good practice, of course...

susanodd:
@ocrasborn do you know if the folders per dataset remain? Or is this cleared up afterwards? Do you typically upload an entire folder with the name of the dataset at the top level? Or do you upload files to a dataset folder? (How do the files get into the folder writable/import_videos ?) The url command in question assumes there are already files and folders at this location.

In order to rewrite the code (on my local server), I commented out the specific code that stores the videos in the database and removes files. I replaced this with print statements to see what's going on, and facilitate development using test data.

I will make the output as you describe, by maintaining a data structure during processing.
At the moment, the code seems to just ramble along listing stuff and "printing" to the output string that is given at the end as the result.
If you actually try it with absolutely nothing in the directory, it still prints stuff by no data (no file names).
It looks like the code removes things it has copied to the normal location.
So from what you describe, it leaves files it cannot process?

susanodd:
The code is a bit weird. This is what it looks up for the Gloss:

https://github.com/Signbank/Global-signbank/blob/1d9c7b6c9bd58ebd16c1d9bbbc61cf89509b032b/signbank/dictionary/views.py#L544-L555

The error message shown to the user mentions gloss ID. But the code is looking up on annotationidglosstranslation.

susanodd:
@ocrasborn I had a look at the folder on lightning. Indeed, there are folders for datasets underneath import_videos.
The subfolders underneath the dataset acronym (or name) are e.g.: nld, eng
So this is a subfolder named by a language code, as observed in the code:

https://github.com/Signbank/Global-signbank/blob/1d9c7b6c9bd58ebd16c1d9bbbc61cf89509b032b/signbank/dictionary/views.py#L538

So far I can't find any code that puts the files in this location. I assume you upload them?
The folders do not get deleted. From the code, it looks like videos that have sucessfully been moved get deleted from this location, otherwise there is an error message feedback.

[...]

PR #804, opened 2022-04-22
Issue #799: video import feedback per dataset in table view

susanodd:
This implements user friendly feedback on importing videos from folders.
To eliminate extraneous feedback about other datasets, it is limited to the selected datasets.
A template view table shows the status of imported videos for the datasets and their language folders, as described in #799.


susanodd:
Good point about the column dataset.
Originally there wasn't a template. Then @ocrasborn suggested that. 
Then since the parameters to show the menu header were used, the selected datasets were at hand.
Hence only the selected datasets are shown in the table.

susanodd:
Finished!

Issue #800, opened 2022-04-13
Cannot change values for Weak Drop/Weak Prop


ocrasborn:
Once a value is selected in the pull-down menu for Weak Drop or Weak Prop (Panel Phonology), this cannot be undone: an error pops up. (Setting the initial value works fine, it's saved correctly after leaving the edit mode.)

susanodd:
Good find! Indeed, it gives an error if you want to set it back to Neutral.
What's weird is it gives a constraint violation of NOT NULL.
However, in the model, this is allowed to be NULL, so it's very strange it gives this error.

susanodd:
In the error log, it's giving this error from model GlossRevision (which is where the history of changes is stored).

susanodd:
Fixed! Here it's visible in the Revision History.

![setting_weakdrop_to_Neutral](https://user-images.githubusercontent.com/24272544/164705958-a310a7b9-9509-4840-87ef-c13cc61693e3.png)

susanodd:
@ocrasborn does the Revision History need to show the "pretty" values in the Old value column?
I'm not sure what else is done with Revision History.
I don't think we allow undo's using the values stored there or use history items for anything. (?)

susanodd:
I pushed the fix as shown above.

susanodd:
This has been awaiting in a pull request for quite a while.

susanodd:
This is live now!

Issue #760, opened 2021-07-15
Add 'tags' to Revision History


ocrasborn:
Tag editing/adding/removing is not yet listed in the revision history, but this would be helpful to us at times. (See #115 )

susanodd:
Okay, this one is next.

susanodd:
Here's what this looks like.


![revision_history_tags](https://user-images.githubusercontent.com/24272544/165090127-01de9e86-5ea9-4313-b1a4-d57f85a1c93e.png)


Because the table for the Gloss Revision History includes columns for field, old value, and new value, I've made the list of tags become the value.

susanodd:
The revision history items have existed for a while. So far there isn't any conversion of already existing field into user friendly names. For the labels code I made use of Django translation as `_('Tags')`
If needed this could also be done for other fields, or it could be done in the template.
Issue #800 reveals a similar quirk in the display of the gloss revision history.

susanodd:
I see in #115 that it looks like the intention was that the display is the "original human value".

So what about field names? These as well?

(I ask because at the moment, the stored data in the history uses the field names. This can be made user friendly during generation of the template for the interface, including translation.)

susanodd:
Okay, I've added dynamic translation of the field names in the Gloss Revision History. The actual field names are stored in the history.


![gloss_history_fieldname_translation_en](https://user-images.githubusercontent.com/24272544/165289941-1f3502d0-087a-4251-9cfd-b7f03d03ce9e.png)

![gloss_history_fieldname_translation_nl](https://user-images.githubusercontent.com/24272544/165289973-f739d0cf-9bae-4cc9-b8a5-b6bf6a995f56.png)

susanodd:
Regarding Tags, the set of tags is shown, before and after.
Tags is not a field, it's a relation.

At the moment, other types of modification to a gloss (i.e., relations rather than fields) are not stored in the history.
These could also be saved, although the GlossRevision has fields for `old_value` and `new_value`, which doesn't lend itself to relations. That's why it's done as shown above.

The table above could be "split" to show the history as changes to fields and to relations.

Above, the actual tags of "before" and "after" are retrieved when a tag is added or removed from a Gloss.

At this moment, this update to tags is added to the GlossRevision history. Above, this has been done by storing the before and after lists of tags (reformatted as a string). (Basically what you see in the table for `old_value` and `new_value` is stored in the GlossRevision history.)

[...]

PR #805, opened 2022-04-22
Issue #800: Setting weakdrop back to Neutral; #760: Tags in GlossRevision; Refine Display

susanodd:
This implements issue #800.
The bug was located in saving to GlossRevision when the new value is None (aka Neutral).
This has been repaired by saving the human readable value Neutral in the history.


susanodd:
This branch will also cover #760 since that adds pretty human readable field names to the template.

susanodd:
This is ready now.

Issue #801, opened 2022-04-13
Searching for translation equivalents on the search page doesn't work anymore


ocrasborn:
Searching for any string in the Translation Equivalent fields per language, in the search form, incorrectly returns all signs in the selected data sets. Searching for a translation equivalent in the menu bar search field works fine, though.

susanodd:
This has probably been broken for a while. I kind of didn't remember there was such a functionality. But yes, it's right there in the header! There didn't used to be many glosses for which this information had been filled in.

susanodd:
I found it. There is a discrepancy in the identifier used in the form.
One place uses "keyword_" and the other uses "keywords_".

susanodd:
We kind of ignored a forms setting that defines the "keywords_" prefix and basically hard-coded it as "keyword_" in lots of templates where this search field occurs.
I chose to make it be "keyword_" as the constant in forms, since this is how it's hard coded elsewhere.
Less coding to fix it this way.

susanodd:
This has been repaired and is on branch translation_equivalents.

susanodd:
This is live now.

PR #806, opened 2022-04-22
Issue #801: Repaired search field translation equivalents.

susanodd:
This issue effectively renames a constant defined in forms to match the hard coded usages of it.
Also the reverse is done as both identifiers were being used.

keywords_  => keyword_

Because of the numerous usages, it was easier to modify things to match up with the "keyword_" identifier.


Woseseltops (changes requested):
Good find. I see there's still some debugging print statements. I would say either remove them or make it more clear what they are logging and why

susanodd:
Changes completed!

Issue #492, opened 2018-09-05
Search by Phonology for Morphemes


susanodd:
The search morpheme query page currently resembles that for signs.
The main difference is the Search by morphology panel, since it's morphemes we're looking for rather than morphology of a gloss.
Check whether e.g., the phonology search fields are relevant, or should be hidden for morphemes.

ocrasborn:
The search fields are relevant, but the list of fields needs to be adapted: #638 and #639

susanodd:
I'm looking at this issue now.

On the normal Search page, you can do either Sign or else Sign or Morpheme.

When searching on Morpheme Type (Hands: Handedness) via the Search Morpheme page, this returns five morphemes.

But the same query (Morpheme Type: Hands: Handedness) on the normal Search (using Sign or Morpheme) does not return any results.

susanodd:
Here is the revised signs/search on morpheme type, sign or morpheme (#689):


![search-sign-or-morpheme-type-handedness](https://user-images.githubusercontent.com/24272544/110640462-22532180-81b1-11eb-928b-b6050e91dc18.png)

![search-sign-or-morpheme-type-results](https://user-images.githubusercontent.com/24272544/110640484-27b06c00-81b1-11eb-934f-4141ebeb86a0.png)

susanodd:
This has been pushed to theoddbranch.

Woseseltops:
Hey @susanodd are you introducing a new setting MORPHEME_DISPLAY_FIELDS here? If so, it should be added to `settings/server_specific/default.py` . Other than that nothing to complain ;)

susanodd:
Check!

susanodd:
The morpheme lookahead in the Search by Morphology was broken. I just fixed this.

susanodd:
This has been pushed to branch `search_by_morphology`.

susanodd:
This has been awaiting in a pull request for quite a while.

susanodd:
This is live now.

PR #809, opened 2022-05-06
Issue #492: Typeahead for morphemes in Search by Morphology.

susanodd:
This solves a bug in the typeahead for morphemes in the Search by Morphology panel of List View.

### DIFF
--- a/signbank/dictionary/adminviews.py
+++ b/signbank/dictionary/adminviews.py
@@ -6032,30 +6032,22 @@ def handshape_ajax_complete(request, prefix):
     return HttpResponse(json.dumps(result), {'content-type': 'application/json'})
 
 def morph_ajax_complete(request, prefix):
-    """Return a list of morphs matching the search term
+    """Return a list of morphemes matching the search term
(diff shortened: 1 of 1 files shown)


susanodd:
Okay, I noticed that when I read the code myself after  not looking at it for some time, it's unclear why it was necessary to change it the way it's been changed.

The bug was with this check:

https://github.com/Signbank/Global-signbank/blob/d28c843475141b8a0ede17b7977f88d93b27aa85/signbank/dictionary/adminviews.py#L6058

For some reason this was causing zero results to be found in spite of there being results.


I rewrote the code to first get morphemes with matching annotations and then get the annotation text from that.

Checking for the appropriate dataset is done elsewhere:

https://github.com/Signbank/Global-signbank/blob/d28c843475141b8a0ede17b7977f88d93b27aa85/signbank/dictionary/update.py#L1315-L1318

(And lo and behold, yet another text that's not translated. . .)

susanodd:
The code in update.py where the translation is lacking has been revised in #703.

Issue #623, opened 2020-02-18
Check rights for users with 'write' permission on dataset


ocrasborn:
A dataset manager reported to me that a new user whom he gave change/write access to his dataset was not able to add signs. I had to assign the role of 'editor' to get around this, but would like to understand what the situation is. People with change permissions should be able to alter and add signs/glosses, lemmas, and morphemes (but not handshapes). If I as a superuser need to do something in the admin for these users, the management of dataset permissions for dataset managers wouldn't mean much. Do we need to change something here to make it work, or was something else going on for this particular case / at that particular time (late last week, when the registration was also broken, #622 )?

susanodd:
In the code of template add_gloss.html
there is a permissions check
dictionary.add_gloss

vanlummelhuizen:
@ocrasborn, what exactly does this mean, in terms of giving permissions?:
> gave change/write access to his dataset

In Signbank a permission applies to an action for either a model (e.g. Gloss) or for an individual object of a model (e.g. dataset 'NGT'). Normally a permission does _not_ apply to a model/object _and_ all things that are connected, e.g. dataset 'NGT' and all it's glosses.

So if a user has the 'change dataset X' object permission, but not the 'add gloss' model permission, the user may change the dataset's name, acronym, description and other Dataset fields, but not add a gloss to that dataset, since adding a gloss is about the Gloss model, not about a Dataset object.

So at first glance giving someone change-dataset permission and add-gloss permission (by adding the user to the Editor group) seems a about of right.

We could think of a way to check permissions in a different way, e.g. if a user had 'change dataset X' permission, the user also may add/change/delete glosses/lemmas/lemmaidglosstranslation. This may involve quite some changes to the codebase.

ocrasborn:
Perhaps it need not be too complex: dataset managers have different roles than editors. On the page Datasets > Manage Datasets, the managers configure who has view access to the dataset including its lemmas/glosses/translations, and who has add/change/delete permissions for that content of the dataset, and thus can act as an editor for that dataset.
So "Users who can Change Dataset (Group permissions excluded)" that is modified on the Manage Datasets page should not be about modification of the dataset (that remains a privilege of the manager), but about modification of glosses/lemmas/translations.

vanlummelhuizen:
@Woseseltops @susanodd As discussed with @ocrasborn, let's discuss more about this after a week or two when @ocrasborn return from his trip.

susanodd:
[...]

PR #811, opened 2022-05-12
Issue #623: Grant change dataset puts user in Group Editor.

susanodd:
In the Manage Datasets View, granting change permission to a user now puts the user in Group Editor if they are not in this group yet. Similarly, revoking change permission and leaving no datasets the user can change removes the user from Group Editor.


Woseseltops (approved):
I like this piece of code in particular, really readable

```python
        datasets_user_can_change = get_objects_for_user(user_object, 'change_dataset', Dataset, accept_global_perms=False)
        datasets_user_can_view = get_objects_for_user(user_object, 'view_dataset', Dataset, accept_global_perms=False)
        groups_user_is_in = Group.objects.filter(user=user_object)
```

Issue #648, opened 2020-04-28
Create way for users to see account expiry date and apply for extension


ocrasborn:
Users start making new duplicate accounts because their old account is expired... Check: is it possible now to create two accounts linked to the same email address? That should be made impossible, but since people have so many email accounts nowadays, that's not a universal catch and we'd better facilitate for people to ask for an extension.

susanodd:
@vanlummelhuizen and I are both looking at the accounts right now.

One observation, the test user accounts share your email @ocrasborn.
That may be why it is the way it is, to allow testing with different permissions, with real accounts.

susanodd:
I'm creating a User Profile page for the logged in user.
Here it will show the expiry date.
There can be a button to make this farther in the future.
Can a user do this themself (i.e., automated)?
Should it go via the semester?

The user will need to be logged to view their profile (expiry date).

Do you want some way to alert users if their account is due to expire (automated)?
Or to check when the user attempts to login to an expired account, that something happens (automated)?

There's a Forgot Password facility.
Perhaps a Forgot Login? or Login Expired?
Is it possible people forget their username and therefore create a new username?

ocrasborn:
Hi @susanodd : I really only set an expiry date for students who want access, and for whom I'm not sure they will stay in the field. For other users, I just leave the field blank. I'd rather keep control over this, so please do not allow users to update their own expiry date. Sending them a message is not necessary either.

ocrasborn:
About the 'forgot password' etc.: no opinion myself, I would always ask @Woseseltops or @vanlummelhuizen for what they think is safe enough and/or what is considered best practice.

susanodd:
The Forgot Password is already there. It's been there since donkey's.

So to clarify, merely let the user view the Expiry Date.

susanodd:
Here is a mockup showing a Profile link to the left of the Logout link.
Clicking you get shown the following (without the actual data):

![userprofile-expiry-date](https://user-images.githubusercontent.com/24272544/94274794-69709180-ff46-11ea-8ad9-034a38a1c7f6.png)

ocrasborn:
Seeing this in a screen shot, I realise what obvious information we should add there: datasets to which the user has read resp. write access! Could you add that?
Had you thought of how to make this accessible to the user, @susanodd ? By clicking on the user name next to 'logout', or would that be confusing? Perhaps a menu entry 'User profile' under the 'About' menu?

vanlummelhuizen:
[...]

PR #812, opened 2022-05-13
Issue #648: Extend Expiry facility added to User Profile.

susanodd:
This branch adds a button to user profile to extend the expiry date of their account.
The button is shown when there are less than 60 days remaining.
Pushing the button results in extending the expiry by 6 months.
This is done without need for the Admin.

### DIFF
--- a/signbank/dictionary/update.py
+++ b/signbank/dictionary/update.py
@@ -2562,27 +2562,37 @@ def upload_eaf_files(request):
         return HttpResponseRedirect(reverse('admin_dataset_manager'))
 
 def update_expiry(request):
+
     # if something is wrong with the call, proceed to Signbank Welcome Page
     # This can happen if the user types in the url rather than getting there via the User Profile View
     # And the Extend Expiry button was not shown on their profile
+
+    # Check for request type
     if request.method != "POST":
         return HttpResponseRedirect(settings.URL + settings.PREFIX_URL + '/')
 
+    # Check if we have a username
     if 'username' in request.POST.keys():
         username = request.POST['username']
     else:
         username = ''
 
-    if username and request.user.username == username:
-        try:
-            user_profile = UserProfile.objects.get(user=request.user)
-        except ObjectDoesNotExist:
-            HttpResponseRedirect(request.META.get('HTTP_REFERER'))
+    if not username or request.user.username != username:
+        HttpResponseRedirect(request.META.get('HTTP_REFERER'))
 
-        from dateutil.relativedelta import relativedelta
-        expiry = getattr(user_profile, 'expiry_date')
-        if expiry:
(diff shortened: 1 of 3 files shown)


Woseseltops:
Can I suggest working with guard clauses here again, to make the happy path clearer? So instead of

```python
def update_expiry(request):
    # if something is wrong with the call, proceed to Signbank Welcome Page
    # This can happen if the user types in the url rather than getting there via the User Profile View
    # And the Extend Expiry button was not shown on their profile
    if request.method != "POST":
        return HttpResponseRedirect(settings.URL + settings.PREFIX_URL + '/')

    if 'username' in request.POST.keys():
        username = request.POST['username']
    else:
        username = ''

    if username and request.user.username == username:
        try:
            user_profile = UserProfile.objects.get(user=request.user)
        except ObjectDoesNotExist:
            HttpResponseRedirect(request.META.get('HTTP_REFERER'))

        from dateutil.relativedelta import relativedelta
        expiry = getattr(user_profile, 'expiry_date')
        if expiry:
            user_profile.expiry_date = expiry + relativedelta(months=+6)
            user_profile.save()

    return HttpResponseRedirect(request.META.get('HTTP_REFERER'))
```

Something like this

```python
def update_expiry(request):

    # if something is wrong with the call, proceed to Signbank Welcome Page
    # This can happen if the user types in the url rather than getting there via the User Profile View
    # And the Extend Expiry button was not shown on their profile

    #Check for request type
    if request.method != "POST":
        return HttpResponseRedirect(settings.URL + settings.PREFIX_URL + '/')

    #Check if we have a username
    if 'username' in request.POST.keys():
        username = request.POST['username']
    else:
        username = ''

    if not username or request.user.username != username:
        return HttpResponseRedirect(request.META.get('HTTP_REFERER'))

    #Check if user exists
    try:
        user_profile = UserProfile.objects.get(user=request.user)
    except ObjectDoesNotExist:
        HttpResponseRedirect(request.META.get('HTTP_REFERER'))

    from dateutil.relativedelta import relativedelta
    expiry = getattr(user_profile, 'expiry_date')

    #Check if we have an expiry date
    if not expiry:
        return HttpResponseRedirect(request.META.get('HTTP_REFERER'))

    #Update the exprity date
    user_profile.expiry_date = expiry + relativedelta(months=+6)
    user_profile.save()
    return HttpResponseRedirect(request.META.get('HTTP_REFERER'))

```

susanodd:
Yes, I see that is much easier to read!

susanodd:
This has been done as suggested!

Now just awaiting approval from @ocrasborn.

susanodd:
This has been approved off-line by @ocrasborn.
I'm merging it into master.

PR #813, opened 2022-05-18
Issue #703: Translations, Caps in Feedback pages.

susanodd:
This issue adds translation to the Headings and User Forms in the Feedback section.

It also fixes a bug with a missing Feedback Video.
It tests that the (previously uploaded as part of a Feedback comment) video exists on the file system.

On my local server this was generating errors because I don't necessarily have copies of videos that are on Global as part of Feedback. These files are located here:
https://github.com/Signbank/Global-signbank/blob/cbfe53d2b452a88068a7649e6e436f70ed906228/signbank/settings/server_specific/default.py#L11
This will also catch Feedback video files that have not been ported to a new server.


susanodd:
I see I need  to do something about the self.get_object() methods in the try-except codes inside of the overridden get methods. The exceptions should include Http404.

PR #814, opened 2022-05-19
Csv user friendly repairs to delimiter parsing

susanodd:
This modifies the three Import CSV functions to use the Delimiter in parsing the Header row.

The code has been enhanced to do the things related to parsing the CSV file at the front, so that the Import File Form is shown to the user. Any errors related to this are shown as messages in the template.
Namely, if the wrong delimiter is chosen, or the encoding is wrong, or a required column is missing, these are all shown at the start.

This essentially repairs a bug with Apple formatted CSV files not being imported correctly.


Woseseltops (approved):
Appreciate the code improvements! I feel like there are even more opportunities for improvement, but that's outside the scope of this issue.

PR #815, opened 2022-05-20
Translations fuzzy

susanodd:
Resolved fuzzy translations in the NL po file.


Woseseltops:
Up to you, @susanodd , but I personally would not have made a pull request for this, as these are not the kind of changes that are hard to reverse, bug prone, or might slow others down in the future (which is what we try to prevent with code reviews).

Issue #816, opened 2022-06-01
Add tab 'List view' in detail view to go back to the search results


ocrasborn:
Or alternatively, create a button to go back to the search result list -- if any was generated to get to this sign, of course. (Now users have to use their browser buttons to go back if they don't want to recompose the search, but just going back page by page often means they are first sent back to the edit view of the sign, which confuses them: am I undoing what I just edited?! A separate function in the interface to move back to the search result list would be user-friendly and avoid confusion.)

susanodd:
I noticed while working on this that the code keeping track of all the search parameters is extremely lengthy.
I've made a data structure to save the ones that are filled in while going through the parsing.
I'll try to improve this code. It looks like we just kept adding additional if-constructs as new things were added.
Perhaps this will combine with #34 .

susanodd:
I've added the current query parameters to the top of the "query results list view".
The results shown below the parameters are those of the search_results horizontal bar.
The results are just links to the other glosses. Other information could be shown in the list.
Next is to add a link to go to the query page.
At the moment, I've made several helper functions, several for pretty printing the query parameters to the table (in human readable form), and another for generation of a Q query filter with all the specific fields and values of the parameters. That is passed as a parameter to the filter on the (Gloss) model objects.


![list_query_view_en](https://user-images.githubusercontent.com/24272544/174280406-aa334cee-03bb-4938-b601-480489665e03.png)

![query_list_view_nl](https://user-images.githubusercontent.com/24272544/174280434-cd1de244-ab79-4cd4-b5e6-0f642e519550.png)

susanodd:
It's working now!

Here is the List View showing the parameters and results and a button Run Query.
The button goes back to the Gloss Search page with those parameters filled in.


![query-k-1_curved-chin-nose](https://user-images.githubusercontent.com/24272544/174793890-aea5e2d1-e884-48e0-bff8-ada962fea030.png)


![run_query_search_parameters](https://user-images.githubusercontent.com/24272544/174793921-9ec0c57e-956f-4629-8059-dc5f64afd964.png)

susanodd:
Oh, the Values of the query parameter Fields are merely shown blue if they are multiple select fields.
It looked a bit weird in a list because they are independent.
At the moment, the query parameters are done with a session variable.
A new url parameter "query" has been added when going back to the Gloss Search List View.
This is needed in order to make it unambiguous as to whether this is a previous search parameters.
[...]

PR #824, opened 2022-06-23
Query Parameters Tab for Gloss Detail View

susanodd:
Okay, I can see it's a bit confusing to come up with a title for what this is.

See #816.

All the query parameters in the GlossSearchForm are now incorporated.
The Tab view shows the query parameters and the results.
At the moment, the results are merely a vertical list that is the same as the horizontal search results bar.
Clicking on Run Query (shown below the parameters) returns to the GlossSearchForm with the same parameters selected.
A URL parameter "query" is used to indicate that the parameters are being used.

FIXED by adding a new context variable for it.
There is one fuzzy thing in showing the parameters in the form on returning to the search page.
If you've done a "Search for gloss with this as morpheme" and chosen a morpheme from the lookahead list,
this is correctly shown in the Query Parameters.
However, when returning to the Search, it shows the id of the morpheme rather than its idgloss.
But the Global server also does this!

@Woseseltops see if this looks okay. I'm pleased with how it works. The main search page is a bit unwieldly and this seems a bit more user-friendly.

A next step is to introduce tabs for refining the results further, as well as facilities to store queries.
Some of the queries are a bit weird in that they don't necessarily have anything to do with the "Gloss".
Like if you search for things like whether a Note has been published.


susanodd:
I removed the unused code.

susanodd:
Added the Query popup as proposed by @Woseseltops  in #816.
Added settings for the Query popup and Query View alternatives, plus the infrastructure to support this.

susanodd:
@Woseseltops if you take a look at GlossDetailView in this branch, you will see that the code of the Query Parameters popup modal is included inline in the template. I would like to put this in a separate file. But importing it that way did not allow "trans" to appear in the file to be imported. (Because of needing to include this inline, I've only put the Query button in the Detail View, not repeating it in all the views...)
Perhaps a newer Django would allow to import an other template?

Woseseltops:
Not sure @susanodd, but I guess we will want to merge this feature before we upgrade Django ;)

susanodd:
This is kind of becoming quite a lot of code. I've been developing the Query View (see screenshots in #816).

Woseseltops (commented):
Noticed various comments describing each 'paragraph' of code, that's good! Also noticed that you removed unused print statements and log strings to the console, also good!

Suggestions and questions:
* I see a lot more stuff has been added `tools.py`... do you think what you added could also go together in separate .py file? I'm asking because `tools.py` keeps growing and growing, and the stuff within it is getting more and more varied. It would be nice to split it up into more functionally coherent bits.

* About the `SEARCH_BY_X` settings you added... you currently have the groups hardcoded. How would you feel about mimicking the structure we have for the `FIELDS` setting, where the field category are dictionary labels? So instead of `SEARCH_BY_SEMANTIC_FIELDS`, it would be something like `SEARCHABLE_FIELDS_PER_CATEGORY['semantics']`.

* I noticed in `gloss_query_view.html` there is also a bit of CSS. Perhaps that's better off somewhere in the static files, as a separate file?

* Also in `gloss_query_view.html`, I think this bit`{% if not forloop.last %} {% endif %}`, which I see 4 times, doesn't do anything, right?

* `In tools.py`, `list_to_query`, perhaps the 'happy flow' can be un-indented?

* I think in `tools.py`, the two `pretty_print` functions could use some more comments explaining what is going on... it's taking my a lot of mental effort to parse ;)

* In `adminviews.py`, I see a list of `if`-`else` statements of this format (shortened by me for clarity):

```python
if fieldname == 'dialect':
    column_headers.append((fieldname, _("Dialect")))
elif fieldname == 'signlanguage':
     column_headers.append((fieldname, _("Sign Language")))
 elif fieldname == 'hasothermedia':
     column_headers.append((fieldname, _("Other Media")))
```

I think it would be slightly cleaner to do it like this:

```python
fieldname_to_column_header = {'dialect': 'Dialect', 'signlanguage':'Sign Language', 'hasothermedia':'Other media'}
column_headers.append((fieldname,_(fieldname_to_column_header[fieldname]))
```

* I'm a bit confused by this bit, can you help me understand what it does?

````javascript 
/**
 * @param {string} field_name - name of the field to sort on
 * @param {string} action     - one of: desc, asc, del
 * @param {string} frmName    - name of the <form ...> that contains the 'sortOrder' <input> field
 * @returns {void}
 */

```

susanodd:
Yes, that's a good idea to put things in a new file.
Very welcome code suggestions!
I think a few things got out of hand because I kept adding new cases as I worked them out.

Where is this last bit, in which file?
/**
 * @param {string} field_name - name of the field to sort on
 * @param {string} action     - one of: desc, asc, del
 * @param {string} frmName    - name of the <form ...> that contains the 'sortOrder' <input> field
 * @returns {void}
 */

susanodd:
I see there might be confusion because I didn't delete the "List View" of GlossDetailView (the original view was a function in views, whereas the improved Query View under Analysis is an actual View class in AdminViews.
So the original version didn't get removed (yet).
There is also the (wish-list) template for the popup for Query Parameters, which isn't used because Django did not like it, so the modal popup is inside of the GlossDetail template.

susanodd:
The forloop in gloss_query_view.html is adding a space between the buttons.

susanodd:
The "analysis" query view also has the buttons, but that template came later so the looping code that generates the buttons uses the settings. And these buttons actually toggle columns in the newer version.

I removed the offending `gloss_query_view.html` template and its `views.py` function. Of course, if @ocrasborn ends up choosing that version, it's a different story.

susanodd:
The pretty print functions arise because things set up by the Gloss Search Form are slightly different than the actual fields (or relations) that are being queried upon. The idea of the Query Parameters data structure is to save the "data" that was entered in the original Search Form. So the pretty printing is mapping between the different representations in order to display them.
These got worked out further when the Query View moved away from being Gloss-centric, so now all the relations can be shown in the results (under Analysis->Query View).

susanodd:
I removed the unused code. Sorry I should have done that earlier I guess. I have local copies of it, but it's removed from the branch. I'm not sure what to do with the `query_parameters_popup.html`. That's not used. But the intention was to use that so it's not inside the Gloss Detail View template.

susanodd:
I've added a lot of comments to the two pretty-print functions to describe why they are needed. These are in a new `query_parameters.py` file.

susanodd:
I just found the code that the last bullet refers to.

I see why this looks suspicious.

The table of results is generated by numerous Ajax calls. One for the header and the others for each row of the table.
These share code with the same ajax calls of the other search templates.
In this case, the sort order row is not shown. So this has to be hidden from the ajax-generated header.

susanodd:
I've removed the popup template file, since it's not being used.
I revised the default query parameters (which show up in the Query popup if you haven't done a search).
I needed to refine this in order to not apply them unless the user want to. Hence the popup for the "no query has been done" case goes via the GlossDetailView instead of being a direct link to the GlossSearch page. (This required some other changes to achieve this.)

Now I'm on the last bullet. The SEARCH_BY settings. What a mess!
I introduced those to make the GlossSearch template more concise. But now there are way too many settings.
Stay tuned...

susanodd:
Hmm. If you look at the GlossSearch display regarding the only two "SEARCH_BY" occurrences in the template:
https://github.com/Signbank/Global-signbank/blob/0a8fd45ea88161910f4c87703d0f98658a07c15f/signbank/dictionary/templates/dictionary/admin_gloss_list.html#L665-L670
and
https://github.com/Signbank/Global-signbank/blob/0a8fd45ea88161910f4c87703d0f98658a07c15f/signbank/dictionary/templates/dictionary/admin_gloss_list.html#L681-L686

You will notice that the functionality doesn't match the settings (of the same name). Indeed here in the View something else is done. First the settings are used (this first bit has been repaired to use the already existing settings):
https://github.com/Signbank/Global-signbank/blob/0a8fd45ea88161910f4c87703d0f98658a07c15f/signbank/dictionary/adminviews.py#L314-L319
But here, something unexpected has been done:
https://github.com/Signbank/Global-signbank/blob/0a8fd45ea88161910f4c87703d0f98658a07c15f/signbank/dictionary/adminviews.py#L413-L421
And following through gets you this:
https://github.com/Signbank/Global-signbank/blob/0a8fd45ea88161910f4c87703d0f98658a07c15f/signbank/tools.py#L2077-L2082

So I've "made" something to use in the GlossSearch template to shorten the code (#830) but have introduced confusion and inconsistency in how the things are named.

susanodd:
I toggled the case of the two "other" search-by things.

susanodd:
The above settings have been repaired as suggested!

Issue #833, opened 2022-08-25
Create tool to clean up video and other_media files


Woseseltops:
Create a subcommand to `develop.py` called 'remove_unused' (or something like that) with an argument that can be either `video` or `other_media` (and it would be cool if it helped the user if the command is not one of those two). This command will:
* Go over all video files, or all files in the other_media folder
* Try to find a matching Django instance in the database
* If there aren't any, add it to a list
* Show the list to the user, ask them to agree
* If so, delete all files in the list from the server

Tip: what we call `develop.py` is called `manage.py` in the rest of the world, might help googling how to add subcommands.

susanodd:
https://github.com/Signbank/Global-signbank/blob/9dd10ca2523aae0e261557cc36bb0da1e1725540/signbank/settings/server_specific/default.py#L9-L10

https://github.com/Signbank/Global-signbank/blob/9dd10ca2523aae0e261557cc36bb0da1e1725540/signbank/settings/server_specific/default.py#L13

They're under writable:
https://github.com/Signbank/Global-signbank/blob/9dd10ca2523aae0e261557cc36bb0da1e1725540/signbank/settings/server_specific/default.py#L8

susanodd:
If you look at the various urls, you can see some commands that relate to videos, images, and other media. The database keeps track of these files and they may or may not be in the file system. Previously, video filenames were being suffixed with ".bak" or with a number instead of overwriting them. There are many of these dangling files. When they get deleted by the user, they are deleted in the database, but the files end up staying in the file system. There are some commands to look for these files, but nothing has been done to actually remove them. That's why it will be helpful to get user feedback. It's also something say @ocrasborn ought to run since some of the files are quite old. Although there are also multitudes of "test" files which are mostly totally wrong videos that are not for the sign.

Woseseltops:
Just remembered, in `dictionary/views.py` there is also a view `get_unused_videos` that does something similar. It makes more sense as a subcommand, though, so feel free to take inspiration from the view, and delete it when you're done :) . 

> where is the other_media folder located? And the video files?

As @susanodd showed, you can specify this in your settings file. In the production server, the paths are `/var/www/writable/glossvideo` and `/var/www/writable/othermedia` .

> Should this command be added to the video app?

Sounds like a good place to me ;)

Jetske:
self.stdout.write(os.listdir(mediapath), ending='')
[...]

PR #840, opened 2022-09-08
Command added for finding unused files and test #833

Jetske:


Woseseltops (commented):
This seems like good code to me, thanks! :) 

General things I noticed:

* I see relatively short variable names, like `notexist` or `todelete`. Given that we all use autocomplete, I slightly prefer more description ones, like  `non_existant_file_names` or `files_to_delete`.
* Code that is split up into 'paragraphs', containing a few lines that are functionally coherent, is easier to review. So instead of:

```python
    self.stdout.write("\nUnused files found are:\n")
    for unknown_file in notexist:
        self.stdout.write(unknown_file)
    if len(notexist)>0:
        # Ask user to agree
        todelete = input("Delete files? [Y/N]")
        if todelete.lower() == "y":
            for subdir, dirs, files in os.walk(chosen_path):
                for file in files:
                    if file in notexist:
                        os.remove(os.path.join(subdir, file))
            self.stdout.write('Files are deleted')
        else: 
            self.stdout.write('Files are not deleted')
    else: 
        self.stdout.write('No unused files found')
```

you could do 

```python 
    self.stdout.write("\nUnused files found are:\n")

    for unknown_file in notexist:
        self.stdout.write(unknown_file)

    if len(notexist)>0:

        # Ask user to agree
        todelete = input("Delete files? [Y/N]")

        if todelete.lower() == "y":
            for subdir, dirs, files in os.walk(chosen_path):
                for file in files:
                    if file in notexist:
                        os.remove(os.path.join(subdir, file))
            self.stdout.write('Files are deleted')

        else: 
            self.stdout.write('Files are not deleted')

    else: 
        self.stdout.write('No unused files found')
```

* We try to work with guard clauses where we eliminate the error conditions first, and then have the happy path unindented. So instead of the above, you could do

```python 
    # First condition: no unused files
    if len(notexist) <= 0:
        self.stdout.write('No unused files found')
        return

     # Second conditins: ask user to agree
     if input("Delete files? [Y/N]").lower() != 'y':
            self.stdout.write('Files are not deleted')
            return

    for subdir, dirs, files in os.walk(chosen_path):
        for file in files:
            if file in notexist:
                os.remove(os.path.join(subdir, file))

    self.stdout.write('Files are deleted')
```

Two specific thoughts about:

```python
        if pathtype == "other_media":
            chosen_path = os.path.join(os.path.join(settings.WRITABLE_FOLDER),settings.OTHER_MEDIA_DIRECTORY)
            for other_media_file in OtherMedia.objects.all().iterator():
                filenames.append(other_media_file.path.split("/")[-1])
        elif pathtype == "video":
            chosen_path = os.path.join(os.path.join(settings.WRITABLE_FOLDER),settings.GLOSS_VIDEO_DIRECTORY)
            for gloss_video in GlossVideo.objects.all().iterator():
                filenames.append(gloss_video.videofile.name.split("/")[-1])     
        else: 
            self.stdout.write("Wrong argument: try 'video' or 'other_media'")
            return
```

* This function seems to consists of 3 steps (find file names, figure out which are unused, and delete them), but only the latter two have a separate function... why not all of them for consistency?

* This approach avoids some repetition, what do you think?

```python
if pathtype == 'other_media':
    directory = settings.OTHER_MEDIA_DIRECTORY
    objects = OtherMedia.objects
else pathtype == 'video':
    directory = settings.GLOSS_VIDEO_DIRECTORY
    objects = GlossVideo.objects

chosen_path = os.path.join(os.path.join(settings.WRITABLE_FOLDER),directory)
for gloss_video in objects.all().iterator():
    filenames.append(gloss_video.videofile.name.split("/")[-1])                    
```

Jetske:
Great! I've changed most of the code according to your review. The only issue is with the last note to avoid repetition. I wanted to do that, but 'videofile' does not work for both video and other_media. For other_media you need to search for 'path' and for video to 'videofile.name'. But let me know if you know of another way.

susanodd:
Here is a screenshot of GlossVideo in the Admin. You see some filenames with ".bak" or ".bak1234" at the end. Although these are in GlossVideo and are on the filesystem, they are not used anymore in display. They are also inaccessible to the user.

glossvideo-admin-filenames

It would be welcome if these can be tracked down and removed as well.

vanlummelhuizen:
> Here is a screenshot of GlossVideo in the Admin. You see some filenames with ".bak" or ".bak1234" at the end. Although these are in GlossVideo and are on the filesystem, they are not used anymore in display. They are also inaccessible to the user.

@susanodd I believe these `.bak` files can become accessible again if the current video/other media file is deleted. So, they should not be removed from the DB and system.

vanlummelhuizen:
About this pull request: I am happy. So we could merge this as far as I'm concerned.

vanlummelhuizen (approved):

susanodd:
@vanlummelhuizen can you point out in the code where this is actually done?

vanlummelhuizen:
@susanodd 
> can you point out in the code where this is actually done?

See 
https://github.com/Signbank/Global-signbank/blob/8c3ae490e4084ff8e7df2a6ae8f35b658787b832/signbank/video/views.py#L57-L60
and
https://github.com/Signbank/Global-signbank/blob/8c3ae490e4084ff8e7df2a6ae8f35b658787b832/signbank/video/models.py#L331-L373

susanodd:
Okay, fine.

susanodd (approved):

Issue #843, opened 2022-09-08
Rename X_SENDFILE setting and put in default.py


Woseseltops:
X_SEND_FILE is a plugin for Apache that we use on the production server (it's about making sure images and videos can only be accessed when you are logged in). It's not needed when running Signbank locally with `develop.py runserver`, so we need a setting to turn it off.

Turns out, we already half-implemented this:

https://github.com/Signbank/Global-signbank/blob/9dd10ca2523aae0e261557cc36bb0da1e1725540/signbank/dictionary/views.py#L2667-L2676

Changes needed:
* Rename the setting to something cleaner, like `USE_X_SENDFILE`
* Add it to `default.py`, so everybody can see this settings exists... I would argue `False` by default

PR #844, opened 2022-09-09
X_SENDFILE renamed and added to default settings #843

Jetske:


susanodd:
This looks good to me.

vanlummelhuizen:
This is live now and works after I figured out that server_specific.py needed `USE_X_SENDFILE =True` :smile:

Issue #832, opened 2022-08-18
Move admin to something else than /admin


Woseseltops:
The Radboud Security Operations Center asks us to move the admin to another location than just `/admin`:

> Overweeg om de Django administrator login pagina te verbergen en niet default onder /admin te plaatsen.
Dit soort standaard URL’s zijn de eerste die gezocht zullen worden door hackers voor nader onderzoek en exploits.

While not hard to realize, it could be yet another source of confusion/frustration for new Signbank devs... other opinions wanted.

vanlummelhuizen:
Perhaps one common replacement for `/admin` for all TSG projects?

Other option: shield `/admin` for ip-addresses outside the RU network.

vanlummelhuizen:
> Perhaps one common replacement for `/admin` for all TSG projects?

We should be aware that hardcoding a alternative for `/admin` will still be public in the repo. Perhaps make it an instance specific setting.

Some interesting stuff: https://opensource.com/article/18/1/10-tips-making-django-admin-more-secure

Woseseltops:
> We should be aware that hardcoding a alternative for /admin will still be public in the repo. Perhaps make it an instance specific setting.

I like this option! This way, the default will still be /admin so we won't confuse new devs trying to install Signbank elsewhere, but for our own installation we can make it something else.

Having said that, that of course only postpones the problem, because our settings file is also in this repo. Thinking about this, I realize I've seen a problem like this multiple times at my other employer, and there it's always solved by creating a second `x-config` repo, like `Global-signbank-config`. This repo should be private, and as a free bonus would also be a good place to store issues related to the Lightning installation, a related problem that we've discussed before.

vanlummelhuizen:
For Mediate I use a script that reads variables form a `.env` file: https://github.com/Mediate18/Mediate/blob/master/mediate/decouple.py. The settings.py file contains things like

https://github.com/Mediate18/Mediate/blob/53b3b5a4fc41e736761c08b26ef53d0718d6e1a3/mediate/settings.py#L24-L30

Other options:

* https://django-environ.readthedocs.io/en/latest/
* https://pypi.org/project/python-decouple/

Woseseltops:
@Jetske , a separate config repo has been created in the mean time, but now the question is how this repo should work together with this one... could you make a proposal here? Do we need git submodules? Some sort of setup with symlinks? Something else?

vanlummelhuizen:
[...]

PR #846, opened 2022-09-23
Admin url changed to default or server specific setting #832

Jetske:


Woseseltops (approved):
Nice and simple solution... I don't think other reviews other needed, so I'll merge into master right away.

Issue #848, opened 2022-10-20
Change text on login page for registration


ocrasborn:
The registration option now says "Register for free to provide feedback on the Signbank." That's long outdated, users have better reasons now to get an account: to get access to datasets. Therefore, please change to "Register in order to get access to datasets"

Jetske:
Heb je daar misschien ook een Chinese vertaling van?

PR #852, opened 2022-11-03
 po files updated and text for reason to register changed #848

Jetske:


Woseseltops (approved):
Just PO files and textual updates; I don't think this requires more reviews ;)

PR #853, opened 2022-11-03
New FieldChoice Model and Model Translations

susanodd:
This is the Waypoint of branch Issue #658.

(Branch known as Fieldchoice field translations 658 661 merge)

This branch merges the new FieldChoice Foreign Key model of @vanlummelhuizen into the master, by hand.
It includes model translations for languages that used to be hard-coded, English, Dutch, Chinese.
Furthermore, Handshapes are now foreign keys. The code that references handshapes has been modified to accommodate this.

Numerous migrations exist that also copy data among other things.
Helper functions have been used to hide as much as possible to (temporary) new field names that end with '_fk' and '_handshapefk'. In some places the new field names are visible.

The followup to this will remove the old FieldChoice fields and rename the new fields.
The settings remain in tact and are used as often as possible (i.e., the helper functions to obtain the correct field).


vanlummelhuizen:
@susanodd @Woseseltops Do we want waypoints to be pull request of tags?

susanodd:
@Woseseltops asked for it to be a pull request so he can review it.

susanodd:
I made some slight revisions to the tests to remove obsolete code.

One very weird thing is this assertion in `testSettings.test_settings_field_choice_category`, that a phonology field has attribute `choices`:

https://github.com/Signbank/Global-signbank/blob/8a01ad7a22246861e68fee8f8391aba7d470852c/signbank/dictionary/tests.py#L2650-L2654

As pointed out in the comment, the fields actually all have a choices attribute.
Here is a print of this debug code:

`            for f in phonology_fields:`
`                # make sure all phonology fields in settings are defined in Gloss`
`                print(f, gloss_fields_names[f].__dict__)`

For (`NullBooleanField`) `Gloss` field '`repeat`' has a choices attribute, which is empty:

`repeat {'name': 'repeat', 'verbose_name': 'Repeated Movement', '_verbose_name': 'Repeated Movement', 'primary_key': False, 'max_length': None, '_unique': False, 'blank': True, 'null': True, 'remote_field': None, 'is_relation': False, 'default': False, 'editable': True, 'serialize': True, 'unique_for_date': None, 'unique_for_month': None, 'unique_for_year': None, 'choices': [], 'help_text': '', 'db_index': False, 'db_column': None, 'db_tablespace': '', 'auto_created': False, 'creation_counter': 193, '_validators': [], '_error_messages': None, 'error_messages': {'invalid_choice': 'Value %(value)r is not a valid choice.', 'null': 'This field cannot be null.', 'blank': 'This field cannot be blank.', 'unique': '%(model_name)s with this %(field_label)s already exists.', 'unique_for_date': '%(field_label)s must be unique for %(date_field_label)s %(lookup_type)s.', 'invalid': "'%(value)s' value must be either None, True or False."}, 'attname': 'repeat', 'column': 'repeat', 'concrete': True, 'model': <class 'signbank.dictionary.models.Gloss'>, '_get_default': <function Field._get_default.<locals>.<lambda> at 0x7f03342099d8>, 'cached_col': Col(dictionary_gloss, dictionary.Gloss.repeat)}`

susanodd:
@vanlummelhuizen @Woseseltops what to do, how to procede for the part after the waypoint?

Everything is working as it should. All tests pass.

I need to start removing and renaming fields. I prefer to spawn a new branch from the merge branch.

susanodd:
I'm proceeding on a new branch starting from this branch, which deletes and renames the fields.
`fieldchoice_field_translations_post_waypoint`

vanlummelhuizen:
@susanodd Did you test whether merging this into master, migrating etc. and testing also goes without problems?

susanodd:
I assume it's fine.
I didn't do it from a new copy of the master database because it takes extremely long to run and I can't do anything else while it's running. (The system is kind of unusable while it's running. It takes at least two hours. But perhaps longer.)
Would you be able to run it on your desktop computer, since you also have your laptop?
Yes, the idea is that it will be deployed first. All the tests pass.

vanlummelhuizen:
Ok, I will try to apply this to a recent backup from the live server and merge this into master.

Woseseltops:
Currently reviewing this, and I already want to add: wow, I underestimated how much cleaner this makes the code.

vanlummelhuizen:
@susanodd When trying to migrate on a live Signbank database backup I got the following error:
```
  Applying dictionary.0048_auto_20210602_1016...Traceback (most recent call last):
  File "bin/develop.py", line 11, in <module>
    execute_from_command_line(sys.argv)
  File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/sb-venv/lib/python3.8/site-packages/django/core/management/__init__.py", line 364, in execute_from_command_line
    utility.execute()
  File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/sb-venv/lib/python3.8/site-packages/django/core/management/__init__.py", line 356, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/sb-venv/lib/python3.8/site-packages/django/core/management/base.py", line 283, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/sb-venv/lib/python3.8/site-packages/django/core/management/base.py", line 330, in execute
    output = self.handle(*args, **options)
  File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/sb-venv/lib/python3.8/site-packages/django/core/management/commands/migrate.py", line 202, in handle
    post_migrate_state = executor.migrate(
  File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/sb-venv/lib/python3.8/site-packages/django/db/migrations/executor.py", line 115, in migrate
    state = self._migrate_all_forwards(state, plan, full_plan, fake=fake, fake_initial=fake_initial)
  File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/sb-venv/lib/python3.8/site-packages/django/db/migrations/executor.py", line 145, in _migrate_all_forwards
    state = self.apply_migration(state, migration, fake=fake, fake_initial=fake_initial)
  File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/sb-venv/lib/python3.8/site-packages/django/db/migrations/executor.py", line 244, in apply_migration
    state = migration.apply(state, schema_editor)
  File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/sb-venv/lib/python3.8/site-packages/django/db/migrations/migration.py", line 129, in apply
    operation.database_forwards(self.app_label, schema_editor, old_state, project_state)
  File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/sb-venv/lib/python3.8/site-packages/django/db/migrations/operations/special.py", line 193, in database_forwards
    self.code(from_state.apps, schema_editor)
  File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/repo/signbank/dictionary/migrations/0048_auto_20210602_1016.py", line 23, in copy_handshape_values
    gloss.final_domhndsh_handshapefk = handshapes[gloss.final_domhndsh_fk.machine_value]
KeyError: 0
{"id":168,"appid":2,"message":"Migration done","title":"Signbank","priority":5,"date":"2022-11-14T08:37:54.692592644Z"}
```
Do you know  how to solve this?

vanlummelhuizen:
I looked into in this some more. The `FieldChoice: Handshape: - (0)` and `FieldChoice: Handshape: N/A (1)` are not copied from FieldChoice to Handshape so that Handshape with machine_value 0 or 1 do not exist and cannot be used for e.g. Gloss with 3107.

I think https://github.com/Signbank/Global-signbank/blob/fieldchoice_field_translations_658_661_merge/signbank/dictionary/migrations/0039_auto_20200910_1020.py is not doing what I expect: it should make the two default FieldChoices (0 and 1) and then copy them to Handshape among others.

I will try to fix this.

susanodd:
Some of this has to do with the fact that the migrations were being applied "while" the branch was developing.
The thing with the empty choices was solved in a later migration.

I ended up debugging the code to work with the new model after migrations has been applied.
A bit of chicken or the egg.

vanlummelhuizen:
@susanodd I see that https://github.com/Signbank/Global-signbank/blob/fieldchoice_field_translations_658_661_merge/signbank/dictionary/migrations/0052_auto_20221024_1633.py creates the default values for SemanticField and DerivationHistory, similar for FieldChoice in https://github.com/Signbank/Global-signbank/blob/fieldchoice_field_translations_658_661_merge/signbank/dictionary/migrations/0052_auto_20221024_1633.py.

Is there a similar thing for Handshape that you know of?

susanodd:
Yes, there is also a migration that does this for handshapes.

In your original branch from which everything has been incorporated, there was one "commit" that merged from a different branch.
But after that, some migrations seemed to have disappeared. (That other branch was already using handshape objects.)

It should be in one that uses get_or_create

susanodd:
It was unclear to me if we want "empty" handshapes or just use the Null value.
The user is allowed to edit the handshapes.
The handshape list view only shows the ones with machine values greater than 1.
But the searches show all of them.
The querying checks for both when the glosses with empty handshapes need to be ignored.
This is used to identify when the phonology hasn't been filled in (there was previously an issue to not include them).

vanlummelhuizen:
@susanodd I now have done this:

```diff
diff --git a/signbank/dictionary/migrations/0048_auto_20210602_1016.py b/signbank/dictionary/migrations/0048_auto_20210602_1016.py
index b09970a9..75d92d65 100644
--- a/signbank/dictionary/migrations/0048_auto_20210602_1016.py
+++ b/signbank/dictionary/migrations/0048_auto_20210602_1016.py
@@ -5,6 +5,28 @@ from __future__ import unicode_literals
 from django.db import migrations
 
 
+def add_default_handshapes(apps, schema_editor):
+    """
+    Add 0 and 1 choices to every Handshape category
+    :param apps:
+    :param schema_editor:
+    :return:
+    """
+    Handshape = apps.get_model('dictionary', 'Handshape')
+
+    new_handshape_0, created = Handshape.objects.get_or_create(name='-', machine_value=0)
+    if created:
+        new_handshape_0.name_en = '-'
+        new_handshape_0.name_nl = '-'
+        new_handshape_0.name_zh_hans = '-'
+        new_handshape_0.save()
+
+    new_handshape_1, created = Handshape.objects.get_or_create(name='N/A', machine_value=1)
+    new_handshape_1.name_en = 'N/A'
+    new_handshape_1.name_nl = 'N/A'
+    new_handshape_1.name_zh_hans = 'N/A'
+    new_handshape_1.save()
+
 
 def copy_handshape_values(apps, schema_editor):
     Gloss = apps.get_model('dictionary', 'Gloss')
@@ -33,5 +55,6 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
+        migrations.RunPython(add_default_handshapes),
         migrations.RunPython(copy_handshape_values),
     ]
```

susanodd:
Probably in forms, for this branch, in order for it to work without modifying the settings, we need to hard-code that line in forms that references the LANGUAGES (see above), and just hard-code name_... instead. Then use it as is in the post waypoint.

susanodd:
Some of the settings tests check that the settings fields include choices. I revised these tests in the post-waypoint code since the field_choice_category is missing. It was still making use of that attribute sometimes. That's basically what a lot of the post-revisions are about.

susanodd:
@Woseseltops question about LANGUAGES.

The setting MODELTRANSLATION_LANGUAGES is used for the new name fields for field choices.
This setting is in global_sb_local.py but not in default.py.

In order to elegantly "exclude" the new languages in this branch that do not yet have model translation fields, but are in the setting LANGUAGES, it would be most elegant to iterate over LANGUAGES and check whether the language is in MODELTRANSLATION_LANGUAGES. (Namely, in the forms.py FieldChoiceForm.)

Is there any reason not to include this in default.py ?
It would be equal to just ['en'].

Woseseltops (commented):
I went over every single change in all 59 files that contained changes, but I have only a few minor things to complain:

* In `adminviews.py`, I notice these two lines follow each other 10 times
```python
default_language = Language.objects.get(id=get_default_language_id())
default_language_code = default_language.language_code_2char
```
And in 7 out of those 10 times, it's followed by these 3 lines
```python
interface_language_3char = dict(settings.LANGUAGES_LANGUAGE_CODE_3CHAR)[self.request.LANGUAGE_CODE]
interface_language = Language.objects.get(language_code_3char=interface_language_3char)
interface_language_code = interface_language.language_code_2char
```` 
I think we should be wetter here :) (DRY principle)

* And if we're making things wetter, in `adminviews.py` I also saw these lines 2 times:

```python
        for category in choice_categories:
            if category == 'SemField':
                field_choices = SemanticField.objects.all()
                print('SemField field_choices: ', field_choices)
            elif category == 'derivHist':
                field_choices = DerivationHistory.objects.all()
            elif category == 'Handshape':
                field_choices = Handshape.objects.all()
                print(category, field_choices)
            else:
                field_choices = FieldChoice.objects.filter(field__iexact=category)
```

* In `adminviews.py`, I feel like this line could be written more reader friendly, perhaps in more steps with clear variable names:
```python
        order_dict = dict([(v, i) for i, v in enumerate(
            list(FieldChoice.objects.filter(field=field_choice_category, machine_value__lt=2)
                 .values_list('id', flat=True))
            + list(FieldChoice.objects.filter(field=field_choice_category, machine_value__gt=1)
                   .order_by('name').values_list('id', flat=True))
        )])
```
* I notice 4 new functions have been added to `tools.py`. Because `tools.py` has grown into such a big ball of unrelated functionality, perhaps this would be a good moment to put these new functions into separate files, called something like `field_choices.py`?

* In `models.py` , the try-except around this line was removed. I think it was there so a new installation of Signbank could run without having SignLanguages defined, what is the reasoning behind removing it?
```python
try:
      signLanguageChoices = [ (0, '---------') ] + [ ( sl.id, sl.name ) for sl in SignLanguage.objects.all() ]
except (OperationalError, ProgrammingError) as e:
      signLanguageChoices = []
```

* In the ASL settings, `LANGUAGES_LANGUAGE_CODE_3CHAR` is added with value `('en', 'eng')`. However, the ASL Signbank uses 'American English' (en_US) as its interface language... will this still work?

* In a few places, most notably `update.py`, I see commented out debug prints. While they are nicely documented, so it's clear why they are there, wouldn't it be prettier to add an extra debug setting so you can toggle these print statement on and off?

* I see the variable name `morpheme_toggle`, which to me feels like a reference to a UI element. How about `is_morpheme`? Or if you really want to make clear that it is a morpheme OR a gloss, a descriptive variable name could be `is_morpheme_not_gloss`. Other programming languages might use an enum here, which is also possible in Python (but just uncommon)

* Related to the previous comment, there is also `is_sign_or_morpheme`... if this is not the same thing, I think it should be made clearer what is the relationship between the two

* There's still an unused print statement in `signbank/dictionary/templatetags/annotation_idgloss_translation.py`

susanodd:
@Woseseltops some assumptions about languages now in question.

I used the 3char abbreviation setting LANGUAGES_LANGUAGE_CODE_3CHAR because it was without the hyphen to lookup. Some of the code broke with all the conversions between "-" and "_" and it wasn't correctly obtaining the result for zh-hans.
Hence this setting.
(Incidentally, this is also the reason some of the extra assignment statements involving languages and codes, to force eager evaluation.)

I tried to find the 3-char abbreviation for en_US. With help from a colleague, could it be "ame"  (AmE)?
en_US isn't included in the dictionary_language table for Global.

I'm not sure about the removal of the try except from SignLanguage (above). This class exists in the code.
I assume if the class exists, it will just be empty. But if the class doesn't exist, that's a compile-time error not a run-time error.

susanodd:
@Woseseltops this is what making a migration does when 'en-us' is added to the model translation languages:

`    - Add field name_en_us to derivationhistory`
`    - Add field name_en_us to fieldchoice`
`    - Add field name_en_us to handshape`
`    - Add field name_en_us to language`
`    - Add field name_en_us to semanticfield`

susanodd:
@Woseseltops related to the above:

https://github.com/Signbank/Global-signbank/blob/d210723f86b55574c4fd2761a8e62367a1ac6073/signbank/dictionary/forms.py#L1078

This is the code that potentially causes problems.
(This remains also in the post-waypoint branch.)
While it works to just use "name" and let Django fill in all the fields for the model translations in Admin.
The tests do not work: If only "name" is used, the "translations" end up not included in the form by the time it tries to save a field choice object.
But it also does not work to use "both" "name" plus the "translations".
Only one or the other can be used in the form.
But only the model translation variant works for tests.

So for ASL, this will either need to be using "name_en_us" (parametric as above) or else using "name" (change code).
If you want to do minimal code modification to have it work for ASL, then probably allow the 'en-us' as the only model translation.

susanodd:
The morpheme_toggle is in Feedback.
This illustrates very poor coding. My bad.

For example, the field "like" of SignFeedback is actually being used as a tag for whether this is a morpheme or not.
Currently nothing is being done with Feedback. It was in the past resurrected from the Auslan code and the fields reappropriated for Global instead of doing a proper migration. Hence the reuse of the field name for something else.)

We decided it would be best to make a new issue for Feedback rather than spend time modifying it in this Field Choice / hard-coded language fields issue. Feedback contains minimal changes to make it work with the new FieldChoice model. Unused fields have been hidden from the Admin.

susanodd:
@Woseseltops about these lines:

`interface_language_3char = dict(settings.LANGUAGES_LANGUAGE_CODE_3CHAR)[self.request.LANGUAGE_CODE]`
`interface_language = Language.objects.get(language_code_3char=interface_language_3char)`
`interface_language_code = interface_language.language_code_2char`

This was to avoid the bug converting the 'zh-hans' code to something that could be looked up.
The interface language codes do not appear in the Language table.
This code looks hideous. It was meant to hide the "-" and "_" string manipulation.

susanodd:
Okay the tests still work after the above revisions.

susanodd:
The revisions as per review have sucessfully been applied to the post-waypoint branch.

susanodd:
@Woseseltops @vanlummelhuizen I added a necessary setting and reference for model translations.
This is needed to avoid crashing Global because there are languages which are not yet in model translations.
(See commit.)
(This is also done on the post-waypoint.)

susanodd:
**try:
      signLanguageChoices = [ (0, '---------') ] + [ ( sl.id, sl.name ) for sl in SignLanguage.objects.all() ]
except (OperationalError, ProgrammingError) as e:
      signLanguageChoices = []**

This code doesn't exist in the post-waypoint.

susanodd:
This branch is okay now.

susanodd:
The post-waypoint branch is ready.

Woseseltops (approved):
Great. I don't really understand what American language thing, but I'll do a deepdive once we will upgrade the ASL Signbank.

PR #856, opened 2022-11-17
Drag and drop uploading added for editing a gloss #975

Jetske:


Woseseltops:
Hey @Jetske , I noticed there are quite a few lines of Javascript. Would you say it makes sense to put them in a separate .js file and import it, to prevent `gloss_detail.html` from growing too large?

Jetske:
@Woseseltops  Yes, that would make sense! I've changed it in the new commit. Just to be clear, the javascript will then be added to the general media/js folder with all other .js files (so not in the dictionary app).

Woseseltops (approved):
Perfect

Issue #850, opened 2022-10-20
Send dataset manager a notification when a new user has been given access


ocrasborn:
This is a follow-up to #849 : instead of sending an email to notify the admin (or dataset manager?) that a user is asking for permissions to see the dataset, send a notification that a new user has been given access.
Again, this pertains both to new users just signing up and to existing users getting access to additional datasets.
The subject of the new mail should read "A new user has been given access to dataset <name>"
Feel free to compose the text of the message yourself, but make sure to include the full name of the user, and the 'motivation' that was filled in in the request form.

Issue #849, opened 2022-10-20
Change registration and dataset access procedure


ocrasborn:
When new users register, they can apply for access to a dataset. Similarly, when users already have an account, they can apply for access to more datasets. When they do, a notification email is sent to... the dataset manager... or an admin? (not sure, because i'm both). I then have to go to Datasets > Manage Datasets to give the user access. Please remove this manual intermediate step, immediately giving users access when they apply for it, for any dataset for which the property "accessible by others" is True (I don't know where this is in the admin, this is what it's called on a dataset's detail page in the interface).

PR #857, opened 2022-11-17
Immediate dataset access procedure #849 and #850

Jetske:
Needs to be reviewed! 

I _think_ it works alright, tested it by checking whether access was granted and by looking at the emails that are sent when trying to access new datasets as regular user. But I did not test the actual send_mail command. 

From how I understand it, requesting access to private (not "accessible by others") datasets should not be possible because the buttons for that are hidden in the menu 'datasets > view available datasets'. However, if this does happen somehow, a request email is supposed to be sent to the owner. 

And lastly, I couldn't figure out the difference between dataset owners and dataset managers and assumed the words are used interchangeably. Correct me if I'm wrong.


susanodd:
There exist test users that @ocrasborn created that belong to different groups. (I'm not sure if he knows the passwords anymore.)

In the code you'll see places where print statements (commented out) exist near the send_mail. 
Also it's possible to create a test for this and grant the user specific permissions or assign them to specific groups.
There are some tests for datasets that do this. (With the print statements instead of the send_mail.)

Woseseltops:
> From how I understand it, requesting access to private (not "accessible by others") datasets should not be possible because the buttons for that are hidden in the menu 'datasets > view available datasets'. However, if this does happen somehow, a request email is supposed to be sent to the owner.

It think this outside the scope of the current issue, but indeed, when a new user X is created, X can only see public datasets. Not 100% sure, but it might be possible that an admin user adds X to another user group with more permission, and then X might be able to see and request access to all datasets.

Woseseltops (commented):
Good work. I like that you:
* Nicely divide the code into 'paragraphs', each with their own explanation comment
* Thought of changing the actual email prose :)

Minor complaints:
* I see there a few commented debug print statements. They are nicely explained, so it's clear why they are there... but perhaps it would be cleaner as a setting? So you can all turn them on and off in one go.
* Still an (active) print statement in `adminviews.py` ;)

Jetske:
Thanks for checking @Woseseltops !
I just added the debug option as setting to default.py, not sure if it is cleaner but it is definitely faster to activate this way

Issue #847, opened 2022-09-30
Allow users to set their Interface Language inside of their Profile


susanodd:
Problems with the PyCharm emulator make it impossible to "select" language icons in the pull-down.

Although it's possible to do this via the virtual environment interface, this is extremely annoying. My computer is switching between NL and Chinese. I can't "select" the EN from the list.
To solve this on my own PyCharm emulator, I removed Chinese from the list (temporarily) in order to have only two languages in the list to be able to select the "other" one.

susanodd:
This is probably unnecessary. It was originally proposed because of a glitch in the PyCharm Runserver.

PR #860, opened 2022-11-24
Fieldchoice field translations POST WAYPOINT

susanodd:
This branch is the followup to the merge branch. It finishes #658.
It removes the _fk and _handshapefk fields, removes helper functions and code that referenced or hid these.

The semantic field and derivation history many to many fields have replaced the original field choice fields.
This had some consequences for other code since the many to many fields are treated differently than other fields of Gloss.
They are relations rather than fields. Some code has been revised because of this.

Model translations have been added for the two new languages.
The tests have been re-written to use the new names of fields and are parametric wrt languages.

As per the review of @Woseseltops of the waypoint (#853), many loops have been rewritten to avoid excessive typed querying of different models. (The wet/dry principle.)


vanlummelhuizen:
I executed the migrations on a recent backup of the live db. First on the waypoint, then on this pull request. All went well and also the tests did not ring alarm concerning this pull request.

However, when I switch to NL as interface language on http://localhost:8000/dictionary/gloss/3117, I got

```
Traceback:

File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/sb-venv/lib/python3.8/site-packages/django/core/handlers/exception.py" in inner
  41.             response = get_response(request)

File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/sb-venv/lib/python3.8/site-packages/django/core/handlers/base.py" in _legacy_get_response
  249.             response = self._get_response(request)

File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/sb-venv/lib/python3.8/site-packages/django/core/handlers/base.py" in _get_response
  187.                 response = self.process_exception_by_middleware(e, request)

File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/sb-venv/lib/python3.8/site-packages/django/core/handlers/base.py" in _get_response
  185.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/sb-venv/lib/python3.8/site-packages/django/views/generic/base.py" in view
  68.             return self.dispatch(request, *args, **kwargs)

File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/sb-venv/lib/python3.8/site-packages/django/views/generic/base.py" in dispatch
  88.         return handler(request, *args, **kwargs)

File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/repo/signbank/dictionary/adminviews.py" in get
  1285.         context = self.get_context_data(object=self.object)

File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/repo/signbank/dictionary/adminviews.py" in get_context_data
  1302.          default_language, default_language_code) = get_interface_language_and_default_language_codes(self.request)

File "/vol/tensusers3/mhulsbosch/projects/signbank/source/signbank_review/repo/signbank/tools.py" in get_interface_language_and_default_language_codes
  2048.     interface_language_3char = dict(settings.LANGUAGES_LANGUAGE_CODE_3CHAR)[request.LANGUAGE_CODE]

Exception Type: KeyError at /dictionary/gloss/3117
Exception Value: 'zh-hans'
```

@susanodd any idea?

susanodd:
Did you modify your local settings?

susanodd:
@vanlummelhuizen See #847. I had trouble with choosing the language from the pull-down (menu bar) when running via PyCharm runserver. I commented out languages in my local settings to force the menu bar to show only two languages, otherwise it ignored the middle one and I couldn't grab it with the cursor. It's weird that it's looking for zh-hans instead of nl. But it works for me with all five languages. Did Python 3.8 change the interface abbreviations? Ot the dict() coercion?

Woseseltops (commented):
Only two things:

* Is `build_choice_list` in `models.py` still used? Because I noticed a reference variable as default argument:

```python

def build_choice_list(field, list_start=[]):

```

Reference vars as default arguments are not recreated each time the function is run. For example, this...

```python
def bite(l=[]):

    l.append('wessel')
    return l

bite()
print(bite())
```

... will print `['wessel','wessel']` .


* I noticed the functions in `field_choices.py` are long lists of instructions. I think it would help for readability if they could be split up in 'paragraphs' of 3-8 lines of code, each with its own comment explaining the paragraph? Or simply put: this needs more empty lines and comments :)

susanodd:
Great, @Woseseltops, thanks for the review.
That is weird for the build_choice_list. It's not used anymore. I missed that one, so did @vanlummelhuizen !
build_choice_list was being used for type checking in the old fields.
Something similar may be used in some form fields. (need to check) The build_choice_list was only used in the declarations (of the old fields). It had to be just a dummy function because they needed to be choices to type check.)

I'll see if other places have problems with the argument default reference vars like that. Good find!

I'll work on the functions in field_choices.py. Yes, these are very difficult to read. I agree!
I had several goes just getting the function get_frequencies_for_category to take parameters.
It replaces really awful code that was in FieldChoiceView. But it needs comments. As do the others.

https://github.com/Signbank/Global-signbank/blob/5a14e5d8d39cf5a331d43f11c4c05bf1e1a613a1/signbank/dictionary/field_choices.py#L108

vanlummelhuizen:
> Did you modify your local settings?

@susanodd Yes, I did. I added the `LANGUAGES_LANGUAGE_CODE_3CHAR` dict and there is no error anymore. The modeltranslation is now working as expected. But now I see that the label of the interface does not switch to Dutch or Chinese, but remain English. 

If you have an idea, please tell me. In the meantime I will also try to find what is going on.

vanlummelhuizen:
> But now I see that the labels of the interface does not switch to Dutch or Chinese, but remain English.

OK, my bad. I did not have the file containing the compiled messages in my instance. Now I have and all is ok.

PR #862, opened 2022-12-02
Sign app api

TobiasCunnen:
Misha pointed out that the last pull request had code that was not mine pushed with it. This was an old function that was deleted and for some reason git tried to add that code again with my push. 

Tried to fix it but only made it worse so now a new pull request that should be correct. There is still one line in the urls.py that I did not add but that line seem to still be there in master so I left it like it is. (also because I am not sure how to fix it).

I tagged everyone as reviewers because Wessel did that last time. Feel free to leave comments or talk to me directly.


Jetske (commented):
Looks good!

Minor remarks:
- In gloss_api (views.py) and tests, some more comments would be helpful to follow along
- It's slightly better practice to use 'path.join' than '+' to make paths

Woseseltops (commented):
Looks good overall! I have some complaints about things you probably couldn't have known ;)

* I see you use guard clauses, but I notice these clauses themselves are inside `if request.GET`. Mybe this would be a good guard clause as well, so just do:

```python
if not request.GET:
    return HTTPResponse(...)
```

and the rest of the guards can be non-indented.

* I think 

```python
print('Maximum number of results is not set so default to 100')
```

should return a HttpResponse, like the other guards.

* Functions starting with `_` are not meant to be called directly in Python. Instead, the agreement is to do `str(gloss)`

* Your main iteration could be a short list comprehension (fancy Python feature). Something along the lines of:

```python
response = [{'sign_name': str(gloss), 'video_url': gloss.get_video_url(), 'image_url': gloss.get_image_url()} for gloss in glosses if gloss.get_video_url()]
```

Woseseltops (approved):
Now that I look back at the list comprehension I proposed, I wonder whether it really improves maintainability ;). I'll approve and merge anyway.

Issue #855, opened 2022-11-14
Make WRITABLE_FOLDER in create_test_db not hard coded


vanlummelhuizen:
https://github.com/Signbank/Global-signbank/blob/master/signbank/dictionary/management/commands/create_test_db.py has WRITABLE_FOLDER.

We better make this use the setting WRITABLE_FOLDER.

vanlummelhuizen:
@Jetske Is this something you could do? (@Woseseltops )

susanodd:
Good idea. I keep needing to edit this. I've been making lots of test databases in the past months.

Woseseltops:
Yeah @Jetske was interested in tests, this is a good opportunity. Also I apologize for this bad decision.

Jetske:
@vanlummelhuizen  :+1: I think I could do this

Jetske:
![image](https://user-images.githubusercontent.com/10157482/205011765-857ecebf-d4d1-495a-91b3-deafac2e7a0e.png)

It's possible with the setting WRITABLE_FOLDER but why not use settings.DATABASES? This works for me locally. 

(it was:)
![image](https://user-images.githubusercontent.com/10157482/205012210-d22a9166-cdd4-42db-b61a-290ddb9a9e0e.png)

vanlummelhuizen:
@Jetske That seems sensible. Go for it.

PR #863, opened 2022-12-05
Database and test database location not hardcoded anymore #855

Jetske:


Woseseltops (commented):
Small style suggestion: UPPERCASE is for static hardcoded things. However, you're loading in data dynamically from the settings, which I would say would justify used 'normal' variable names.

Jetske:
It was already in uppercase so I didn't think about changing it, but fair point. It's all lowercase now.

Issue #859, opened 2022-11-18
Prefill dataset field when only one dataset is selected


ocrasborn:
To speed up data entry, when creating new signs or new morphemes, please pre-select the selected dataset if there's only one. It's silly to ask the user to select a single option from a menu where only one option is available. See screenshot for adding a morpheme:
<img width="720" alt="Screenshot 2022-11-18 at 16 20 21" src="https://user-images.githubusercontent.com/10242207/202738925-68e07c59-cbe8-41ff-b966-ffe97794d800.png">

susanodd:
There's a bit of inconsistency in the last used dataset type.

susanodd:
This issue probably needs to wait until the post-waypoint.
I started fixing it on master, except there are too many inconsistencies in the last_used_dataset type being passed around.
Sometimes it's None. Sometimes it's a dataset object, sometimes it's a dataset name, sometimes it's a dataset acronym.
The templates seem to use this willy nilly.

susanodd:
This has been pushed and is awaiting review and deployment.

Issue #866, opened 2022-12-09
Show all shows nothing when user is anonymous


susanodd:
This bug came up in #865.

susanodd:
This is fixed now. I'm not sure if this is the same as issue #865 or not, thus a new issue.

susanodd:
I'm fixing some glitches with the sort order in the list view in the different templates.

susanodd:
This has been pushed and is awaiting review and deployment.

susanodd:
This is working now.

Issue #865, opened 2022-12-08
Searching not possible if interface language is not English


ocrasborn:
Just found out that I cannot search in the Dutch interface anymore; tested Chinese, same thing.

susanodd:
I tried it just now and it works for both.
(Try logging in again in case the browser is using old javascript ?)

The new field choice model and model translations were deployed yesterday by @Woseseltops .
It could be that you tried before it was finished?

susanodd:
Ahha. I just tried being not logged in and using the select datasets for a not logged in user.
(For some reason, Kata Kolok also shows up as a possible choice. But it only has a couple of public glosses.)
If as a not logged in user I click on Show All Signs, it doesn't show any.
But if I'm logged in it does show them.
This could have something to do with the permissions checking.
(It could also have something to do with the last used dataset. Some code was using name and other code was using acronym and other code was using dataset objects -- which get mapped to id's.)
The users last used dataset is stored in the session. But the selected datasets are stored in the user's profile.
So it could be that there is a type mis-match causing the "selected datasets" to not work properly.
(I'm trying to duplicate the error to decipher what is causing it. It seems intermittent.)

susanodd:
@ocrasborn were you logged in when you attempted searching?

I found a bug related to permissions of anonymous users.
They are apparently not allowed to search by publication and notes.
However this is missing in the template parameters so the javascript fails before completion (trying to generate the Search by Publication panel) and no results are shown.

susanodd:
I fixed the problem just described for anonymous users and the publication panel.

@ocrasborn can you check whether it works okay today if you're logged in?

susanodd:
Hopefully this is fixed now. (Not deployed yet.)

susanodd:
I pushed many revisions to the code. Hopefully this will resolve this issue.
I'm only able to mimic this issue intermittently. Sometimes I can duplicate the error. But then it seems to resolve itself.
The gloss list view didn't work at all for anonymous users. That has been fixed. Plus there were inconsistencies with the last used dataset that is stored in settings. Since all the searches depend on the selected datasets and as a side effect, the last used dataset, hopefully those repairs will fix it.
The changes are awaiting review and deployment.

susanodd:
On master branch on my **local** machine, if I have tstMH as the selected dataset and the interface language set to Chinese, then search on Handedness is 2a, then I get 8 results:
[...]

Issue #851, opened 2022-11-01
Wrong CSV export for 'show all signs'


ocrasborn:
When I try to export a CSV file of all signs in a dataset (by selecting Show all signs from the Signs menu), I do not get the full dataset (nor a broken pipe error #432), but I get a CSV file with the results of the previous search. (Only tried this out with gloss searches.)

susanodd:
This is solved in the field choice branches.

susanodd:
Okay, this appears to be broken!
But now it's worse.

susanodd:
This is awaiting review and deployment.

vanlummelhuizen:
@susanodd @ocrasborn The merge-commit for this issue is on the live server. Could you try it out and if alright, close this issue?

susanodd:
This works now.

ocrasborn:
I'm afraid not, @susanodd , still the same problem.

susanodd:
This is working again, but needs a bit more testing to check all the possible workflows.

susanodd:
This has been pushed to master.

To test the solution, I tried repeating a query (with phonology) then export csv, then another query with show_all, then export csv, repeat, repeat, repeat.
All times sucessful in both cases.

susanodd:
Awaiting deployment.

susanodd:
This should be working now @ocrasborn .

PR #867, opened 2022-12-14
Miscellaneous issues related to searching templates

susanodd:
This branch fixes numerous glitches.

- Fixes issue #859: Prefill dataset when only one dataset is selected for creating glosses and morphemes. There are many entry points to add glosses/morphemes: via the pull-down menu bar and inside of the gloss and morpheme list views; plus a create lemma inside of lemma list view. All of these have a dataset select.

- Related to "Last used dataset", the code has been fixed to use a consistent representation. There had been a mix of "dataset objects", "dataset name", and "dataset acronym" used in various routines. These have all been adjusted to use "acronym". In code where the selected datasets are referred to, the acronym has been retrieved. (The last used dataset session variable is related to #859.)

- Added modals for related sign and related foreign sign deletion in Gloss Detail View. The formatting of the trash icon confirm/canel popup when clicked upon was appearing _behind_ the new drag and drop videos. #675.

- Major revision in functions in update.py to move the simple cases to the front of the functions, to remove levels of indentation. So-called Happy code. Related to #759, some of the functionality in update that checks whether the request method is POST and checks whether the form is valid, might actually be code that belongs in forms. These are functions that update the data in the various models. @Woseseltops please suggest.

- The search/show all signs for anonymous users had a bug in it because they do not apparently have permission to view the publication and notes data. That was inconsistently being checked by setting it to empty. But that caused a bug in the javascript for multiselect fields. This has been fixed. (#866) (This bug behaved the same as this when the user is anonymous: #865.)

- Morpheme list view had the abstract meaning columns only for the dataset of the morpheme (of the row). This was causing the columns to be misaligned. This has been repaired so the all possible languages are in the data structure for keywords (abstract meaning), with empty values for those languages not in the morphemes translation languages. This makes the looping in the template cleaner. (#639)

- Sorting on handedness was not working for (e.g.,) morphemes. This is partly caused by creating test morphemes with no data. If the handedness was not set to anything (i.e., None in the database), this was causing a crash. The sorting has been revised to account for possible None values by not looking up the machine value. (This bug only appears if glosses or morphemes have been created that lack the minimal fields shown in the display list view. The new field choice model has field choice objects for the machine values 0 and 1. But it is still possible to have glosses or morphemes with a None value in the fields.) (Related to #639)

- Some obsolete templates and urls have been removed related to preview of colors. This needs to be adjusted in Settings after deployment. The two list views for semantic fields and derivation histories appear under Settings. But these are actually urls under dictionary, not under Settings. They have been moved to be with the other Settings. Once this is deployed, the old urls can be removed. (This concerns settings related to #762, #726, #727)

- Also fixed the display of Handshapes, Semantic Fields, and Derivation History in the CSV export for Glosses and Morphemes. The display was broken for #851.


susanodd:
I asked for review from @Woseseltops because the modifcations to the Happy case code in `update.py` is related to previous reviews for code improvement.

Woseseltops:
Comment before I start reviewing: although it would have been good if these were smaller pull requests, I do appreciate the structured overview of what I'm going to review 👍

Woseseltops (commented):
Not much to complain! I do note that a lot of the guard clauses you (beautifully) introduced contain these two lines:

```python
messages.add_message(request, messages.ERROR, _("Please provide a dataset."))
return render(request, 'dictionary/add_gloss.html', {'add_gloss_form': form,
                                                           'dataset_languages': dataset_languages,
                                                           'selected_datasets': get_selected_datasets_for_user(request.user),
                                                           'SHOW_DATASET_INTERFACE_OPTIONS': settings.SHOW_DATASET_INTERFACE_OPTIONS})
```

I'd rather see something like

```python
show_error(request, _('Please provide dataset'), form, dataset_languages)
```

Issue #93, opened 2015-10-02
Add two fields to table 'Relation to foreign signs'


ocrasborn:
1) Foreign translation equivalents (field: other_lang_transl) 
2) Link (field: other_lang_link)

...and add those to the table in the detail view and to the search panel Relations to foreign signs. To make some space in the table, "Gloss in related language" can be cut to "Gloss" (translation "

vanlummelhuizen:
@ocrasborn is this issue still relevant?

ocrasborn:
yes, it is @vanlummelhuizen

susanodd:
See (2) in Issue description.
@Woseseltops what about the link url field wrt to the database?

susanodd:
Here is a mockup showing the two extra fields (columns) to be added. (To see what this would look like.)

![gloss-related-signs-foreign-signs](https://user-images.githubusercontent.com/24272544/207890925-182e3877-e478-4c45-9c8b-ce1573780bfc.png)

susanodd:
The revised Relations View is awaiting review.

susanodd:
Some time ago, the two new fields were added to the Relations View for a Gloss.
But they are not yet added to the model.

Bullet (1) of the issue is Translation Equivalents. This is perhaps relevant for Senses.

http://localhost:8000/dictionary/gloss_relations/4619

vanlummelhuizen:
@susanodd What should we do with this issue?

susanodd:
I'm not sure. It's easy to just add the fields. It probably just needs to mention senses.

I didn't implement it except in the template because at the time it would have involved a migration.
We'll assign it to @ocrasborn .

ocrasborn:
Haha, yes, let him decide! ;-) @susanodd , no complex senses needed here for 'foreign language translations' here. Adding a URL would still be helpful. But let's make this neutral priority, focusing on the senses first.

PR #870, opened 2022-12-22
Issue #703, Issue #93: Buttons in Frequency View translatable

susanodd:
This branch makes small modifications to translate the buttons in the Frequency View, as well as adds Relations to Foreign Signs to the Relations View. (The two new columns are added, but not implemented.)

- Buttons in Frequency View translatable

- Relations to foreign signs in Relations View table.

### DIFF
--- a/conf/locale/nl/LC_MESSAGES/django.po
+++ b/conf/locale/nl/LC_MESSAGES/django.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-10-27 10:43+0200\n"
+"POT-Creation-Date: 2022-12-22 14:45+0100\n"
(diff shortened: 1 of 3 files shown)


Woseseltops (approved):
Looks fine to me. 

One minor thing (not necessary to fix): I see you putting translatable strings in a Javascript variable. I think the beauty of Django's localization system is that you don't have to summarize UI strings into a variable name, you can just copypaste the string itself, and Django will figure out it needs the same translation:

In other words, you replaced this

```javascript
 $(e).html("Hide empty frequencies");
```

with this

```javascript
 $(e).html(html_button_hide);
```

but I think you could also have done this:

```javascript
 $(e).html("{% blocktrans %}Hide empty frequencies{% endblocktrans %}");
```

For strings in the UI I would say I prefer clarity over the advantages of being DRY, and Django handles this well.

susanodd:
I had problems figuring it out at the time. The translations kept sometimes being translated and then not being translated.

Issue #150, opened 2016-03-09
Create new record by duplicating an existing one


ocrasborn:
Sometimes a new form has to be entered that is highly similar in e.g. phonology to an existing form. It would be convenient if we would have a 'duplicate sign' function. By a separate button in the detail view, or by a menu item under Sign? The gloss could be SOURCEGLOSS-duplicate, for instance, reminding the user to alter it right away.

susanodd:
@ocrasborn do you want this to be restricted to the same dataset? Perhaps, not.

susanodd:
This is nearly done.

susanodd:
This issue has become related to #851 and #544.
Creating a new gloss from another needs to fill in the new fields.
I noticed that if something is left blank in the new copy gloss, it breaks the gloss list (hence CSV) and the sorting.
This is because those rely on fields not being empty.
I coded the copy-gloss function at the object level rather than via a form. While working on this, I accidentally created a bunch of gloss copies on my local signbank. Then these ended up inaccessible, but the software doesn't actually complain. We have assumed that the user will fill in the required fields themself.
I started out using the "-duplicate" suffix as shown above.
But this will also best make use of the "copy annotation fields to lemma fields" as requested by @ocrasborn in #869.
Then things won't be left empty by accident.

susanodd:
@ocrasborn is there any desire to be able to use this functionality cross-dataset?

At the moment, I've added a "Copy" button to the Gloss Detail View.
That has the effect of 

1. creates a gloss as described in the issue (with -duplicate suffixes in the annotations).
2. copies the phonology fields to the new gloss
3. copies the dataset by way of the Lemma -- (recall that the dataset is a field of lemma)

At the moment, the new gloss then falls underneath the same lemma as the original one (in the same dataset)

Creating a gloss with identical phonology (copy gloss) requires that the new gloss is created before copying the phonology into the fields.

At the moment, the user has no means to change the dataset (via the lemma) of the gloss.
This was to avoid users hijacking signs.

susanodd:
@ocrasborn if you desire to have this work cross-dataset, then it would probably need to be with two Gloss Detail Views adjacent on the screen, to see both glosses from both datasets. In order to copy-create the user needs to select the original gloss as well as the dataset, and this would likewise apply to the destination if it's a different dataset.

susanodd:
[...]

PR #878, opened 2023-01-19
Copy gloss button in Gloss Detail View

susanodd:
This allows the user to create a new gloss by copying the phonology of a gloss to a new gloss.
This implements issue #150. See the screenshots there.
The interface is done as a model. The user is allowed to select a different target dataset.
The annotation and lemma fields of the copy are suffixed by "-duplicate" as described in the issue.
If the gloss the user is trying to copy already has this suffix the user is asked to change it first and not allowed to make another copy. This to avoid multiple -duplicate suffixes after each other.
The user is brought to the Gloss Edit View of the newly created gloss, as for normal gloss creation.
The last used dataset is also updated.


susanodd:
This is completed. Changes to layout to accommodate new Copy button.
Repairs to `edit_text` fields because columns were messed up. (That wasn't related to this issue.)

Issue #879, opened 2023-01-27
Some changes to the interface text


ocrasborn:
These are some heterogeneous requests, but all to do with changes to the text I see in the interface, so probably easiest to deal with in one go; all applying to the English interface:
- There's a typo: field label "Conception Concept Set" should read "Concepticon Concept Set" (with a c inserted)
- In the detail view of signs, the labels for the different tabs at the top are a little messy, let's try to shorten them like this: Public View, Details, Relations, Frequencies, Revision History, Videos; by removing 'view' as much as possible it becomes a bit more compact. The 'view' is a bit programmer speak, I realise now.
- The panel Morphosyntax used to be called and should be Semantics, I'm not sure why it changed at some point. Please rename back to Semantics.

Please try to adjust the Dutch translations (in the .po file?) accordingly; let me know if you need any suggestions there @Jetske

susanodd:
The Query View isn't shown with the gloss, it's shown as a pull-down in the Analysis Menu.
It's true there is a `nav-tab` (a single one) in the template.
This will be expanded to allow users to see saved queries (#118), hence the use of a `nav-tab` in expectation of other functionality.

Reviewing the code #880, I think it looks weird (in English) if it's called Query List and Query (by removing the View).

The view itself is the list of search results from the most recent search, with the query parameters shown above them.
So it's **Query Parameters** plus **Query Results**.
Could also be **Query Display**
I'm not exactly sure what it is, so it was called Query View.
It used to be a panel for the Gloss, but during development, @Woseseltops pointed out that it wasn't specific to a gloss, so we moved it to the Analysis menu.

susanodd:
I deployed the changes last week and compiled the new .po files on Global.

But now (on Apple) it has started using **en_us** instead of **en**.

This is observed by **Weak Hand** becoming **Nondominant Handshape** in the Minimal Pairs List (under Analysis).

How did this happen? Is it Safari overriding something? This is very strange because I have Safari in het Netherlands. Indeed, my MacOS is in het Nederlands.

susanodd:
Fixed it by removing the compiled file for en_US

PR #880, opened 2023-01-30
Textual changes and translations edited #879

Jetske:


susanodd:
In order to allow translation of the fuzzy strings, you need to remove the `#, fuzzy` line and `#| msgid <string>` above the (fixed) translation.

There is a problem here:

https://github.com/Signbank/Global-signbank/blob/e9071d857d6f74dc9a63d3cbd623a1faa3a53f36/conf/locale/nl/LC_MESSAGES/django.po#L2513-L2523

Above, Django is being instructed to translate Query View to Frequencies. (Because it didn't find anything appropriate, it is basing its translation on the fuzzy part.)

If we just want to get rid of View everywhere, then the first could be `Query Lijst` and the second `Query`.

Jetske:
Thanks Susan, I didn't know about the fuzzy part. 

I think for Query View that 'view' is okay to stay (for now). What is the reason that it was translated before to 'Frequentie-weergave' instead of just 'Query-weergave'?

susanodd:
Django does that automatically. The fuzzy ones are ones it can't find any matches for, then it tries to guess. That's why on those fuzzy ones you always need to check what the thing it's using right below the line fuzzy, that it's the correct thing.
Sometimes it's hilarious what it's attempting to translate. (!)

Woseseltops:
I guess the suggestions by @susanodd still need to be processed before this can be merged, right?

Jetske:
@Woseseltops no, I believe they were processed in the 2nd commit

susanodd:
Approved.

PR #881, opened 2023-01-30
Feedback menu

susanodd:
This contains reworking of the Feedback Menu code. Stage 1 has removed obsolete and unused code. (The template show.html is no longer used, it has been replaced by separate templates per type of feedback.) The interface feedback has been removed and everything related to Auslan has been removed. This leaves a kind of skeleton type of feedback. Stage 2 would add additional fields, if deemed necessary.

[SCROLL DOWN YOU CAN SKIP THIS NOW, IT'S BEEN CODED AWAY]
I would appreciate feedback. Specifically in the templates where the code obtains the sign language, dataset, and annotations.
The original code used the attribute `idgloss`, but this ended up giving unexpected results.
The code includes lots of checks on whether fields are empty or not. I used the Sign Language in e.g., Missing Sign Feedback, since I figured the user would know the Sign Language whereas the Dataset is a construction for researchers.

https://github.com/Signbank/Global-signbank/blob/67e764c182d6ab65f3fd0354ef9736ddded2c8e1/signbank/feedback/templates/feedback/show_feedback_signs.html#L88-L103

In the code above, it refers to "translation". This is a trick of the original code. If the original gloss the user is giving feedback on did not have any keywords, an "empty" keyword is created and that is used to link the gloss to the feedback. This obtaining the gloss via the keywords has been left, since the original feedback model used this field. A new field has been added for Gloss.

In the case of Morphemes (now separated from Gloss feedback), it becomes unwieldly that the information about the dataset isn't directly available. Moreover, we originally only had morphemes for NGT. So this seems more like an afterthought. Perhaps some methods in the models would help the code be less obtuse.


susanodd:
@vanlummelhuizen shall we get rid of the `translation` field of SignFeedback? It was used to identify the gloss originally. (See the obtuse code above in the template.)

https://github.com/Signbank/Global-signbank/blob/3184b3e381e629eeeef30fbb335ed28f75e2694b/signbank/feedback/models.py#L68


(I'm busy restricting the feedback overview to the selected datasets. I just realized the translation field of the MorphemeFeedback class is unnecessary, since this class is newly created so it's not been used.)

For SignFeedback, there are some extremely old feedback entries that use the translation field to identify the gloss. Perhaps these can be removed? Or do we need a migration to copy the calculated gloss to the new Gloss field?

susanodd:
Okay this is pretty much finished. I just need to examine the Admin for Feedback.
Also, @ocrasborn would like dataset managers to receive appropriate feedback. Any ideas how best to do this?

susanodd:
@vanlummelhuizen **read this first**.

You can go ahead and read the code. I repaired the things I contemplated above.

susanodd:
@vanlummelhuizen this is pretty much completed.
One thing perhaps you can shed light on is what to do about permissions?
So far, only group Editor is required.
For the recent feedback page, it shows the feedback for the selected datasets.
Do we need to do anything about permission `change_dataset`? Or is this implicit based on selected datasets and editor?

As it is, I didn't do anything to the delete functionality. It merely changes the status to deleted. (These are then not shown anymore. Nothing has ever been done to modify a status to 'read'. I don't think we want to do anything about those.)

vanlummelhuizen:
@susanodd At first glance, it all looks like a major improvement. :+1: 

A few remarks/responses:

> Caveat, do we really want to send them email directly? Or how? I guess it would be something like a notification when new feedback comes in. Or perhaps it could be paired with Recently Added Signs. That somehow, it's directed but lazy.

I think for now sending an email directly  is ok since there is not too much feedback right now.  I think it is wise though, to think about how we could improve this experience later. For instance, send a daily email with all new feedback message, if there is new feedback (so no daily message if there is no new feedback that day).

> One thing perhaps you can shed light on is what to do about permissions?
> So far, only group Editor is required.

I think it would be cleaner to test for concrete permissions instead of checking whether the user is in a group. The group 'editor' should have the permission 'can view feedback' or something similar. Technically, the result will be the same, but it would be more in line with the intentions of the user/group/permission functionality of Django. Moreover, if other groups need permission, we simply set it in the admin instead of changing code.

> For the recent feedback page, it shows the feedback for the selected datasets.
> Do we need to do anything about permission `change_dataset`? Or is this implicit based on selected datasets and editor?

I think using the 'can view feedback' permission here would solve this.

susanodd:
@vanlummelhuizen there are actually quite a few no longer used permissions in the database.
It seems they are not removed automatically.
I found this command mentioned on stackexchange and in the djangoproject (for 4.1):
`remove_stale_contenttypes`
I ran it on my own database, but it shows lots of stuff not related to feedback.
To me, the output of the command isn't clear what it found. So I kept choosing "no".

susanodd:
There are 4 permissions in the database related to `interpreterfeedback` (this class has been deleted, but the permissions still exist)
```
add_interpreterfeedback
change_interpreterfeedback
delete_interpreterfeedback
view_interpreterfeedback
```

The `view_ permission` was created by a migration (`0001_initial.py`)

@vanlummelhuizen what is the best way to get rid of these?

The implicit permissions created by Django don't include any view_ permission.

susanodd:
I've added specific permissions for the respective feedback classes.
I checked that group editor has the respective feedback permissions. It does.
But of the other groups, group publisher also has them.

Issue #118, opened 2016-01-17
Save and display search history


ocrasborn:
- Allow the user to browse his past searches, in principle an unlimited list.
- See the screen shot of the Asian Signbank (Hong Kong) for how they display it: rather useful, I find: users can re-run previous searches (updating the no. of results), and combine previous searches.
- Missing, please add: The date the search was last executed.
- List searches by last-used-date, most recent one at the top.

![Uploading Screenshot 2016-01-17 14.18.33.png…]()

henrinie:
In FinSL we are saving the latest search results for every user (in users session), and then we show those results as an inline list on the gloss page. User can then browse the search results without having to leave the gloss_detail page.

Do you want to save the search results or the search terms?

vanlummelhuizen:
@ocrasborn Is this still or your wishlist?

ocrasborn:
Yes, I would still like to see this; not sure about the priority though.
@henrinie : I would prefer to save the search terms, so that it can be run again over either updated data or another selection of datasets.

susanodd:
Here's what it looks like now. I added a Save button to the Query View, and another panel showing the Search History.
In this example the last query in the list is the one saved.

This needs some basic user functionality, to remove queries, and to modify the name. (At the moment that can be done in the admin.)

I experimented with different displays for the parameters of each query. This is the most concise as a string. I'll make a View button where the user can examine a saved query.

![query-view-save-button](https://user-images.githubusercontent.com/24272544/217548943-b4d11305-d5a0-4b5d-b950-da8ffc77deb7.png)

![search_history_last_shown](https://user-images.githubusercontent.com/24272544/217548979-6341f003-7ece-4456-b858-4e2a1e5ad648.png)

susanodd:
Here's with the boolean parameters added, queries most recent first.

![query-view-boolean-parameters](https://user-images.githubusercontent.com/24272544/217813454-8107ce05-267b-44bc-9ca9-1d74e2f64796.png)

![search-history-boolean-parameters](https://user-images.githubusercontent.com/24272544/217813487-b590c89a-0861-46db-bc7d-3d999de11fe4.png)

susanodd:
I added some buttons to the search history view.
Here's an example showing editing the query name of a new query, and that it gets updated in the list outside of the modal

![query-view-save-query-multiselect](https://user-images.githubusercontent.com/24272544/218112170-06b6fad1-41fe-4cca-8769-b1bc12cf58b3.png)

[...]

PR #883, opened 2023-02-15
Search history functionality (#118)

susanodd:
This branch adds models and functionality for storing query parameters and queries in the database

- models use inheritance from a superclass QueryParameter, with subclasses for each of the specific types of data queried upon
- a new file `search_history.py` includes specific functions for translating a query stored in the search history into the form used in the session variables, as well as a pretty print for display in the search history template
- the search history template works on a list of queries for the user. The javascript code has been parameterized so as to work on queries in the list (as opposed to other uses where just one thing was in the template, such as a gloss or dataset)
- the buttons in the search history template make use of modals
- also in the template is a column for languages (multilingual fields specific to Annotation, Lemma, and Keyword). If the languages used in the query are not in the selected datasets, the Run button is not offered (although caught behind the scenes)
- the admin for queries (SearchHistory objects) disallows modifying the query parameter objects stored. Merely viewing them, except for changing the name of a query.
- the user can save queries via the Query View template
- via the Search History view, the user can see their own queries and can Edit the name of the queries

TO DO

- resolve the reference to `GlossSearchForm` fields that appear in queries (these are hard-coded at the moment since the models cannot get stuff from the forms (@vanlummelhuizen @Woseseltops how to resolve this in an elegant way?)
- some obtuse code in Admin for pretty printing the queries. This might be combinable with those in search_history.py


susanodd:
Anybody have any comments? Or shall I just put it live?

PR #885, opened 2023-02-16
Record webcam videos 266

Jetske:


Jetske:
Ready for a review, but please don't merge yet, as I would first like to fix #886 (a problem that may also pop up here)

Jetske:
The last commit and the fix in #888 should solve this as well, so if that works this can be merged too.

susanodd:
The changes in #888 have been merged and deployed.

PR #887, opened 2023-02-20
Fixed database test by changing gloss api test setup method

TobiasCunnen:
The test did not run because the gloss video folder was deleted when the test class was done running.
Now the folder is created before a single test and after that test it is also deleted and that seems to fix the problem.

I am not sure what the actual problem was, another test probably used the folder when it was removed.

There are still 4 failures and errors but i believe those have nothing to do with this problem (but again not sure)

Linked this branch to issue 882 Running tests in production container doesn't work

### DIFF
--- a/signbank/dictionary/tests.py
+++ b/signbank/dictionary/tests.py
@@ -3475,54 +3475,55 @@ def decode_messages(data):
 
 class GlossApiGetSignNameAndMediaInfoTests(TestCase):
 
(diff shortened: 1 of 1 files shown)


susanodd:
Looks good!

These are the only two failures I get now:
test_create_and_delete_utf8_video
test_Settings
(those have been broken for a long time)

Issue #886, opened 2023-02-16
Problem when playing an uploaded video


Jetske:
An error occurs in the drag and drop code, or the form that it uses, when opening a gloss detail page
See: https://signbank.cls.ru.nl/dictionary/gloss/44008/

Might be a browser problem, or maybe it is saved somewhere different than in my local setup, or it is something else. 

Check why "id_videofile" is nowhere as 'id' but is only used for 'for'?

susanodd:
It won't play in Safari either. It doesn't show the normal image for not available.

Jetske:
No, I think that's because the video should be there (it also shows up in the writable/glossvideo/NGT/LO folder) but somehow it still can't find the video. This happens for both LOPEND-VOORTBEWEGEN-B and LOPEN-L, which are both only shown when logged in. Could that be related? 

When I try to upload the same video for those two glosses locally, it works fine.

susanodd:
Is the group or permissions messed up?
I get the same on Firefox. But it finds the image for the gloss okay, so the path ought to be correct.

I tried searching on (Dutch annotation) ^LO
and there are 31 results. But in the console, only the matches for 44010 and 44008 show errors. (There are 3 messages about each.)

Jetske:
Maybe? But if I (locally) give a test user only permission for "dictionary | gloss | Can Search/View Full Gloss Details" then it finds the corresponding videos for those glosses without problems. 

By the way, I did solve the error for 'onchange' in drag_drop_files.js, that was about the input field 'id_videofile' and presumably not related to the video loading problem.

susanodd:
I adjusted the permissions on the server, so they're visible now.

susanodd:
I'm trying this now locally on master branch.
I can successfully upload a video (using the normal button not drag and drop).
It successfully shows up.
However, the new video created has the "lemma" of the gloss, not the annotation in its filename.

`writable/glossvideo/tstMH/MA/MAAL-42088.mp4`

(This will be a problem for the variant names.)
(See the fields below. It chose the Dutch Lemma as the filename.)

![add-video-to-gloss](https://user-images.githubusercontent.com/24272544/220581062-ae54ad9e-c5f5-45e6-a08b-09390396f549.png)

susanodd:
I have the current global database now.
(I am the owner of the uploaded file, so it shows up on PyCharm runserver.)
I'll try your branch again @Jetske

susanodd:
[...]

PR #888, opened 2023-02-21
Uploading video problem #886

Jetske:


susanodd:
@Jetske I'm running your branch on my computer now.
It's kind of impossible to figure out if the code works, because I'm using PyCharm runserver, so the "server" is actually my own account. This is causing problems with file system permissions. Normally I need to use "sudo" commands if I want to modify things Signbank has created. But for some reason, without apache (the good old days of having an apache server!!!) it's not possible because "www-data" should be signbank, but it's somehow "susan" that owns the files.
So the Signbank server "www-data" can't change stuff that "susan" has created.
If I could run apache again, I could easily test the code on my computer. (@Woseseltops !!!!!)
We may need to just deploy the software and test it on the real signbank.

susanodd:
To upload a video, I needed to modify the dataset subdirectory of glossvideo to 777, just to be able to allow the software to save the file. (As your code is written.) But then the code wants to change the permission of the file itself and change the owner.
But this isn't allowed because on my computer, that would need to be done with "sudo".

What do you say? Just deploy?

susanodd:
Okay, I had to remove (comment out) this line:

https://github.com/Signbank/Global-signbank/blob/b6fdd5b63a8be7aebfea7d5ebe809499a2310389/signbank/dictionary/models.py#L1829

It was actually having trouble doing stuff to the "old" file that was there.
I also removed the old files (on my local machine, the reversion files it makes)
Then commented out that line and ran it again and it works correctly.
I am the owner of the uploaded file, but it is visible to others.

susanodd:
I say comment out the `ch_own_mod_video` (as mentioned above)

Except it also needs to use the Annotation not the Lemma for the filename.

Also there is something messed up with the reversion files. (One one run, it made a _small video of a reversion file instead of the new file. Then I removed all the files and uploaded again. So it seems the execution depends on files already present.)

Jetske:
Okay, I can check annotation/lemma file name. 

But the problem was that the video didn't play because of not having the right owner and permissions, which I then changed in the ch_own_mod_video method. So commenting that out would not solve it.. Or do you maybe know another way? Would it then be enough to change only the permissions, and not the owner?

susanodd:
Oh yes, try that. I suspect that the owner will be the website.
To be sure I have the folder open on the global, the owner is ubuntu, the group is signbank-writable.
I suspect you should not try to change owner or group.
The video files have permission -rwxrwx---
The problem files did not look like that. I can't recall now what they were. I only made them match the rest of the files in the folder.
(There are 33 files in LO.)
So maybe make the permissions match that.

I suspect the only way to test it is to deploy it.
I was doing that yesterday for the tests. Since they were working on my computer I had to debug them on global.

susanodd:
@Jetske I uploaded a video on the live server using drag and drop (very nice!)
It works!
However, on the server, the newly uploaded file:
`writable/glossvideo/tstMH/LA/LAWAR-42089.mp4`
has permission:
`-rw-r--r-- ubuntu signbank-writable`
So it's the permissions that are wrong. Or rather, don't match the other files.

Other files are:
`-rwxrwx---`

This may only lead to problems when a new video is replacing an existing one, as it seems it doesn't have w permission.
Since your original discovery was sharing a lemma with other glosses (?) perhaps that's why, since it would have tried to move the previous video for the lemma.

susanodd:
I'm looking at videos on the server, for similar permission quirks.

For two videos for gloss 4611 (in tstMH) the permissions are correct (`-rwxrwx---`)
`michatest1-4611.mp4`
The files are from Aug 5 2021

For this video (tstMH):
`testsusanlemma_nl1-36460.mp4`
Sep 17 2021
The permissions are `-rw-r-----`
So sometime between Aug 2 and Sept 17 2021 the code was changed.
The file (tstMH)
`HAKEN-42091.mp4`
Nov 26 2021
has the same permissions as they are now `-rw-r--r--`
So the code was changed again between Sep 21 and Nov 26 2021 to what it is now.

Again, I suspect whatever is being done on backup files is causing problems, since write permission isn't there.

susanodd:
I found examples!!!

See the gloss for HERSENVLIESONTSTEKING-C
(Its video is not visible, but it is on the server.)

The "hidden" glosses have the following:
`-rw------- 1 ubuntu www-data`
rather than `signbank-writable`

That would be that the path is to STATIC rather than WRITABLE when uploaded
(There used to be some other media files that were being put in static instead of writable. That should not happen anymore.)

susanodd:
I'll fix the ones I find, since they ought to be visible.

susanodd:
I found the glosses via recently added signs

susanodd:
@Jetske is it possible the drag and drop is copying the temp files to its own space?
The javascript files are under STATIC.
Somehow the wrong (not visible) files have group www-data instead of signbank-writable
I suspect they are being uploaded to something under static before being copied.
Like to where the javascript lives or the signbank code lives.
The writable is a totally different directory. It's not reachable from the python (GitHub files)
There shouldn't be any paths trying to get to writable. It needs to use the settings.
And that protected media is actually a url that gets to the files.

susanodd:
Another possible cause could be people dragging from another "website" instead of their own computer?

I went to visit with Lianne yesterday to tell her I was fixing her videos. She briefly had a window open that showed a list of all of her video files. It didn't look like signbank. (It was a browser window.) Perhaps she was dragging from another website? That would explain the www-data being in the group.
The upload should be from a persons own computer.

Jetske:
I don't know about whether it is copying the temp files to its own space. When dropping the file, it ends up in the input field and then the upload button submits it like before. Maybe if the uploaded files are too big to be in memory, they are temporarily put somewhere in static? In add_video, its goal location is somewhere in WRITABLE_FOLDER , so that uses the setting. 

When I try to drag a video from another website, it doesn't allow it ('only 1 file accepted')

Jetske:
-rwxrwx--- rights should be obtained with: stat.S_IRWXU | stat.S_IRWXG

susanodd:
Yes, when there is an error it doesn't complete things.
It's probably because the file you first had the error with was too big.

Jetske:
Correction: the WRITABLE_FOLDER is used for a GlossVideoHistory object. The new videofile should be saved to GLOSS_VIDEO_DIRECTORY (writable/glossvideo).

susanodd:
Here are some examples that were not readable, they are all quite large:
```
-rw------- 1 ubuntu www-data          4342630 Feb 21 09:57 RUIKEN-B-44019.mp4
-rw------- 1 ubuntu www-data          4841485 Feb 21 10:19 APART-D-44020.mp4
-rw------- 1 ubuntu www-data          5459319 Feb 21 10:57 TERRAS-44023.mp4
-rw------- 1 ubuntu www-data          4329786 Feb 21 11:59 ACHTER-D-44024.mp4
-rw------- 1 ubuntu www-data          8002597 Feb 21 12:04 BRIL-B-44025.mp4
```
Here are normal ones:

```
-rwxrwx--- 1 ubuntu signbank-writable  180828 Aug  5  2021 BREKEN-4536.mp4
-rwxrwx--- 1 ubuntu signbank-writable   27464 Aug  5  2021 BREKEN-4536_small.mp4
```

susanodd:
Here's two done on the same day:

```
-rw------- 1 ubuntu www-data          5289583 Feb 21 13:19 OPGELUCHT-C-44027.mp4
-rw-r--r-- 1 ubuntu signbank-writable  111571 Feb 21 14:23 OPLETTEN-A-2785.mp4
```

Jetske:
Ok, that's strange. 

I do get 'Error resizing video' locally, something with avconv/ffmpeg not being installed correctly. I assumed that resizing works on the live server, but maybe it doesn't? But then, resizing happens after saving to the right location and I can't imagine that that changes ownership.

susanodd:
It might be that the requirement isn't installed properly.
On pip freeze locally
`cngt-scripts==0.0.1`
But live
cngt-scripts @ git+https://github.com/vanlummelhuizen/CNGT-scripts@adebb61f5cb0c0004593c506d2211ab2e303d2d9

It looks like maybe they're not installed.

The live tests also give errors about resizing on the server.
Locally I have this setting
`FFMPEG_PROGRAM = "ffmpeg"`

Jetske:
FILE_UPLOAD_MAX_MEMORY is the setting that determines at which size the file gets stored to a /tmp directory instead of in RAM. If that is indeed the problem with www-data, we can try:

FILE_UPLOAD_MAX_MEMORY_SIZE = 5242880 # 5 MB

And then it's probably possible to add in javascript that it does not allow to drop a file of a bigger size. I think 5mb is enough.
Still, after that ideally it should resize the video.

susanodd:
This is the error on the live server when it can't make a small video:
```
Error resizing video:  /var/www/writable/glossvideo/tstMH/LA/LAWAR-42089.mp4
Expecting value: line 1 column 1 (char 0)
```
I modified the code to print the exception text.

susanodd:
> FILE_UPLOAD_MAX_MEMORY is the setting that determines at which size the file gets stored to a /tmp directory instead of in RAM. If that is indeed the problem with www-data, we can try:
> 
> FILE_UPLOAD_MAX_MEMORY_SIZE = 5242880 # 5 MB
> 
> And then it's probably possible to add in javascript that it does not allow to drop a file of a bigger size. I think 5mb is enough. Still, after that ideally it should resize the video.

I added this to the server-specific settings file. It's live now.

susanodd:
This issue is separate from the small videos.
There's a command `create_small_videos` but using this (on tstMH) doesn't work even after modifying the permissions and group as above, so that is a separate problem.

susanodd:
I'm going to merge this. Lianne is uploading videos right now.

PR #895, opened 2023-03-02
Max file size for drag&drop, #886

Jetske:
The file size was not limited for dropping files and through the normal input file field, so this is added here


susanodd:
Approved.

PR #899, opened 2023-03-09
Matroska files converted to mp4 issue #886

Jetske:
Please run this to test, as my database is not up to date!


Jetske:
@susanodd yes this needs ffmpeg (version 1.4)

Jetske:
@susanodd It gives an ffmpeg error (on live signbank) when uploading a video that I believe was recorded on an apple device

Issue #911, opened 2023-03-14
Uploading videos truncates the video


susanodd:
I tried using this feature and I signed a test sign twice but the video cut off after 1 sec

_Originally posted by @rosestamp in https://github.com/Signbank/Global-signbank/issues/266#issuecomment-1464213011_

susanodd:
This happens with all video uploads.

It doesn't happen with Other Media uploads.

susanodd:
The other media model uses a `CharField` type for the `path`:
https://github.com/Signbank/Global-signbank/blob/ea4378edef1818f0eb7919eb0da8d0ce3bca6005/signbank/dictionary/models.py#L2308-L2317
whereas the GlossVideo uses Django FileField
https://github.com/Signbank/Global-signbank/blob/ea4378edef1818f0eb7919eb0da8d0ce3bca6005/signbank/video/models.py#L208-L219
which in turn uses:
https://github.com/Signbank/Global-signbank/blob/ea4378edef1818f0eb7919eb0da8d0ce3bca6005/signbank/video/models.py#L98-L102
where the setting is:
https://github.com/Signbank/Global-signbank/blob/ea4378edef1818f0eb7919eb0da8d0ce3bca6005/signbank/settings/base.py#L24
(But this is actually wrong.)

susanodd:
We could define a new model field path for GlossVideo and store the actual path rather than let Django do it. (Using a simple migration to just copy the existing paths from the FileFields.)

Or this really old issue: #320.

This is probably related to large video files because they aren't finished being uploaded when the FileField needs to be stored in GlossVideo.

Jetske:
I tried some more videos and the cutoff seems to happen in convert_to_mp4 so I switched that off for uploads and am working on another solution for recordings..

susanodd:
I deployed your changes, @Jetske.
It's back to normal for the normal uploading.

susanodd:
I had this comment in the pull request:

> I'm getting the conversion error on my local PyCharm server now.
> But because the convert video uses "stderr" the actual message can't be captured in the except clause, so it's not possible to see what the error is. It's also not possible to determine an appropriate exception type for the try except clause.

Perhaps that's why we can't catch the errors to take appropriate action. It seems the exception is from inside the ffmpeg python module and that is printing it to stderr. Maybe the newest version will solve this with the Django and Python bumps?

susanodd:
[...]

PR #916, opened 2023-03-15
Retrieve convert_to_mp4 #911

Jetske:


susanodd:
I had just finished editing it back on my local code.
It also needs the undo on the file name because that's the part that copies the file.
The test video doesn't have an extension, so that was causing it to fail.

I pushed mine to the master.

susanodd:
@Jetske you can just pull from master again now.

Issue #946, opened 2023-04-05
Review *view* permissions


vanlummelhuizen:
At some point between version 1.11 and 4.x, Django added view permissions. We also had our  own view permission before Django had them and now they may conflict or may be confusing, not working properly, etc. See e.g. #940. Please review the use of those view  permissions and try to conform to one system, I suggest the Django system. This may involve adding Django view permissions to users or groups where there is already a view permission that we made.

vanlummelhuizen:
I think https://github.com/Signbank/Global-signbank/issues/940#issuecomment-1497113069 is also caused by this.

susanodd:
I wrote up a lot of comments in #940 about what is visible in the Admin.

susanodd:
Okay, here is something weird. The test-user I made, I tried granting the user View permission for tstMH (local server).

![assign-view-perm-test-user](https://user-images.githubusercontent.com/24272544/230093680-f36cc2f0-f8dc-4a00-8bd7-224681b197f2.png)

susanodd:
It seems it's the send_mail after assigning permission. (The send mail to the user. Although PyCharm doesn't report anything about this. The user does have permission. It's just the notification that failed.

susanodd:
Here's what that users new permissions look like among all the dataset permissions:

![tstMH-dataset-permissions-test-user](https://user-images.githubusercontent.com/24272544/230097291-36e05c9d-3ecd-4b57-ab45-84aca66b165d.png)

So now the test-user has the can_view_dataset permission (assigned via Dataset Manager) but not the view_dataset permission.

susanodd:
Maybe the Django permissions have evolved enough that the Groups idea can be used? Once upon a time, group NGT was created with this in mind.

susanodd:
The **page** Dataset Manager is visible to users in group Editor as well as group Dataset_Manager. That's why it shows up to users that are not able to use it. (See permissions in Admin for the page.)

susanodd:
@ocrasborn wrote:

> Sorry all @susanodd @vanlummelhuizen , not quite solved yet: when selecting one of the items in the 'Recently added signs', the person who added the items gets the message 'The gloss you are trying to view is not in a dataset you can view.' Confirmed by me by logging in as the dataset test user. I imagine this must be related, but feel free to open a new issue :-)

So this is related to selected datasets. (That's what I changed in the pull request, so all places that look up the view permission check both permissions.)

susanodd:
[...]

PR #948, opened 2023-04-05
Changes to accommodate can_view_dataset as well as view_dataset

susanodd:
This combined with the patch outlined in #946 which renames the new permission.


susanodd:
Since Django has a prefab view_dataset, we actually could just remove can_view_dataset and not declare it and just use view_dataset.

Issue #984, opened 2023-06-01
Tags field entry in Gloss Search Form formatted weird


susanodd:
Gloss List View something happened to the formatting of the multi-select Tags field.
It's not displaying properly.

susanodd:
This issue was the multiselect field of Tags.
I modified it so it appears the same as the other multiselect seach fields.
It now behaves the same as the other fields being an "or" instead of an "and".

susanodd:
This is live now.

PR #991, opened 2023-06-08
#984: Made tags search multiselect OR rather than AND.

susanodd:
Made tag choices dynamic in forms. Adjusted query and search history code to accommodate.

Made tags in Gloss List also visible on Morphemes.

### DIFF
--- a/signbank/dictionary/adminviews.py
+++ b/signbank/dictionary/adminviews.py
@@ -1064,6 +1064,16 @@ class GlossListView(ListView):
             query_parameters['dialect[]'] = get.getlist('dialect[]')
             qs = qs.filter(dialect__in=vals)
 
+        vals = get.getlist('tags[]')
+        if '' in vals:
+            vals.remove('')
+        if vals != []:
+            query_parameters['tags[]'] = get.getlist('tags[]')
+            tags = Tag.objects.filter(name__in=vals)
+            tqs = TaggedItem.objects.filter(tag_id__in=tags)
+            glosses_with_tag = [taggeditem.object_id for taggeditem in tqs]
+            qs = qs.filter(id__in=glosses_with_tag)
+
         # allows for multiselect
         vals = get.getlist('signlanguage[]')
         if '' in vals:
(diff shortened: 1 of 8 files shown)


signbank/dictionary/forms.py
```diff
@@ -260,8 +261,7 @@ class GlossSearchForm(forms.ModelForm):
 
     search = forms.CharField(label=_('Search Gloss'))
     sortOrder = forms.CharField(label=_('Sort Order'))       # Used in glosslistview to store user-selection
-    tags = forms.MultipleChoiceField(label=_('Tags'), choices=tag_choices)
-    nottags = forms.MultipleChoiceField(label=_('Not Tags'), choices=not_tag_choices)  # this field is not used in the template
+    tags = forms.ChoiceField(label=_('Tags'), choices=tag_choices)
```
vanlummelhuizen, line 264: Does this line and `__init__(...)` mean there is a 'tag' and 'tags' field?

signbank/query_parameters.py
```diff
@@ -196,6 +196,12 @@ def convert_query_parameters_to_filter(query_parameters):
 
         elif get_key == 'dialect[]':
             query_list.append(Q(dialect__in=get_value))
+
+        elif get_key == 'tags[]':
+            tags = Tag.objects.filter(name__in=get_value)
+            pks_for_glosses_with_tags = [item.object_id for item in TaggedItem.objects.filter(tag_id__in=tags)]
```
vanlummelhuizen, line 202: This could be a single line. Similar to my comment in adminviews.py.

vanlummelhuizen (changes requested):
Good, @susanodd . I mainly have minor coding style remarks, except for the tag/tags field thing. That you probably need to look into especially.

signbank/dictionary/adminviews.py
```diff
@@ -1064,6 +1064,16 @@ def get_queryset(self):
...
             qs = qs.filter(dialect__in=vals)
 
+        vals = get.getlist('tags[]')
+        if '' in vals:
+            vals.remove('')
+        if vals != []:
+            query_parameters['tags[]'] = get.getlist('tags[]')
+            tags = Tag.objects.filter(name__in=vals)
+            tqs = TaggedItem.objects.filter(tag_id__in=tags)
+            glosses_with_tag = [taggeditem.object_id for taggeditem in tqs]
```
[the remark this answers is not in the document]
susanodd: > Suggestion to do it in one line:
> 
> ```python
> glosses_with_tag = list(TaggedItem.objects.filter(tag__name__in=['add_phonology']).values_list('object_id', flat=True))
> ```

Doesn't this add an extra "dict" to the result, so you need to do glosses_with_tag['object_id'] to retrieve?

signbank/query_parameters.py
susanodd: > This could be a single line. Similar to my comment in adminviews.py.

This ended up being as it is because it's awkward to get things out of the TaggedItem result.
The objects themselves can be any object, not just glosses.

What confuses me is that the TaggedItem supposedly have tag_id and object_id, but it doesn't need to be an id.
I suspect I wrote it so to force it to evaluate it as specified. :)

susanodd: Hmmm. It seems impossible to create a query that has a '' inside the getlist !!

This may be a remnant from before the field choice model was created, when they used to be "char" fields and could be empty.
It could also be a remnant from when the lists were hard coded and also included empty choices.

I'll remove them!

signbank/dictionary/forms.py
susanodd: No, it's overwritten by the `__init__` (They're both 'tags'.)
It's there in the class declaration so it's available statically. Some of the code inspects the form to look at the labels.
(Since the fields in the gloss have different labels in the search form, so when they are displayed as query parameters, the field label of the search form is used.)

susanodd:
@vanlummelhuizen finished!

signbank/dictionary/forms.py
vanlummelhuizen: My point is that in `__init__` is says `self.fields['tag'] = `, so singular 'tag', and below it says `tags = ` with plural 'tags'. Is that as it should be?

susanodd: I found it, it's in the TagUpdateForm.
The field is called just 'tag'.
I added it to the init form so it also gets the dynamic tag names.

https://github.com/Signbank/Global-signbank/blob/833b720835f6970803850c44a73078b15a5a433a/signbank/dictionary/forms.py#L205-L218

PR #992, opened 2023-06-08
App endpoints

TobiasCunnen:
Hi all, 

I created this draft pull request because I added the gloss id to the data that returns but I can't test if this change breaks anything because I do not have the correct database anymore to run Signbank locally.

I spoke with Susan last week and she said I could get the latest database from the applejack server but I do not have permission to access the server.

I updated my own unit test and because it is a really small change I don't think anything else will break but like mentioned before I have not tested it.

How can I get a copy of the new database or can someone else test it let me know if things break?

### DIFF
--- a/signbank/dictionary/tests.py
+++ b/signbank/dictionary/tests.py
@@ -3654,6 +3654,7 @@ class GlossApiGetSignNameAndMediaInfoTests(TestCase):
         """
         assert_json = []
         sign_json = {
+                "id": self.gloss_id
                 "sign_name": self.gloss_name,
                 "image_url": "",
                 "video_url": self.video_url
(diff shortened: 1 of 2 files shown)


susanodd:
https://github.com/Signbank/Global-signbank/blob/28e9998a8bc2d52f93102cd637900d45e3a34c41/signbank/dictionary/views.py#L2925

You might need a `str() `around the gloss.id

TobiasCunnen:
Local environment is now working with some help from @susanodd.
Now everything seems to works and the all the test pass.

TobiasCunnen:
> https://github.com/Signbank/Global-signbank/blob/28e9998a8bc2d52f93102cd637900d45e3a34c41/signbank/dictionary/views.py#L2925
> 
> You might need a `str() `around the gloss.id

I want to return the id as an integer so I can use it, this way I do not have to parse it in the app.

Issue #884, opened 2023-02-16
Add Sense, ExampleSentence and ExampleSentenceTranslation models


vanlummelhuizen:
1. Add a Sense model between Gloss and Translation+Keyword models.
2. Add ExampleSentence and ExampleSentenceTranslation with a many-to-many link to Sense.

Schematic:
![IMG_20230216_143820293](https://user-images.githubusercontent.com/16682631/219388389-370ea6e3-89ad-4531-aa22-eda5f85e8124.jpg)

vanlummelhuizen:
@ocrasborn Can upload your proposal you handed out?

ocrasborn:
[Signbank_NieuweInfo_20230118_ontwikkelaars.pdf](https://github.com/Signbank/Global-signbank/files/10793248/Signbank_NieuweInfo_20230118_ontwikkelaars.pdf)

ocrasborn:
Additional request: in the Sense model, please add an index field (like for the comments) that allow us to order the different senses per gloss manually.

Jetske:
I made a new field for Senses which shows some keywords to define the sense. And a table of example sentences. 

![image](https://user-images.githubusercontent.com/10157482/231691210-5f8a6ccb-af4c-4865-97c3-ae1ca8649392.png)

In this example there are no translation equivalents for English given but is it true that that should show the same structure as for Dutch, with the senses given as keywords and literal translations of the Dutch example sentences? @ocrasborn 

And do you have a preference for opening the video of an example sentence (if there is one) as pop-up or listed in the 'videos' tab?

ocrasborn:
Hi @Jetske , 
1. That looks good! 
2. Yes, this should be available for all the languages in a dataset. It's true we haven't spent much attention adding translation equivalents for English for the NGT dataset, but for Kata Kolok for instance this has been done. Does this perhaps require more discussion? Is it related to how @vanlummelhuizen re-designed the language coding?
3. In the screen shot above I still miss the 'type' for each sentence (like interrogative, negative, etc.). (Well, and links to the movies.)

vanlummelhuizen:
> 2\. Is it related to how @vanlummelhuizen re-designed the language coding?

The re-designing of the language coding was related to FieldChoices and similar. The translations for these entities should parallel the *interface languages*, since they are presented in drop down lists, as part of the interface.

[...]

PR #1002, opened 2023-06-28
Add senses and examplesentences #884 

### LINKED ISSUE #966: Search field in menu bar should change from Translation Equivalents to Senses
This follows on #884 ; see also #964

### LINKED ISSUE #967: Modify list view to display Senses instead of Translations
This follows on #884

### LINKED ISSUE #968: Incorporate Sense and Example Sentence in detail view
Follows on #884 ; see mockups there.

### LINKED ISSUE #976: Button 'Select None' on page Select Datasets doesn't work anymore
Clicking it leads to first turn off all the selected checkboxes, then immediately reloading the page with those buttons selected again.
Related to #940 ?

Jetske:
To make it work:
- run the migrations
- run glosstranslation_to_senses [dataset] (optional, you can also add senses from scratch)
- in admin, add at least one sentencetype as fieldchoice (e.g. 'Declarative'/'Interrogative'/'Exclamative'/'Imperative')
- make a folder: writable/sensevideo/[dataset]

Things to consider:

- the example sentence videos are saved in a folder named by the id of the examplesentence it belongs to
This may cause problems, for example: sometimes it randomly saves it with a random string added, for example [id]_k4jk32o3.mp4 and then it can't find the file anymore. I don't know why (too short filename/SQLite?). Other than that it works the same as the glossvideos, so expect to encounter similar problems here. 

- 'with atomic()' is used anywhere in update.py where several changes are made to the database but I'm not exactly sure about its placement

- So far I only tried the command to make senses from existing translations for tstMH, NGT and Kata Kolok. It takes a long time so it didn't finish it for the last two but it seems to do what it's supposed to. 

- for modifying a gloss: at the moment this looks and works differently in the 'map senses to keywords' page than on the gloss detail page. I think eventually it would be nice to have those editing modals be similar (and preferably without needing to refresh the page), but that can be changed later. 

This will also close #968, close #967, and close #966


susanodd:
Update,

- the "map keywords to senses" works without reloading the page. It works with senses now.

- the senses can be populated from the existing keywords/translations by using the alternative command in the virtual environment:` translations_to_senses '<dataset acronym>'`
This command is intended to be used like a migration from the previous "Translations" to "Senses".
For datasets that make use of Translations, the existing data should be copied using `glosstranslations_to_senses` or `translations_to_senses` ...

- commands to import senses from a CSV are not available yet (under construction)

vanlummelhuizen:
Nice work! The interface looks nice and intuitive.

I see that if you add/change/remove something there is a full page load. That is ok, but it has the side effect of leaving edit mode. I think it would improve the user experience if the loaded page is also in edit mode. Just adding `?edit` to the  end of the url would do the trick. The alternative is to only update the part of the page that is about senses.

We may have to look into this problem before deploying:
> * the example sentence videos are saved in a folder named by the id of the examplesentence it belongs to
>       This may cause problems, for example: sometimes it randomly saves it with a random string added, for example [id]_k4jk32o3.mp4 and then it can't find the file anymore. I don't know why (too short filename/SQLite?). Other than that it works the same as the glossvideos, so expect to encounter similar problems here.

When you say randomly, do you mean it is not reproducible?

susanodd:
@vanlummelhuizen , yes, it is not reproducible. It just keeps creating more and more files with the weird characters.

vanlummelhuizen:
https://github.com/Signbank/Global-signbank/issues/884#issuecomment-1613111014 may also needs some attention.

Woseseltops (commented):
Great, this is well readable code I like how you put some functionalities in admin commands as well.

Suggestions for improvement:

* The file `dictionary/models.py` is becoming quite full... would the functionality you added here be a good candidate for a chunk to cut loose and put in its own file?

* In `senses_kewords_edit.js`, I see html elements are being created inside strings (and I think we might do that in various other places in Signbank as well). My gut feeling says it's probably safer to use Javascript's createElement instead, but it does require more code.

* I see lines like these often appear together, would it make sense to make to give them their own reusable function? Any differences between the similar blocks of code code be arguments to this function.

```python
sense_translations = {}
for dataset_translation_language in self.get_dataset().translation_languages.all():
    if self.senseTranslations.filter(language = dataset_translation_language).exists():
        translation_list = sorted([st.get_translations_return() for st in self.senseTranslations.filter(language = dataset_translation_language)])[0]
        sense_translations[str(dataset_translation_language)]= translation_list
    else:
        sense_translations[str(dataset_translation_language)]= ""
```

* I see you use guard clauses in most places, but not in `update.py`. For example

```python
    for dataset_language in dataset_languages:
        if str(dataset_language) in request.POST:
            values = request.POST[str(dataset_language)].split("\n")
            if not values[0] == '':
                for k, v in enumerate(values): 
                    values[k] = v.strip()
                values = sorted(values)
                vals[str(dataset_language)]=values
```

could be (also notice the list comprehension):

```python
    for dataset_language in dataset_languages:
        if str(dataset_language) not in request.POST:
            continue

       #Take the values form the request and split
        values = request.POST[str(dataset_language)].split("\n")

        if values[0] == '':
            continue

        vals[str(dataset_language)]= sorted([value.strip() for value in values])
```

Smaller things I noticed:
* In `update.py`, there are instances of `except` without error type (so it catches everything, which is always risky).
* In `update.py`, it says somewhere `add code to delete`... is something missing here?
* In `dictionary/forms.py`, there's a bit of commented code without explanation of why it's commented out.
* `translations_to_senses.py` could use some empty lines and possibly commands, so I can parse it in separate steps
* I feel the method `get_examplestc_translations_dict_with` could use a more descriptive name... don't be afraid to make it longer if needed, we all have autocomplete



Misc:
* I've never encountered the `with atomic()` construction before, interesting... why made you use it here?

* This is quite a large pull request. I don't know how, but if in the future it is possible to split up functionality into smaller subprojects with separate pull requests, that would make the reviewer's life easier :)

susanodd:
@Woseseltops the floowing javascript file is for the Keywords Mapping page. It's only been integrated with senses recently.

> In senses_kewords_edit.js, I see html elements are being created inside strings (and I think we might do that in various other places in Signbank as well). My gut feeling says it's probably safer to use Javascript's createElement instead, but it does require more code.

I rewrote some of it as strings to pass to the jquery elements to keep the nesting correct.
I'll move some of the (dynamically generated) hidden input nodes to be data- parameters on the single input instead. That will reduce a lot of volume.

The DOM createElement won't work here. It needs to be jquery because the entire thing is dynamically generated, and continuously modified as the user updates things.
The original template is generated by Django. So all the correct ids are put in place on each element.
These are then found using jquery and then the elements modified.

susanodd:
@Woseseltops  this is mine:

`In update.py, it says somewhere add code to delete... is something missing here?
`

It's not missing per se.
I added functionality to remove senses in the Keyword Mapping when they become empty if the user has moved keywords around to other senses.
The comment would be to erase the translation (keyword) if it gets set to empty. At the moment, it ignores setting them to empty and just leaves it.

The original Translation Equivalents (on master) were continuously being erased and created again every time a user modified the Keywords in the list
The intention here is to actually replace the "keyword" of the translation object rather than generate a new translation object.

susanodd:
@Woseseltops 

`In dictionary/forms.py, there's a bit of commented code without explanation of why it's commented out.
`

This code was removed, not commented out.
It is no longer relevant since it applies to the original Translation Keywords.
This particular code was creating translations with empty keywords '' for every new gloss or morpheme in the default language.

susanodd:
@Jetske this code looks iffy:

https://github.com/Signbank/Global-signbank/blob/02fbb405587a22186a45d4b439dc329f691de9e2/signbank/dictionary/update.py#L393-L406

This query:

            for st in SenseTranslation.objects.filter(language = dataset_language):

this will retrieve sense translation objects for totally different datasets that use the same language.

(Ignore the print statement, I'm trying to figure out what is going on. But so far nothing has brought the code to this print line.)

susanodd:
@Jetske this looks very inefficient:

https://github.com/Signbank/Global-signbank/blob/9f6caa61b9c1642c343facef52db6f81c62c2ff4/signbank/dictionary/update.py#L197-L199

susanodd:
@Jetske Morpheme keywords editing disappeared.

The fields show up in the List View, and headers show up in the Detail View (but they are empty
and editing is not available).

susanodd:
@Woseseltops review:

> In update.py, it says somewhere add code to delete... is something missing here?

This has been implemented.

susanodd:
> [#884 (comment)](https://github.com/Signbank/Global-signbank/issues/884#issuecomment-1613111014) may also needs some attention.

I added a setting SHARE_SENSES and prefaced the specific code with a test on this so it's skipped if not sharing.
I made the default be False.

I don't think the solution for sharing as it is now can stay. The "shared senses" literally share SenseTranslation objects and the Translation objects stored in them. However, the Translation objects have a link to gloss.

susanodd:
As per @Woseseltops suggestion I put the main function code of the update functions used by keyword mapping into a new file.
The stubs remain in dictionary/update.py

susanodd:
UPDATE

The legacy use of Translation objects that refer to (the) Keyword object with  `text=''`
is causing huge problems with consistency.
This is also causing problems with editing senses in the gloss detail view.
(There are numerous subsequent crashes with constraint violations.)
(This is caused by the empty keyword string as well as sense translation sharing.
@Jetske and @susanodd decided that the keywords should be allowed in multiple senses of the same gloss. I.e., that's the purpose of the constraint on Translation over` translation__text, language, orderIndex`
This allows the Keyword to appear in different senses of the same gloss. (See for example things that translate to the same keywords in the other language.)

I've written some functions to repair this and fix the constraint violation.
It was during implementation of the CSV routines that the inconsistencies (Sense Translations / Sense Keywords) are encountered.

susanodd:
@Woseseltops review:

> - translations_to_senses.py could use some empty lines and possibly commands, so I can parse it in separate steps

This has been done.
Oh, I see I read that as 'comments'.
It has lots of blocks and comments now.

susanodd:
For migrating the translations to senses for a dataset, this is the command sequence now in the virtual environment:

1. `python bin/develop.py delete_empty_translations  tstMH`
2. `python bin/develop.py translations_to_senses tstMH`

@Jetske needs to debug the code for shared senses
I made the default be:

`SHARE_SENSES = False`

All the Sense code works with this set to False.

I did not try out anything with Example Sentences or their videos yet.

susanodd:
@Jetske the test I made for checking consistency prints (to the console) if the senses are inconsistent.
The Keywords Mapping page also does this.

To resolve inconsistent glosses, it works to go to the Detail View and edit one of them.
The consistency is then checked again there and fixed. But if there are empty translations, that seems not to work.
It also works to go to Keywords Mapping view. (I turned the fix back on to delete empty things.)

While you're working on shared senses, this is useful to figure out when it becomes inconsistent.
You can set the delete_empty=False and leave the things inconsistent if you want.
I repaired code for editing senses in GlossDetailView (for the case when senses are not shared).
I was not able to figure out the shared senses code so I didn't do anything with that.

If you use the commands above to copy the translations data to senses:

1. `python bin/develop.py delete_empty_translations  tstMH`
2. `python bin/develop.py translations_to_senses tstMH`

The consistency problems don't seem to arise.

I tried the fix out now on a tstMH dataset with inconsistent senses to see if the fix repairs it. It does. (As described above.)

susanodd:
@Jetske , @vanlummelhuizen , @Woseseltops 

This command works quite fast for mapping the keywords to senses:

`python bin/develop.py translations_to_senses NGT
`

susanodd:
> @susanodd @Jetske @Woseseltops This pull request is getting bigger and bigger. How far are we from merging? I think we should try merge soon and just see what happens (easy for me to say, I am going on holiday soon 😄 ).

I'm wrapping up the Relations View now.
Then it can be merged as far as what I've been working on.
The "populate" method:

- python bin/develop.py delete_empty_translations  NGT
- python bin/develop.py translations_to_senses NGT

takes minimal time to complete. It needs to be done for all datasets in the virtual environment.
(It's not in a migration because there's lots of calculation.)
This needs to be done because in the new code, the "old" translation equivalents are no longer visible in the interface as they are replaced everywhere by senses. (Relations View will be finished today.)

Morphemes still use the translations. (That code has been debugged so they keep using them, plus additional checks to keep the morphems on the morpheme detail view after updates.)
(Although morphemes are a subclass they don't have handshapes and other stuff.)

susanodd:
@vanlummelhuizen I modified the Keywords CRUD test to make it test on a Morpheme instead of a gloss, since they still use the old translations.
So all the tests are successful.
Except there are no actual tests on Senses yet.
But lots of consistency checks.

Are you here next week?

susanodd:
Okay, I'm ready to merge now.

One thing that might be buggy still is setting the Sentence Type for an Example Sentence. (In Gloss Detail View).
I'll check that now.

I say go ahead and merge. This pull request has become much larger because the other issues that rely on senses are also on this branch.
But the positive thing is I've been basically continuously fixing glitches.

susanodd:
@Jetske are you ready to merge this now?

vanlummelhuizen:
I altered @susanodd 's management commands so that you don't have to run them for each individual Dataset: https://github.com/Signbank/Global-signbank/pull/1002/commits/201688427d8d35bcf56cff8c7ac036a06d941269

susanodd:
Great!!

vanlummelhuizen:
> Are you here next week?

Today is my last day. Then 3 weeks vacation. I was hoping @Woseseltops would be here so we could discuss this, but apparently he took two days off.

susanodd:
He's on vacation this week.
Let's just merge it.
I've been working on this branch for many weeks now.
We can merge it but wait to deploy it.

susanodd:
I can deploy it on Monday.

susanodd:
Okay, hopefully @Woseseltops is back next Thursday.

Jetske:
The update functions for adding/editing/removing senses in gloss detail view have become buggy again and I think that's better to fix before merging (I'll work on it now). And we should probably decide on whether to share senses or not before running the populate commands. 
Maybe we can run through it together on Monday @susanodd and then merge?

susanodd:
@Jetske yes it's a plan!

susanodd:
@Jetske the senses typeahead doesn't work correctly.
It finds keywords from other glosses.
It needs to have a setting SHARE_SENSES  condition on it.
Or it needs to only look at keywords, not the senses they appear in.
I tried it out on a gloss that had no senses and it was immediately inconsistent.

Is it possible to use the senses typeahead to add senses from other glosses to the example sentence?
So on actually picks the correct sense of the other gloss that is also used in the example sentence?

@vanlummelhuizen and I discussed examples of sense sharing. He has the idea to make it also work between datasets and languages, but leave the translation keywords to the researchers of the dataset. So match at the level of Sense not SenseTranslation.
So say if you have a sentence, then you can pick the senses of the words in the sentence.

I've noticed for NGT there are a lot of "keywords" that contain commas. So there needs to be a lot of work on NGT to make the senses explicit. (The keywords are now all in one sense, but if you just map that into shared senses before they are split up, then it's a potential problem.)

If possible, the populate code could be turned into a page where the user can confirm if gloss senses are actually the same thing.

susanodd:
@Jetske also using dataset_lang.name inside of the template 'id' tag is a problem because some contain spaces. Mostly use the language_code_2char instead.
In the keywords mapping code, I used the actual object id of all the different typed objects in sequence in order to force the ids to be unique.

susanodd:
@Jetske I repaired the delete_ sense update function. Because things are deleted inside of an iteration, there needs to be a separate list that is iterated over because you can't modify the thing being iterated over while iterating.

I also repaired the keywords mapping so it works properly with the gloss edit senses functionality.
I hadn't realised that newly created senses with only one language in the Gloss Edit View only have one SenseTranslation object for the new (filled in) language.

susanodd:
@Jetske during consecutive edits, the url keeps getting appended with additional` ?edit` 's

susanodd:
@Jetske I made changes here and there. I have Shared turned off. (@vanlummelhuizen also used it with it turned off.)
I kept switching between the Gloss Edit view and the Keywords Mapping view to make sure it behaves the same way as you edit the senses.
You'll notice some changes to the delete sense function. I debugged the delete and the reorder in the template.
(Again, for the case sharing is off.)

susanodd:
@Jetske I had to revise the migration that creates the 0 and 1 SentenceType's.
It didn't work properly with changing the setting nor with allowing update.
I also had to complete the FC checks in Admin to check that a sentence type isn't in use, thus to prevent deletion.
In the migrations, the language settings are not available!! So the code had to be hard-coded with the `name_` field name for all the languages. That's why it wasn't working.
My strategy for not requiring all the languages in the FC Admin didn't work correctly.
We need to ask @vanlummelhuizen about this. It needs the languages to make sure all the interface languages show up. But it didn't work to make them non-required in the form. (So annoying when adding a new sentence type. But don't know how to not do that.)

susanodd:
I edited the migration, since it's only the function part that was changed. Not the actual fields.
But you'll need to run the migration again. Although can also wait, since Django will have it marked as already being run.
I'll do the merge and deployment tomorrow.

PR #1015, opened 2023-07-26
Sorting fixed for senses in glosslist view #1012

Jetske:


susanodd:
@Jetske can you keep the class names `translation_` instead of using `sense_` ?

The query parameters code shares the gloss row template and it uses the class name everywhere to hide and reveal, etc.

susanodd:
Looks good!

PR #1018, opened 2023-07-27
Videos view add sensevideo 963

Jetske:
An overview for the videos given for example sentences for senses was added to the Videos tab of glosses

I also changed the indentation a little to improve the readability, so there are a little more edits all over the file


susanodd:
I tried it locally and no errors.
I can't test it completely because I don't have sufficient videos.
I'll merge it.

PR #1033, opened 2023-08-10
Show similar senses in modal

Jetske:
This doesn't share senses (yet, there is a url for that if ever we want to add that) but just shows which senses are more or less similar. Therefore, it should not close the issue to which the branch was linked.


susanodd:
It looks good, @Jetske !

In order to review it I needed to merge it with master. (This has been pushed to the branch.)
Then I adjusted with position and width of the modals since they were way off to the right.
I increased the width for multiple languages and checked it on Kata Kolok.

PR #1048, opened 2023-08-23
Prettier and faster

susanodd:
Some streamlining of the Sense methods so similar methods share code.
To accomplish this, parameters have been added to the methods.

### DIFF
--- a/signbank/dictionary/models.py
+++ b/signbank/dictionary/models.py
@@ -695,18 +695,6 @@ class Sense(models.Model):
     def get_dataset(self):
         return self.glosses.first().lemma.dataset
 
-    def get_example_sentences_orig(self):
-        "Return the example sentences for this sense for every language as one string"
-        translations = []
-        for dataset_translation_language in self.get_dataset().translation_languages.all():
-            sentences = []
-            for exampleSentence in self.exampleSentences.all():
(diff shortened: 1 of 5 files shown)


susanodd:
I've been using the GlossSense to look things up (also elsewhere) because it seems like it's more a straightforward lookup that a query. Then it just looks up the sense in that table.

I am not sure when Django constructs all the sql queries that end up in the template. I was trying to force it to actually evaluate things rather than make queries.

I saw the get_dataset method. But that in turn is a query as well.
Probably the best way to tell is using the debug setting in PyCharm and step through to find out how many steps it goes through (in the underlying implementation) to obtain the result.
Ultimately they all yield the same result. So it would be whatever the step-through debug reveals, if anything useful.
I suspect using GlossSense is less, but I have no idea.

susanodd:
Because Morpheme is a subclass of Gloss, it would mean that you would need to consistently check to make sure the object is a gloss rather than a morpheme when using senses.
At the moment, they're just not inserted into the tables because if you go to MorphemeEditView it has different code. The get method of the view checks to make sure it is a Morpheme and the same is done with Gloss.
I suspect lots of things would break if Morpheme was all of a sudden a totally different model not related to Gloss. (The scroll bar also checks that things are morphemes or gloss to give the correct link to the detail view based on type. They can exist in the same search result if a user does Sign or Morpheme.)

PR #1049, opened 2023-08-23
SenseExampleSentence added, unshare examplesentences and order added for sentences

Jetske:
This should work. But to be sure maybe first test run this locally :)

To migrate but keep the existing through table:
1. First run the first migration 0076
2. Uncomment all lines in model SenseExamplesentence
3. Run the second migration 0077


Jetske:
@susanodd There are also some small edits for senses like checking if the sense is already there before creating a new sense and adding a sense with gloss.senses.add instead of making a GlossSense object, but while testing I got an error for deleting a sense: something with the unique constraint of senses (maybe order/orderIndex/.?). I haven't been able to replicate it and _think_ it was already there so just so you know.

susanodd:
Okay, I'll look at it. I'm used to checking errors from constraints. Hence all the consistency check methods.
The unfortunate thing is that Django enforces/checks the consistency only on data that existed after the constraint was added.
So there can be some old data that is messed up.

I modified the Sense display methods in a different pull request.

susanodd:
@Jetske you probably want to add spaces to the verbose names. Those are the user friendly names.
(For Sense Example Sentences)

susanodd:
You commented out the code. But probably you need to leave the original model class and then "populate" (map) the original ones to new ones by inserting them so they get the proper id field and order.

You probably need to do this by trial and error and modify a new migration that populates this.
Until it does what you need it to do.
(Like don't push the populate thing since you may need to modify it.)
(And keep making a database backup for each try. Then examine the database data.)

Jetske:
So I don't fully understand it but I followed the steps given [here](https://stackoverflow.com/questions/33257530/how-to-add-through-option-to-existing-manytomanyfield-with-migrations-and-data-i), adding SeparateDatabaseAndState into the migration. To my understanding it renames the old through table to the new one, and so the sentences in it are preserved. It's explained [here](https://stackoverflow.com/questions/61683766/what-is-a-state-operation-and-database-operation-in-separatedatabaseandstate). And that means you wouldn't need to populate it again.

susanodd:
The second migration adds the order field.
(This is commented out between the two migrations.)

But when this field is added, it is also added to existing sentences. So they get the default value.
It basically adds a column to the database table and puts the default in the new column for existing rows.

Jetske:
Yes, that's why it was separated into first a step that does some magic to the through table, and then a step that adds the order field to that table. I don't think it's safe to do that in one go. 

The order will be 1 for all sentences that already existed but that is probably fine, at least until we add a way to change the order. When an examplesentence is deleted the order gets adjusted for all its other examplesentences.

signbank/dictionary/models.py
```diff
@@ -782,12 +787,45 @@ def get_senses_with_similar_sensetranslations_dict(self):
...
+class SenseExamplesentence(models.Model):
+    """An examplesentence belongs to one or multiple senses"""
+
+    sense = models.ForeignKey(Sense, on_delete=models.CASCADE)
+    examplesentence = models.ForeignKey(ExampleSentence, on_delete=models.CASCADE)
+    # order = models.IntegerField(
+    #     verbose_name    = _(u'Order'),
+    #     help_text           = _(u'What order to display this examplesentence within the sense.'),
+    #     default = 1
+    # )
```
vanlummelhuizen, line 818: I just tried to migrate 0076 and 0077 with these line uncommented and that works without problems.

susanodd:
@Jetske I thought we decided that there can be duplicate sentences so that users can have multiple videos for the same sentence (text), since the video is identified by the sentence ID as the file name.

Jetske:
> @Jetske I thought we decided that there can be duplicate sentences so that users can have multiple videos for the same sentence (text), since the video is identified by the sentence ID as the file name.

I see, I thought we were talking about duplicate sentences over the dataset that would be allowed, but not within a sense. But  if eventually we do allow senses to be shared between glosses then allowing duplicate sentences would indeed be better. I'll change it.

susanodd:
@Jetske your question about the senses becoming inconsistent.
https://github.com/Signbank/Global-signbank/pull/1049#issuecomment-1689729551

Perhaps we need a command (for the virtual env) that checks to see if there are senses not related to any glosses? (Hopefully that is not the case.)

With the command you suggested (#1048) as a way to obtain the dataset:
self.glosses.first().lemma.dataset

Can it happen that self.glosses is empty? If so, then this will fail.

susanodd:
I've been using the GlossSense table to check consistency.
For example if there are duplicates with the same sense number or that (at the moment) a sense is paired with two different glosses. (That is an inconsistency because the actual Translation objects linked to via the sense translations have a gloss field so they cannot be shared.)

Jetske:
@susanodd is the gloss field in translations still necessary though, as the translations are now linked to a gloss through senses? I don't know if translations are used in other places, but otherwise I would think that that field isn't used any longer. 

At least in the update delete_sense method it completely deletes the sense object if it is not related to another gloss, so in theory it should never happen that self.glosses is empty. But indeed it doesn't harm to have a command to check that..

susanodd:
> @susanodd is the gloss field in translations still necessary though, as the translations are now linked to a gloss through senses? I don't know if translations are used in other places, but otherwise I would think that that field isn't used any longer.
> 
> At least in the update delete_sense method it completely deletes the sense object if it is not related to another gloss, so in theory it should never happen that self.glosses is empty. But indeed it doesn't harm to have a command to check that..

Yes, the gloss field is necessary!!
Morphemes still use Translation objects since they do not have senses.
Django maintains that table and automatically inserts things in it. It still works to do 
gloss.translation_set.all() 
in spite of the senses. It's put things in that table, too when you update the translations of the senses.

gloss.senses.all()
can be empty if a gloss has zero keywords.

(We had exactly this problem with the FieldChoice model because a field can be None (never filled in) which is not the same as FieldChoice machine value 0 (also meaning not chosen).
(It matters in the database because Null is not the same as 0. That's why there are numerous 
string_variable = gloss.field.name if gloss.field else '-'

susanodd:
I put the followup in the other pull request

https://github.com/Signbank/Global-signbank/pull/1048#issuecomment-1691447006

susanodd:
@Jetske I put a backup of the newest database in the usual place.

signbank/dictionary/models.py
```diff
@@ -782,12 +787,45 @@ def get_senses_with_similar_sensetranslations_dict(self):
...
         for sensetranslation in self.senseTranslations.all():
             str_sense .append(str(sensetranslation))
         return " | ".join(str_sense)
+    
+class SenseExamplesentence(models.Model):
+    """An examplesentence belongs to one or multiple senses"""
+
+    sense = models.ForeignKey(Sense, on_delete=models.CASCADE)
+    examplesentence = models.ForeignKey(ExampleSentence, on_delete=models.CASCADE)
+    order = models.IntegerField(
```
vanlummelhuizen, line 814: I am ok with this explicit implementation of order. I just want to point out that there seems to be a Django-native solution for this: https://docs.djangoproject.com/en/4.2/ref/models/options/#order-with-respect-to.

vanlummelhuizen (commented):
Nice work! 

All code seems ok except perhaps for the repeated code. About the order_with_respect_to and Signal comments: see whether you understand those concepts and when you do, consider if it would work here and whether it would help readability/maintainability of the code.

Jetske:
> All code seems ok except perhaps for the repeated code. About the order_with_respect_to and Signal comments: see whether you understand those concepts and when you do, consider if it would work here and whether it would help readability/maintainability of the code.

@vanlummelhuizen I see how order_with_respect_to would work for a one-to-many relationship, but not for many-to-many as is the case here. Signal might be useful!

Jetske:
Say we have this situation:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/s4367707/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/s4367707/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

</head>

<body link="#0563C1" vlink="#954F72">



Sense | exampleSentences
-- | --
1 | 1
2 | 2, 1, 3
3 | 3, 2



</body>

</html>


with through table:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/s4367707/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/s4367707/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

</head>

<body link="#0563C1" vlink="#954F72">



Sense | Examplesentence | order
-- | -- | --
1 | 1 | 1
2 | 2 | 1
2 | 1 | 2
2 | 3 | 3
3 | 2 | 2
3 | 3 | 1



</body>

</html>

The order field is about the order in which the sentences are shown in the sense, thus it is specific to every relation, e.g. as in above: sense(2) has sentence(3) with order(3). However, in sense(3), this same sentence has order(1). 

Basically, order is only relevant if it has been changed (currently not possible) or if it's shared between senses (also currently not possible) but I still think it's good to keep track of it because then no changes to the models are necessary to add that functionality. When left out I think they're just ordered by the order in which they were added to the sense, or in case of using order_with_respect_to in the order of sense/sentence id's probably.

vanlummelhuizen:
> The order field is about the order in which the sentences are shown in the sense

Exactly. The order in which senses are shown for an example-sentence is not relevant. So, order_with_respect_to could be applied to the ForeignKey-field 'sense' in SenseExamplesentence, leaving the rest of the model, and also the ManyToMany-field with through-table of Sense, intact.

One thing that might not work very nice is the add function of the ManyToMany-field on the Sense side. See e.g. https://stackoverflow.com/questions/65887820/how-to-take-advantage-of-m2m-relations-when-using-order-with-respect-to-in-an-in

Bottomline: leave everything as it is now and revisit when order is properly used (other that reordering when a example sentence is remove from a sense).

vanlummelhuizen (approved):
Looks good!

susanodd:
@Jetske did you confirm that this doesn’t break the CSV import and export of sentences?

vanlummelhuizen:
> @Jetske did you confirm that this doesn’t break the CSV import and export of sentences?

We will ignore your comment until you are back from your vacation, @susanodd :wink:

signbank/dictionary/update.py
```diff
@@ -358,6 +358,45 @@ def sort_sense(request, glossid, order, direction):
...
+                senseexamplesentencebeneath = SenseExamplesentence.objects.get(sense=sense, order=order+1)
+            except (ObjectDoesNotExist, MultipleObjectsReturned):
+                print('sort_examplesentence DOWN: multiple or no match for order: ', senseid, str(order+1))
+                messages.add_message(request, messages.ERROR, _('Could not sort this examplesentence.'))
+                return HttpResponseRedirect(reverse('dictionary:admin_gloss_view', kwargs={'pk': gloss.id}))
+            senseexamplesentencebeneath.order = order
+            senseexamplesentence.order = order + 1 
+            senseexamplesentence.save()
+            senseexamplesentencebeneath.save()
+    except (ObjectDoesNotExist, MultipleObjectsReturned, DatabaseError, TransactionManagementError):
```
vanlummelhuizen, line 395: `ObjectDoesNotExist` and `MultipleObjectsReturned` are handled in the respective try-except clauses above. Perhaps also move `DatabaseError` and `TransManagementError` to those try-except clauses?

While we are at it: I think you could squash the if-else clause by first determining what should be added to *order* depending on *direction* (either 1 or -1) and then use that variable. This enables you to remove half of the code. Printing 'UP' of 'DOWN' can be done with `direction.upper()`.

Also test whether *direction* is 'up' of 'down' (not 'sideways' or 'bla' or whatever).

Jetske: Thanks, I rewrote it as well as the function for ordering senses which it was copied from

PR #1053, opened 2023-09-15
Don't show annotation language parentheses by default

Woseseltops:
This is a super small thing, but I'd like a thumbs up from at least one of you before I merge to master. 

The context:
* ASL signbank doesn't use our dataset functionality, as they have only 1
* As a consequence, they also have no 'translation language' for 'annotation ID gloss'
* This resulted in weird empty parentheses in the gloss detail view

image

### DIFF
--- a/signbank/dictionary/templates/dictionary/gloss_detail.html
+++ b/signbank/dictionary/templates/dictionary/gloss_detail.html
@@ -862,7 +862,8 @@ option:not(:checked) {
             {% for lang, annotation_idgloss_translation in annotation_idgloss.items %}
             <tr>
                 <th>
-                    {% trans "Annotation ID Gloss" %} ({{lang}})
+                    {% trans "Annotation ID Gloss" %}
+                    {% if SHOW_DATASET_INTERFACE_OPTIONS %} ({{ lang }}) {% endif %}
                 </th>
                 <td class='edit edit_text' id='annotation_idgloss_{{lang.language_code_2char}}' colspan="2">{{annotation_idgloss_translation}}</td>
                 <td>


signbank/dictionary/templates/dictionary/gloss_detail.html
```diff
@@ -862,7 +862,8 @@ <h2 id='modalTitleDelete'>{% trans "Delete This Sign" %}</h2>
             {% for lang, annotation_idgloss_translation in annotation_idgloss.items %}
             <tr>
                 <th>
-                    {% trans "Annotation ID Gloss" %} ({{lang}})
+                    {% trans "Annotation ID Gloss" %}
+                    {% if SHOW_DATASET_INTERFACE_OPTIONS %} ({{ lang }}) {% endif %}
```
vanlummelhuizen, line 866: I am wondering whether `{% if lang %}` would also work and whether that would be better? Of perhaps `{% if SHOW_DATASET_INTERFACE_OPTIONS and lang %}`?

susanodd:
@Woseseltops what is in the dictionary data structure annotation_idgloss for ASL?

`   {% for lang, annotation_idgloss_translation in annotation_idgloss.items %}`
`            <tr>`
`                <th>`
`                    {% trans "Annotation ID Gloss" %} ({{lang}})`

Are there no keys?

susanodd:
@Woseseltops wrote:

> 
> > Probably other problems will arise because of model translations. Some code uses lang.name and a language object is used.
> 
> I might be mistaken, but aren't you confusing translation languages (linked to datasets) and interface languages here?

I was thinking of the Senses editing modals.
The code generates string dictionaries (see methods in models).
They explicitly use the language name and it’s displayed inside the Edit modals.
The language.name is used for datasets with more than one language. But also for showing the translation to the interface language.
So yes, if you switch to Chinese and start doing stuff with NGT then the languages Dutch and English (ought to) show up in Chinese.
But the actual “textual” name it used in the Senses Edit modal.

So probably a that Senses code would need to be rewritten otherwise English will show up. I don’t know if ASL uses the Senses yet.

Some code passes language objects to the template. I guess if you just write {{language}} you’re implicitly invoking the __str__ method on the object.

susanodd:
@Woseseltops if you look at the screenshot of the search panel in #1053 I made it so the language name is hidden if there is only one language.

However, this might be confusing if your interface is set to Chinese and there is an Annotation (written in Chinese) and your selected Dataset is ASL.

Maybe all the tests need to check how many model transition languages and how many translation languages are active.
And some logic for mismatches.

Because of this issue I’ve been trying to avoid introducing the (language) when only one language is used. But probably an additional setting needs to be checked?
Hoping to make it possible to use the same Signbank code for ASL with just settings.

PR #1054, opened 2023-09-20
Fix for ordering of sentences that already existed

Jetske:
Small fix: if sentences already existed before the last migration they have the default order 1 so this gave a problem when moving one up or down. Instead they first need to be given the right order number and then can be moved just fine.

### DIFF
--- a/signbank/dictionary/update.py
+++ b/signbank/dictionary/update.py
@@ -354,9 +354,7 @@ def sort_examplesentence(request, senseid, glossid, order, direction):
     gloss = Gloss.objects.get(id=glossid)
     sense_examplesentences_matching_order = SenseExamplesentence.objects.filter(sense=sense, order=order).count()
     if sense_examplesentences_matching_order != 1:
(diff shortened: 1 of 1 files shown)


vanlummelhuizen:
I think the code looks good. 

As I understand it, the order is fixed when a user requests to sort example sentences. Would it make sense to go over all SenseExamplesentences and fix ordering once, for instance in a data-migration?

Jetske:
Yes it would make sense to just run reorder_examplesentences for every sense once and then this check shouldn't be necessary

vanlummelhuizen:
@Jetske I think this check is still valuable even after going over all example sentences. Can you construct this data-migration I mentioned? Perhaps this helps: https://medium.datadriveninvestor.com/django-data-migrations-1e13fdb916d4

susanodd:
This works correctly.

However, it is necessary to click on one of the up or down arrows (only for test dataset tstMH) for glosses that have multiple sentences that all have number 1. This gets them renumbered.

PR #1058, opened 2023-09-27
Gloss detail error

Jetske:


Jetske:
I wanted to fix it quickly and thought double check would at least be complete, but indeed I see that 'if not' includes the check for None and more. So it can simply be `if not trans and exclude_empty:`

susanodd:
@Jetske I modified the senses in tstMH so that all the senses that only had one language filled in, now include text for both languages. (That made the bug go away -- temporarily until it's properly fixed.)
So it is something to do with one language not filled in.
(I repaired them in the Map Keywords page by updating the Matrix to fill in data.)

There were no empty translations in NGT. So nothing changed. Only Norwegian had empty translations.

PR #1069, opened 2023-10-11
Search sentences

susanodd:
This is more or less finished.

[There is some "extra" code related to query parameters that is at the moment superfluous.
I revised the code to allow the query in url from the search history, then there can be a button to apply a saved query to the senses search (i.e., of the glosses in GlossSense... Preliminary, this works, just need the button...)


susanodd:
Introduced new issue per review of @vanlummelhuizen #1071

PR #1070, opened 2023-10-11
Make the sense translations line up right

Jetske:


susanodd:
I merged this by hand with the master to resolve conflicts.
As far as I can tell, it works properly.

Issue #1074, opened 2023-10-16
Form Reset not working


susanodd:
Gloss Search Form does not reset the (non-multi) select fields

susanodd:
NOTE TO SELF: This seems to be tied to the re-filling of the previous form fields.

For example, if I search on repeat=True, then the repeat field is filled in.
And if I try to Reset the form, the option of the select is set to the first (-------) choice, but then it gets reset again to True.

If you're busy constructing a query and you hit Reset, then it works as expected.
But if it's displaying the results of a previous query, only the multi-select fields can be erased. Not the normal select.

susanodd:
I fixed this. It needed to be rewritten in jQuery rather than as DOM elements.

susanodd:
I did the same for Gloss Search.
Plus repaired the broken language toggle buttons.

susanodd:
For Handshapes, the Reset search form works. Although it chooses to reload the page and show all the handshapes again. For the Finger Selection 2, if you choose False (radio button) it works the same as "nothing".
I'm not sure what this ought to be. If you choose True then there is only a handshape with Finger Selection 2 filled in. There is only one handshape with Finger Selection 2 filled in.

@ocrasborn can you try out the Handshape search and see if Finger Selection 2 is doing what you expect?

susanodd:
The Reset has been fixed for Handshape Search. Although it seemed to be working , it kept reloading the entire list of handshapes instead. I revised to URL to that if you are searching it goes to that url rather than the url for show all. Now the reset clears the form and sets the results to empty.

susanodd:
I'm working on Morpheme Search now.  The Reset is working. But the search fields of the form are cleared during the query. It looks like the template search fields are not linked to the search form. They are available in the url as get parameters, but end up disappearing from the form.

The template and python code are programmed very weirdly.
(Related to #1071.) There's too much "topic" fragmentation of the code. I can't even find the "search form fields" in the template.

susanodd:
The  Morpheme Search Reset has been repaired. And search parameters show in the form after running the query.

(New function to support this with less code.)

In one place, there is code specific to ASL (for MorphemeListView). I rewrote the code to be much shorter. But it's kind of weird the code is here.

susanodd:
I removed the code that refers to the ASL fields since they are not in the MorphemeSearchForm.

susanodd:
This is complete and awaiting merging and deployment.

susanodd:
This is live.

PR #1079, opened 2023-10-27
ClearForm Reset and Streamline Handshapes and Morphemes

susanodd:
- This fixes the form reset code and makes sure the search parameters just done are still visible in the search form.
- The "yes/no" choices have become translated for morpheme search and the values homogenized.
- Obsolete Django syntax (the `u'`) has been removed from `forms.py`.
- Processing of the "get" parameters has been moved to a separate function to reduce the code in the get_queryset
- To allow sharing of the search form between the get_context_data and the get_queryset, the form has been made a self field of MorphemeListView.
- This allows the code that processes the "get" queryset to receive the search form.
- In order to process the "search form" fields by iterating over the search form, the language fields have been made fields instead of attributes.

### DIFF
--- a/signbank/dictionary/adminviews.py
+++ b/signbank/dictionary/adminviews.py
@@ -62,7 +62,7 @@ from signbank.dictionary.update_senses_mapping import delete_empty_senses
 from signbank.dictionary.consistency_senses import consistent_senses, check_consistency_senses, \
     reorder_sensetranslations, reorder_senses
 from signbank.query_parameters import convert_query_parameters_to_filter, pretty_print_query_fields, pretty_print_query_values, \
-    query_parameters_this_gloss, apply_language_filters_to_results, search_fields_from_get
+    query_parameters_this_gloss, apply_language_filters_to_results, search_fields_from_get, queryset_from_get
 from signbank.search_history import available_query_parameters_in_search_history, languages_in_query, display_parameters, \
     get_query_parameters, save_query_parameters, fieldnames_from_query_parameters
 from signbank.frequency import import_corpus_speakers, configure_corpus_documents_for_dataset, update_corpus_counts, \
@@ -2626,7 +2626,8 @@ class MorphemeListView(ListView):
         context['field_colors'] = choices_colors
 
         if hasattr(settings, 'SEARCH_BY') and 'publication' in settings.SEARCH_BY.keys() and self.request.user.is_authenticated:
-            context['search_by_publication_fields'] = searchform_panels(search_form, settings.SEARCH_BY['publication'])
+            context['search_by_publication_fields'] = searchform_panels(search_form,
+                                                                        settings.SEARCH_BY['morpheme_publication'])
         else:
             context['search_by_publication_fields'] = []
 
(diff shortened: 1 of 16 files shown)


susanodd:
@vanlummelhuizen perhaps you can take a look at the two new functions at the end of` query_parameters.py` ?

For Morpheme Search I moved the code that deals with the Search Form gets.

https://github.com/Signbank/Global-signbank/blob/90128e0ce94206624a3b3b1195c87ca2df144e17/signbank/query_parameters.py#L500

The first function is used to pass the already set search form parameters back to the View after the user does a query so the parameters remain visible. (Called inside get_context_data)

https://github.com/Signbank/Global-signbank/blob/90128e0ce94206624a3b3b1195c87ca2df144e17/signbank/query_parameters.py#L529

The second function filters the query based on the get parameters (Called inside get_queryset)

The first function arose out of #1074.

Next, to do (2) for Gloss Search, adding an additional parameter to the second function for the Form type.

susanodd:
Note: Based on the Query Parameter setup (code in `query_parameters.py`), it does not work to include the language queries inside the creation of a Q filter. For some reason, their order of evaluation affects the results. Thus they are not done inside the functions that process the other "get" parameters.

susanodd:
The code of the morpheme search form has been simplified.
With the goal that the different search forms can share code in functions instead of duplicating it in all the init functions.

susanodd:
The code of the morpheme search form has been simplified.
With the goal that the different search forms can share code in functions instead of duplicating it in all the init functions.

susanodd:
@Woseseltops @vanlummelhuizen can you take a look at the new functions at the end of `query_parameters.py`?
I’m trying to condense code and move the code that deals with the get parameters out of the get_query set and get_context_data functions because that code is totally unreadable.
I just made a new function for the Senses search over GlossSense.
The query filters are different for this one.
However, the code is basically the same.
Could uses one function that passes in the Model name also, so if it’s GlossSense then the filters get ‘`gloss__`’ added to them.
But then I made need to revised the handling of the search form (as is done for MorpheneListView.

I like the code myself since it’s more condensed. But need to apply some of the @vanlummelhuizen magic to the context data that’s not the search form.

susanodd:
Regarding merging, at the moment, this pull request only affects HandshapeListView / Handshape Search and MorphemeListView / Morpheme Search.

A draft function has been written for Sense queries over GlossSense, This function is not used yet because that will impact code in SenseListView.

susanodd:
So I revised the GlossSearchForm, GlossListView and the SenseListview to use the same search form initialisation set up as the MorphemeListView.

susanodd:
This is ready for review (again).
All the Reset Form buttons work on Gloss, Morpheme, Sense, Handshape search.
All the fields of the previous search show up in the search form until you Reset or modify them.
The search forms of GlossListView, SenseListView, MorphemeListView are all self fields of the view class.
This allows the initialisation code to be shared (for example field choice choices and language fields).

vanlummelhuizen:
@susanodd Good work! And you have done a lot. 

I have two things right away:
* this pull request is pretty big and many of its commits are also not small. My advise is to try to keep commits small and pull request not too big. Kudos for the elaborate commit messages!
* in at least `query_parameters.py` you have comments at the beginning of functions. Could you turn them into docstrings?

susanodd:
Okay, I'll add the docstrings.

The pull request didn't start out so big. Originally it was only the Morphemes and Handshapes. :)
Then I kept working on it.
I only did commits in phases when it was more or less working, that's why they are so big.

susanodd:
I added the docstrings. It helps clarify what the intent of the functions are.
Also, removed some unused parameters, obvious when documenting.

susanodd:
@vanlummelhuizen do you agree this can be merged?

signbank/query_parameters.py
```diff
@@ -493,3 +494,503 @@ def pretty_print_query_values(dataset_languages,query_parameters):
...
+        elif get_key in ['search', 'translation']:
+            from signbank.tools import strip_control_characters
+            val = strip_control_characters(get_value)
+            search_fields_to_populate[get_key] = escape(val)
+            search_keys.append(get_key)
+        else:
+            search_fields_to_populate[get_key] = get_value
+            search_keys.append(get_key)
+
+    return search_keys, search_fields_to_populate
```
vanlummelhuizen, line 531: @susanodd Can't you just remove `search_keys` and just do `search_fields_to_populate.keys()` instead, preferably where this function is called, or even in the relevant template?

signbank/dictionary/adminviews.py
```diff
@@ -1226,229 +991,22 @@ def get_queryset(self):
 
             qs = qs.filter(query)
 
+        qs = queryset_glosssense_from_get('GlossSense', GlossSearchForm, self.search_form, get, qs)
+        # this is a temporary query_parameters variable
+        query_parameters = dict()
+        # it is saved to self.query_parameters after the parameters are processed
+        query_parameters = query_parameters_from_get(self.search_form, get, query_parameters)
```
vanlummelhuizen, line 998: This could be condensed to `query_parameters = query_parameters_from_get(self.search_form, get, dict())` or even replace `dict()` with `{}`.

susanodd: No, the query parameters are passed along as a state....
There are more modifications later.
The {} format does not work correctly with dictionary_variable.keys()

signbank/query_parameters.py
susanodd: See the comment below. Somehow if a dictionary is empty, doing dictionary.keys() gives an error.

signbank/dictionary/forms.py
```diff
@@ -930,12 +763,19 @@ def __init__(self, queryDict, *args, **kwargs):
...
+        count_languages = len(languages)
         for language in languages:
             # and for LemmaIdgloss
             lemma_field_name = self.lemma_search_field_prefix + language.language_code_2char
-            setattr(self, lemma_field_name, forms.CharField(label=_("Lemma")+(" (%s)" % language.name)))
+            if count_languages > 1:
+                lemma_label = _("Lemma")+(" (%s)" % language.name)
+            else:
+                lemma_label = _("Lemma")
+            setattr(self, lemma_field_name, forms.CharField(label=lemma_label))
```
vanlummelhuizen, line 774: This could be written as:
```python
lemma_label = _("Lemma")+(" (%s)" % language.name) if count_languages > 1 else _("Lemma")
```

There are more places where something similar could be done.

vanlummelhuizen:
> do you agree this can be merged?

I made a few minor remarks. Other than that, it seems ok. But, since this is such a vast PR, I could not review everything in detail.

signbank/query_parameters.py
vanlummelhuizen: Perhaps it needs to be cast to a list somewhere:

```python
>>> {}.keys()
dict_keys([])
>>> dict().keys()
dict_keys([])
>>> list({}.keys())
[]
```

signbank/dictionary/adminviews.py
vanlummelhuizen: Also see my comment above.

All you do here is assign an empty dictionary to a variable and immediately after, you pass that variable to a function. At that point it is still only an empty dictionary. Finally you assign the return variable of function to that same variable, overwriting the initial value, which was an empty dictionary.

signbank/dictionary/forms.py
susanodd: Ouch! Yes, I see. I am in the habit of doing them straight-line because I don't trust that the model translations and `_()` expressions are going to be combined correctly. (I made an issue about this a long time ago (which nobody responded to, hence closed), because I had to flatten out some comprehensions because they didn't work with non-ascii languages. So superstition makes me code them imperatively to force evaluation order.)

signbank/dictionary/adminviews.py
susanodd: Yes, I understand. But I am more comfortable doing it explicitly. Then it matters less if some code ends up being inserted before the function call but after creation of the query_parameters dict.

This is just passing the state around as argument and result. (This is a style thing.)

signbank/dictionary/forms.py
susanodd: Okay, I've revised that. It looked hyper-paranoid from me. Like I don't trust the interpreter.

vanlummelhuizen:
@susanodd You can merge this now.

PR #1082, opened 2023-11-09
Make Gloss Detail panels clickable across heading

susanodd:
as is done for Search Panels. Removed excess divs. Problems with Django if-else-endif mismatches so some minor adjustments.

### DIFF
--- a/signbank/dictionary/templates/dictionary/gloss_detail.html
+++ b/signbank/dictionary/templates/dictionary/gloss_detail.html
@@ -171,6 +171,8 @@ function checkKey(e)
 </script>
 
     <script type='text/javascript' src="{{STATIC_URL}}js/gloss_edit_color.js"></script>
+    <script type="text/javascript" src="{{STATIC_URL}}js/drag_drop_files.js"></script>
+    <script type="text/javascript" src="{{STATIC_URL}}js/recordscript.js"></script>
 
     <!-- Expand and collapse all panels -->
     <script type='text/javascript'>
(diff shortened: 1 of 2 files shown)


susanodd:
Need to so for morphemes too. (As a separate commit.)

susanodd:
Morphemes are done.

signbank/dictionary/templates/dictionary/gloss_detail.html
```diff
@@ -171,6 +171,8 @@
 </script>
 
     <script type='text/javascript' src="{{STATIC_URL}}js/gloss_edit_color.js"></script>
+    <script type="text/javascript" src="{{STATIC_URL}}js/drag_drop_files.js"></script>
+    <script type="text/javascript" src="{{STATIC_URL}}js/recordscript.js"></script>
```
vanlummelhuizen, line 175: @susanodd Could you point out where these Javascripts are used?

susanodd: Oh, I replied below. They are used in the record video routines. They were inside of a permission-prefaced block with lots of nested blocks inside it.

susanodd:
I removed one of these:

   {% if perms.dictionary.view_advanced_properties %}

from gloss_detail.html
That's the one that wasn't parsing.
There are way too many perms things inside of GlossDetailView. It's very difficult for a human to parse.

susanodd:
@vanlummelhuizen do you agree this can be merged?

susanodd:
Nice idea!!

vanlummelhuizen:
@susanodd Nice alternative for changing the mouse cursor. I think this can be merged.

Issue #659, opened 2020-06-18
Review and revise the process of saving info when deleting a gloss or media.


vanlummelhuizen:
The code below may need some revision because of the use of LemmaIDGlossTranslation and AnnotationIDGlossTranslation. 

https://github.com/Signbank/Global-signbank/blob/853bc4f736b2143097109b0414461c1da12e0461/signbank/dictionary/models.py#L1788-L1810

susanodd:
@vanlummelhuizen are you working on this issue? If not, I will work on it.

This one is also related to #440. That has evolved into Other media paths and filenames.

vanlummelhuizen:
@susanodd I am not working on this issue so please do. I think the main thing is that idgloss and annotation_idgloss have a many-to-one relationship with Gloss while the code above only takes a default one.

susanodd:
Hmmm. What is the intention of the DeletedGlossOrMedia history?
It would seem the Dataset would also be useful?
The code that creates objects of this model is:
https://github.com/Signbank/Global-signbank/blob/d218da656e6a7d02dd547a227efc6f4c90c637aa/signbank/dictionary/views.py#L2145
https://github.com/Signbank/Global-signbank/blob/d218da656e6a7d02dd547a227efc6f4c90c637aa/signbank/dictionary/models.py#L1871
https://github.com/Signbank/Global-signbank/blob/d218da656e6a7d02dd547a227efc6f4c90c637aa/signbank/video/views.py#L71

But the actual image or gloss gets deleted. The video gets copied to a .bak file or overwritten.

A deleted gloss cannot be restored from a DeletedGlossOrMedia object. There is no information about phonology, etc.
It would need to be restored using a backup database or an exported csv as well.

vanlummelhuizen:
> do I need to add a migration?

@susanodd Yes, I believe changing the choices argument in a model field entails making a new migration. @Woseseltops and I decide to hold of new migrations until the whole FieldChoice revision is merged with master. So, if possible, do not make any migrations.

> A deleted gloss cannot be restored from a DeletedGlossOrMedia object. There is no information about phonology, etc.
> It would need to be restored using a backup database or an exported csv as well.

I believe that DeletedGlossOrMedia is not meant for undoing deletion, but just for having some info about what was deleted. Perhaps @ocrasborn can say something about this?

susanodd:
I added an admin class DeletedGlossOrMediaAdmin to be able to browse things that have been deleted.
[...]

Issue #127, opened 2016-02-18
Show warning when deleting a gloss that is referenced elsewhere


ocrasborn:
places to check for references: 
- In all the tables in the Morphology panel
- In the table in the Relations to Other Signs panel

susanodd:
NOTE TO SELF
Bump up to relevant.
Elsewhere has expanded.

susanodd:
Here's what this looks like for a totally nonsensical sign in dataset tstMH:

![delete-sign-modal](https://github.com/Signbank/Global-signbank/assets/24272544/72f004c9-8e13-47ce-bf64-9343d7fd0f76)

susanodd:
And again with the interface language NL:

![delete-sign-modal-NL](https://github.com/Signbank/Global-signbank/assets/24272544/674323bd-91e3-49de-8577-c2f1154d98c9)

susanodd:
Here are the latest screenshots.
This shows the setting GUARDED_GLOSS_DELETE turned ON (delete does not happen if the gloss is related to other glosses. (The default is OFF so the gloss is deleted. In that case the user is redirected to the Gloss List View.)

![verwijder-stekker-B](https://github.com/Signbank/Global-signbank/assets/24272544/511c6b11-b30e-47c7-9eee-168cc354f380)

![verwijder-stekker-guarded-gloss-delete-message](https://github.com/Signbank/Global-signbank/assets/24272544/2d229cf6-ec4d-43ee-9729-9a1b0ae22a90)

susanodd:
@ocrasborn can you have a look and see if this looks all right?

susanodd:
This is live now.

susanodd:
I did the same for the Morpheme Delete button

susanodd:
Show this information to the new API command (#1210 )

ocrasborn:
Looks good @susanodd !

susanodd:
Glosses are archived now.

PR #1083, opened 2023-11-13
Show warning on delete sign in gloss edit view if references other glosses

susanodd:
Information has been added to the modal for Delete this sign.

At the moment, I commented out the actual gloss.delete() in update.py
Do we want to prevent the user from deleting the gloss if it's related to other objects?

### DIFF
--- a/signbank/dictionary/admin.py
+++ b/signbank/dictionary/admin.py
@@ -691,6 +691,19 @@ class GlossRevisionAdmin(VersionAdmin):
         return self.list_display_links
 
 
+class DeletedGlossOrMediaAdmin(admin.ModelAdmin):
+    model = DeletedGlossOrMedia
+
+    list_display = ['item_type', 'idgloss', 'annotation_idgloss', 'old_pk', 'filename', 'deletion_date']
+    readonly_fields = ['item_type', 'idgloss', 'annotation_idgloss', 'old_pk', 'filename', 'deletion_date']
+
+    def has_add_permission(self, request):
+        return False
+
+    def has_delete_permission(self, request, obj=None):
+        return False
+
+
 class RegistrationProfileAdmin(admin.ModelAdmin):
     list_display = ('__str__', 'activation_key_expired', )
     search_fields = ('user__username', 'user__first_name', )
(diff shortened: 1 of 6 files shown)


susanodd:
I'm still working on this one. Trying to condense the code now.

susanodd:
This one is finished now. The delete is turned back on. If you try it on real data that has no related objects, it will actually delete the gloss. (I did this on my local database and deleted something I didn't want to delete. Then typed it all in from the one on Global. Really annoying. But it works.)

susanodd:
You may want to try this out on your local server to see how it works.
Only the tstMH dataset has sufficient (nonsense) relations between glosses and morphology filled in to test it.

susanodd:
This may need to be synced with #659.
(Although totally different issue. I'm afraid to delete signs, so I haven't investigated what happens when they actually are deleted.
This merge request merely does what issue #127 wishes.

susanodd:
Now that we have the query parameters implemented, it would be possible to store the phonology in the DeletedGlossOrMedia info. (It's really annoying if you delete a gloss you didn't mean to.)

susanodd:
I added screenshots to the issue #127.

susanodd:
I added an Admin class for DeletedGlossOrMediaAdmin so it's possible to browse them in the admin.
(No add or delete permissions.)

vanlummelhuizen:
> At the moment, I commented out the actual gloss.delete() in update.py
> Do we want to prevent the user from deleting the gloss if it's related to other objects?

#127 only talks about warning before deleting, not preventing deletion. So, I think you should uncomment the  `gloss.delete()` line.

susanodd:
> > At the moment, I commented out the actual gloss.delete() in update.py
> > Do we want to prevent the user from deleting the gloss if it's related to other objects?
> 
> #127 only talks about warning before deleting, not preventing deletion. So, I think you should uncomment the `gloss.delete()` line.

It's possible to delete now.
There is an if that checks though.

Do we need to physically remove the things it's related to? (For relations I think this is the case because other code explicitly removes things.)

(I don't know.)

vanlummelhuizen:
My rule of thumb: If something is removed, the relations it had are also removed. But the things it was related to, are NOT removed.

susanodd:
Okay. I'll put the button back, but as a danger button.
And I'll add a settings GUARDED_GLOSS_DELETE set to False as default.

susanodd:
This one is ready. I revised the layout and text of the delete modal. Screenshots in the issue.

PR #1088, opened 2023-11-15
Modified Minimal Pairs and Lemma Search forms

susanodd:
as for other search forms. Modified set up of language fields to work if some fields not included. Added missing re.escape to escape of regex input.

### DIFF
--- a/signbank/dictionary/adminviews.py
+++ b/signbank/dictionary/adminviews.py
@@ -6611,6 +6611,11 @@ class LemmaListView(ListView):
         dataset_languages = get_dataset_languages(selected_datasets)
         context['dataset_languages'] = dataset_languages
 
+        # use these to fill the form fields of a just done query
+        populate_keys, populate_fields = search_fields_from_get(self.search_form, self.request.GET)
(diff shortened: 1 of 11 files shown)


susanodd:
This does the same changes as for the other search forms, make a self parameter of the search form, initialize it and its language fields using shared functions. Similarly the templates have been modified to use a filter that extracts the field from the form.

The "escape(val)" has been replaced by "re.escape(val)" a couple of places. (This was related to the language fields in the forms so it got done here prior to revising the forms so they can share code all the places that need the escape.)

susanodd:
I revised the Lemma Reset and Populate code. This all works correctly now. There was an issue with the javascript because it was at the end of the html file and there were two javascript blocks. So the ready code was not accessing the javascript variables that had been defined. These are all in one block now.

I haven't figured out how to condense the get_queryset code of Lemma List View. There is also an annotated query function. Still puzzling on that how to make use of the shared functions in query_parameters.

The Minimal Pairs List still needs revision of get_queryset to use the shared functions.

PR #1090, opened 2023-11-21
Streamline code

susanodd:
At the moment the changes are primarily removing unused, unreachable code and adding more usage information to the shared functions in `query_parameters.py`


susanodd:
TO DO:

- Reduce redundancy of context variable code in other views.
- Reduce redundancy of search form processing in other views.

Issue #1071, opened 2023-10-11
Streamline templates/views that share context variables


susanodd:
Goal Reduce Redundancy in:

- All views and templates
- CSV routines
- Display routines
- GlossListView
- SenseListView
- MorphemeListView
- GlossDetailView

- admin_senses_list.html
- admin_gloss_list.html

vanlummelhuizen:
The views mentioned above share a lot of code. The same goes for the templates. There are probably more places with a lot of overlapping code.

Moreover, the methods in the views are very, very big. 

By taking pieces of code out of those views and templates and placing them in a separate function or method of a parent class, we end up with smaller units of code that is shared instead of copied. This makes the code easier to understand and maintain.

susanodd:
I'm starting out by making separate get_context_data_for_ ... functions in a new python file.

There is an extreme amount of duplication in the GlossListView context function!! And the various variables get set multiple times in the code. This makes it impossible to see any dependencies.

susanodd:
@vanlummelhuizen I would prefer to merge the modified code as I progress.
Otherwise there is a chance that others will make modifications to the old code and make it difficult to merge.

vanlummelhuizen:
> I would prefer to merge the modified code as I progress.

You probably need to coordinate with @Jetske and her work in other branches.

By the  way, @Woseseltops and I are not working next week because of herfstvakantie.

susanodd:
This is an example of what I was trying to say, this setting
SHOW_DATASET_INTERFACE_OPTIONS

In the context variables, it checks to make sure it exists as a setting, otherwise it sets it to False.
https://github.com/Signbank/Global-signbank/blob/682e52dea2d22d374be3f91f6b4421f711c5b579/signbank/dictionary/adminviews.py#L382-L385

However, this could just be replaced by settings.SHOW_DATASET_INTERFACE_OPTIONS in the template and not use a context variable.

For example, as is done here:

https://github.com/Signbank/Global-signbank/blob/682e52dea2d22d374be3f91f6b4421f711c5b579/signbank/dictionary/templates/dictionary/gloss_detail.html#L970

susanodd:
There are extensive lines of code to set up the glosssearchform for the template. Part of this is due to all the multi select field choice fields and the use of colours.
These definitely can go in a separate function. There is also numerous code for setting up all the phonology fields since some need three columns.
Perhaps it’s possible to have a separate template for the different search panels.
[...]

PR #1094, opened 2023-11-23
Streamline continues, bump Python, Django, requirements

susanodd:
Simplified self variable assignment GlossListView, SenseListView, MorphemeListView, HandshapeListView get_queryset; show_all now proper self variable.


susanodd:
This branch passes all the tests!!

susanodd:
I repaired the Sense Search Form. With the new Python and Django, it seems the form submission is more strict in whether the form has been submitted. (It use to work just clicking on the input but now it needs to be a button with an actual value.)

susanodd:
The code works with Python 3.12 and Django 4.2.7.

The tests all are successful.

I'm going through by hand and testing all the views.

I think bootstrap will need to be updated soon. PyCharm has trouble with some of the package code.
(PyCharm runs checks on the code and also highlights stuff that is or will be a problem.)

susanodd:
> @susanodd , great initiative. Can you make a (bullet point) list of what you want the reviewers to pay attention to?
> 
> I notice this changes the Python version. Note to self, or whoever ends up deploying this: don't manually change the python version in the production environment, but instead:
> 
> * Verify the ansible in signbank and signbank-dev are in sync
> * Change the python version in dev, test whether it works
> * If so, update the ansible playbook in the production container and run it

Yes, I'll do that. At the moment, I'm still going through all the templates to make sure everything works.

susanodd:
Changes related to bumps:

1. Regular expression Python literals have "`r'`" at the start
2. CSV export uses a button in the template form rather than an input field (jQuery setup added)
3. `render_to_response` functions parameters changed
4. `render_to_csv_response` functions no longer have a context parameter
5. `render_to_csv_response` uses `StreamingHttpResponse`
6. Python code that generates a row has been moved to helper functions in` csv_interface.py`

Bullets 2-6 are still in progress. The templates with `render_to_csv_response` that still takes a context are gradually being updated.

susanodd:
Note: because the code is being upgraded per template, the code is gradually being debugged for Python and Django bumps.
The changes are all in the template interfaces.
The tests already work.

susanodd:
I'm working on the tests now. After all the above commits, two tests failed.
This one has been repaired:
signbank.dictionary.tests.ImportExportTests.test_Export_csv
(The streaming response was not working in the tests.)

One test fails now:
signbank.dictionary.tests.LemmaTests.test_QueryLemmasWithoutGlosses

There might be a logic error with the toggling of "lemmas without glosses" and "lemmas with glosses".
(I may have made them over lap. The test code is very difficult to read!! There is a huge number of cases for this.)

susanodd:
All the tests pass again.

The code is ready for review. See the list above with specific things.

In general, the Streamline issue is to improve and reduce the amount of code.
That will be continued. But first, Python and Django needed bumps.
Specific complaints about lengthy code in code that is not part of the bumps will be addressed later.

Jetske:
@susanodd I'll review it probably on Monday

susanodd:
@Woseseltops some code improvement (repetition of code, excessively large methods or functions, if-expression simplification, etc.) is being done in issue #1071 (which originated this branch).

I created a spin-off branch `streamline_plus` that follows this branch. (Because I want to keep using Python 3.12 and Django 4.2.7.) 

I just simplified the feedback for regular expression errors to the user.  A separate function for the repeated code that generates the error messages. (I expect you would comment on this. It's in the new branch.)

So if you have comments related to code improvement (undoubtedly), those will be repaired in the `streamline_plus` branch.
This pull request is for the Python and Django bumps. :)

Woseseltops:
> Is it possible to make a new container for this so everybody can try it out, as was done for the previous upgrade?

Would it be okay to use the `signbank-dev` container for this?

> So if you have comments related to code improvement (undoubtedly), those will be repaired in the streamline_plus branch.
This pull request is for the Python and Django bumps. :)

Expectation management: next week will be quite packed, and after that I'm 3 weeks off... so not sure if you want to wait until my review? I trust the other devs :)

susanodd:
> > Is it possible to make a new container for this so everybody can try it out, as was done for the previous upgrade?
> 
> Would it be okay to use the `signbank-dev` container for this?
> 
> > So if you have comments related to code improvement (undoubtedly), those will be repaired in the streamline_plus branch.
> > This pull request is for the Python and Django bumps. :)
> 
> Expectation management: next week will be quite packed, and after that I'm 3 weeks off... so not sure if you want to wait until my review? I trust the other devs :)

Okay, sure. I'd rather have it done sooner, since I spun a new branch from this one, to keep using the bumps.

signbank/dictionary/tests.py
```diff
@@ -615,17 +615,12 @@ def test_Export_csv(self):
...
 
-        # set the language for glosssearch field name
-        default_language = self.test_dataset.default_language
-        gloss_search_field_prefix = "glosssearch_"
-        glosssearch_field_name = gloss_search_field_prefix + default_language.language_code_2char
-
-        response = client.get('/signs/search/', {"search_type": "sign", glosssearch_field_name : "wesseltest6", "format": "CSV"})
-
+        response = client.get('/signs/search/',
+                              {'format': 'CSV'}, follow=True)
```
Jetske, line 619: When entering a query and pressing enter, it directly downloads the dictionary CSV instead of showing the signs for that query

signbank/dictionary/templates/dictionary/admin_lemma_list.html
```diff
@@ -82,28 +82,7 @@
...
-                    </tr>
-                    <tr>
-                        <td align='right'>.*</td><td>&nbsp;&nbsp;</td><td align='left'>{% trans "find all strings in this field" %}</td>
-                    </tr>
-                    <tr>
-                        <td align='right'>(</td><td>&nbsp;&nbsp;</td><td align='left'>{% trans "use backslash to find <br/>special characters in field" %}</td>
-                    </tr>
-                  </table>
-                </span>
+				{% include "tooltip.html" with include_tags=True %}
```
Jetske, line 114: The little tooltip window is not visible for me anymore

Jetske (changes requested):
I ran it on the new python, works mostly! 
See my comments, and also: the search bar doesn't work, it does not find anything. I'm not sure where that would be in the code but could that have to do with show_all or changes in the regular expressions?

signbank/dictionary/templates/dictionary/admin_lemma_list.html
susanodd: Okay, that's really weird.

susanodd:
I'll check the search bar. In a different issue the two search fields are supposed to clear the other if one starts typing in the other. Is it for all pages that it doesn't work?

I removed a submit button that didn't seem to be connected to anything.
Does it work if you use the "search glyph" next to the fields? (Rather than return.)

I think the submit button was for the user hitting return anywhere.

Jetske:
Indeed the search field is cleared if you click the other one, but the search bar also doesn't work if you click the magnifying glass (=search glyph?). I don't think it's cleared because the search get request looks ok. 
It works fine if you then enter a query below in the form and click 'search sign'.

susanodd:
The search bar itself is in the menu.html template that all pages share.
The javascript functions that the focus/blur/click call are in baselayout.html
The two fields always go to the GlossListView template / query.
It's possible that those functions need something different for different browsers?
It works on Safari and Firefox.
Or it's that "return submit" is needed in addition to the search glyph.

susanodd:
The search isn't getting to the python code and the get parameters from the form are empty.
It's not working on Safari either.
There is another field `mirror_bar_search` that is used to pass the menu.html form to the gloss_list_view.html search form.
Maybe something changed in templates that share variables. Or it got messed up with the functions in the menu.html. (Maybe there's a function with the same name somewhere.)

susanodd:
@Jetske I can't figure out where it is going wrong with the menu bar search. The fields don't seem to be connected to anything anymore.

susanodd:
So there is a mysterious context variable `{{glossQuery}}` in the menu bar that is the value of the `'search'` field.

I had to browse the entire history of `menu.html `and look at some old code to search for where this is defined.

It's in` views.py` inside of a function that did not seem to be used anywhere which has hard-coded paths inside it.
But helas, that is where glossQuery is. (It exists in code from 5 years ago.)

susanodd:
@Jetske The menu bar should be working now!!

susanodd:
@Jetske can you try reloading the Lemma List a few times? The tooltip is always visible for me.

Btw, I modified the menus in the admin Pages pull-down under Signs.

`/lemmas/show_all/ `               Show All Lemmas
`/dictionary/lemma/ `                Search Lemmas

I doubt that could be causing problems. But they both use the same template.

Issue #1065, opened 2023-10-04
Make a replaced media file load in the browser immediately after uploading


vanlummelhuizen:
When a media file is replaced, the browser will often not reload the media and retrieve the old version from the browsers cache. We should find a way to make the browser immediately download the newly uploaded media.

One way would be to uniquely name each media file instead of naming it with the text and/or ID of the object it belongs to. When the link to the new media file is used in a web page, the browser will notice it is not yet in the cache and download it.

Note that this is why Django's default method of handling uploaded file is adding a random string to the filename to make it unique.

Jetske:
It seems like the commit above fixes the issue after uploading a new media file. 

However, after deleting a glossvideo it should also work but for some reason it doesn't always retrieve the previous .bak version and still shows the old (/ actually newest but deleted) video from cache.

Jetske:
Fixed: the timestamp wasn't actually added to the url of the media file
Should be working now!

susanodd:
This doesn't work on Safari. It ends up not finding any video instead of the one just uploaded.

Jetske:
@susanodd is this still a problem on Safari or can this issue be closed?

PR #1103, opened 2023-12-12
Load videos not from cache #1065

Jetske:
Time field added as version it tries to find for the video, such that it is never loaded from cache.


susanodd:
@Jetske I'll deploy that tomorrow.

Issue #1119, opened 2024-01-11
Transfer Oefen dataset to NGT dataset?


uklomp:
Would it be possible to put all our entries from the Oefen dataset (all the lemma's, glosses, video's and additional descriptive information) into the NGT dataset? 
I dont think that you will encounter the problem that some of these entries already exist in the NGT dataset, but if you do, the original entry in the NGT dataset can stay (and the Oefen one can be skipped), or the NGT one can be overwritten, both is fine.

susanodd:
i will look into how best to do this. This needs to be done with a command rather than export/import.

The IDs of the glosses and lemmas are already in use (by themselves), so it should be possible to just change the dataset and keep the same IDs. It would be best to move them rather than copy them otherwise they don't keep the same IDs.

The media files would also need to be moved by the command. But there are links to these files inside of the GlossVideo class. (It keeps track of video operations and makes backups.)

susanodd:
Would you need commands over datasets? (Say like queries to find intersections or exclusion over two datasets?)

I'm trying to figure out what you would want to do with this command if it needs to be in the interface rather than a straightforward command.

We have Copy at the gloss level and you can copy the gloss from one dataset to another (Both must be selected and you need permissions in both.) But this command only copies Phonology, not senses or relations.

susanodd:
@uklomp If you just want to create the same signs in dataset NGT, then this can be done via CSV.
The signs will all get new Signbank IDs.
The creation creates new lemmas at the same time as new glosses.
For creation, just the Dataset, and the Lemma and Annotation columns.

The after creating new signs via CSV in NGT, you need to export those to obtain the new Signbank ID in the new dataset. Then you can update the new signs.
From your original export of those signs from Oefen, you can copy the columns you want to update (the other fields).

As you can guess, this would be error-prone, since you need to use the IDs of the new glosses for the right signs. I guess if you sort your original file by the ID, then it will create the new glosses in the order they appear.
So you could sort the "update" of those glosses the same way.

Once you have created the new signs in NGT (from your signs that are in Oefen), you can search the NGT glosses on Created By and Creation Date to limit the results to those new signs. Then export the new signs to CSV.

They should be sorted in the same order as the original signs, so you could copy the phonology and senses columns to the new file to make it ready for import.

susanodd:
@uklomp Are you able to do the above yourself, or would you rather have me do it?
It needs two steps. Two CSV files for import into NGT.
[...]

PR #1120, opened 2024-01-15
Move button to switch dataset of gloss

susanodd:
A button has been added to GlossDetailView to allow moving a gloss to a different dataset. #1119 

This involved updating the lemma field dataset to the new dataset.
To maintain consistency, the dataset of the related glosses (via relations or morphology) are also changed.
To do the morphology, reverse morphology is also retrieved in related objects.
For example, gloss Ijsbeer = Ijs + beer.
If any of these glosses is moved, the others also need to be moved.
This is done in an atomic transaction.

The videos and images of the gloss(es) are also moved to the new dataset.
This is done using a signal. (As for other updates.)

At the moment, there are checks that the user is a superuser in order to do this.
Both datasets must be selected.

The existence of the gloss already in the target dataset is checked.

- An error feedback is shown

If the gloss can't be moved because of a similar gloss in the target dataset:
- Expect the user to change the lemma translations and gloss annotations to remove ambiguity.
- The target dataset must have the same translation languages

Relations can be transitive. Once some glosses that are related are moved, this will also need to move more glosses.

- Assumes that the user will look before jumping.

That the annotation and lemma texts non-overlap are soft constraints on the dataset (This is checked manually by all the code.) The use of one text per language is a Django constraint.

```
      AnnotationIdglossTranslation
                 unique_together = (("gloss", "language"),)

      LemmaIdglossTranslation
                unique_together = (("lemma", "language"),)  # For each combination of lemma and language there is just one text.
```

### DIFF
--- a/signbank/dictionary/adminviews.py
+++ b/signbank/dictionary/adminviews.py
@@ -55,13 +55,18 @@ from signbank.frequency import import_corpus_speakers, configure_corpus_document
     dictionary_speakers_to_documents, document_has_been_updated, document_to_number_of_glosses, \
     document_to_glosses, get_corpus_speakers, remove_document_from_corpus, document_identifiers_from_paths, \
     eaf_file_from_paths, documents_paths_dictionary
-from signbank.dictionary.frequency_display import collect_speaker_age_data, collect_variants_data, collect_variants_age_range_data, \
-                                                    collect_variants_age_sex_raw_percentage
+from signbank.dictionary.frequency_display import (collect_speaker_age_data, collect_variants_data,
+                                                   collect_variants_age_range_data,
+                                                   collect_variants_age_sex_raw_percentage)
 from signbank.dictionary.senses_display import (senses_per_language, senses_per_language_list,
                                                 sensetranslations_per_language_dict,
-                                                senses_translations_per_language_list, senses_sentences_per_language_list)
-from signbank.dictionary.context_data import get_context_data_for_list_view, get_context_data_for_gloss_search_form, get_web_search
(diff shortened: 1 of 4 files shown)


Jetske:
If the gloss and/or its related glosses do happen to already exist, are they overwritten? I think this should either not be allowed, or show a warning with exactly which glosses will be replaced. 

Would it be safer to copy them rather than moving?

susanodd:
I revised to code to give feedback instead of moving the gloss if the translation languages of the target dataset are different, or if the lemma or annotation texts already exist in the dataset.

I did not do checks on the related glosses. They are visible in the modal so the user can see what the related glosses are.

The issue is for use to move the Oefen glosses to NGT. So I am assuming that we will be doing this.
I checked by hand to see how many glosses have related glosses. At the moment, it's in the morphology.

I doubt anything bad will happen. I tried it out on SIBILING-C to see if the warnings work okay for trying to move it to NGT.

I'll see what I can do about the related glosses. (Basically, the only "solution" would be to add -A -B -C to the glosses in order to move them. So not do anything if there will be ambiguity.)

susanodd:
I'll try out what happens if the morphology related glosses are already in the target dataset.
I guess it could give feedback on those glosses and not do anything.

The other media is automatically in the right place because it's stored under a folder gloss id. It doesn't need to be moved.

susanodd:
It's going to be at least Thursday before the code can be used on Global.
I didn't get the ansible scripts working all the way. So the master can't be deployed on signbank yet.

susanodd:
I guess I'll add the bullet explainations above to the modal.

susanodd:
@Jetske I modified the description of the command in the model. (See screenshots #1119)

There are also screenshots of the relations view in the target dataset with the images moved.

PR #1127, opened 2024-01-19
Convert webcam/matroska/quicktime videos to mp4.

susanodd:


susanodd:
The popup error on Mac during upload is:

`magick kan niet worden geopend omdat de ontwikkelaar niet kan worden geverifieerd.
`

Is this something the drag and drop code is using?

susanodd:
@Jetske I'm testing uploading videos on Mac Safari.

I commented out a lot of code to see what the code is doing.

By the time the code gets here:

https://github.com/Signbank/Global-signbank/blob/9d7b48ca4cede2be2fe6a43b8e073c14329b6591/signbank/dictionary/models.py#L651

(With the rest of the operations on the video file commented out....)

The original video (.mov) has been named (.mp4) and the quality is considerably degraded.

But I don't see this in the code.

It's probably inside the save on a GlossVideo.
Whatever that code is, is broken.

susanodd:
I encountered some except prints during uploads. This causes it to fail but nothing is shown to the user. (Because they are caught elsewhere, it's not clear there was an error.)

This is related to the reversion routines.

susanodd:
This code is wrong:

https://github.com/Signbank/Global-signbank/blob/9d7b48ca4cede2be2fe6a43b8e073c14329b6591/signbank/video/convertvideo.py#L125-L135

The `.mov` file is already in `h264` but the probe returns a longer string which contains `h264`:
`h264 (Main) (avc1 / 0x31637661)`
and does not match `format == h264` so the file is converted (to h264 again).

susanodd:
I fixed the code just above.

susanodd:
This probably needs to be a draft branch. I'm finding it difficult to debug all the video code to get it working for mov files.

I think the resize needs to be rewritten. The aspect ration seems to cause problems. (The CNGT script function fails on something giving missing uprobe not found.)

The probe function was actually getting wrong information from the mov file. (Or rather, it was more information than the code was testing for.) I suspect something like this is also going wrong for resize.
It looks from the GlossListView that the little videos have aspect ratio 3:4. But the quicktime macbook camera videos are wider. It looks like ffmpeg needs different parameters for different formatted videos.

susanodd:
I have the code working for quicktime videos now. I think what was going wrong before the repairs was that

1. quicktime is already in h264 format - to fix this the convert_video routine ffmpeg, if the source video is quicktime, it now uses codec "copy" to remove the container from the video and save it as mp4
2. the frames per second is 15, not 30. This was causing the make small video to not work. Instead - new python function - first small png files for the frames are generated, then these are put into a new .mov file using rawvideo format. Then this is converted to mp4.

PR #1173, opened 2024-02-21
Dataset Media Managment View and Api interface 

susanodd:
A new view is added for Dataset Managers to import zip files and see what is available in them.

The API interface now provides functionality for a user with appropriate permissions to upload a zipped videos archive and unzip it into the import_videos folder.
The API also provides functionality to actually import those videos into the named glosses.

This can be used on signbank-dev.


Woseseltops:
I notice you said this is still a draft, so I clicked the button 'convert to draft' :)

susanodd:
This will require a lot of testing on signbank-dev.

susanodd:
This is working now!
The API commands successfully import the videos to the proper location.

susanodd:
CORS is turned on now on `signbank-dev`

Woseseltops (commented):
* Reading through `api_interface.py`, I notice lots of smaller individual functions. I think a bit more structure here wouldn't hurt. How exactly doesn't really matter: perhaps this file could instead be a module with several submodules, but I've also done it in the past by making them classes, all with 3-5 methods. 

* Very nice usage of `StreamingHttpResponse` !

* I feel like `VideoGloss` doesn't really capture what's the class is doing. Not sure, but perhaps something like `VideoImporter` would be more descriptive?

susanodd:
@Woseseltops I agree. There are an awful lot of small functions. I will try to reduce this. Some of the underlying computations are done in slightly different but similar functions.

I added some more functionality today for the template view.

I'll try to strip down the basic calculations and build up that way. (Probably because I was inventing what it was doing while coding it.)

susanodd:
I split the functionality for the API zip/import routines from that for the Dataset Manage Media template.
I moved the shared zip specific functions to a new file.

susanodd:
This is on signbank-dev now.

The Dataset Media template works properly after much revision using a real apache server instead of the PyCharm runserver. (!!!!!)

BUT the API for importing a zip archive DOES NOT WORK from my own computer, because my computer has no IP. (Actually it does, but it has no server to communicate with Signbank.)

(Thanks, CNCZ. We can't debug our code properly.)

susanodd:
Would this work to have a zip file on ponyland? Or does that not serve files either? I'm a bit confused because I can copy files from my desktop Ubuntu computer using sftp and copy between applejack and signbank using scp.

susanodd:
@Woseseltops since you've used it now, do you have an opinion about the zip files?

Should they stay on the server (with functionality to open/close the contents list, and to delete them.)
Or should they just be removed?

If you try uploading the same zip file, it works again. (I've been doing this locally for the past week for debugging -- but forgot the zip archive structure ---) You can try it again and it just replaces the videos.

I'm wondering about the output you said it didn't report back. (I can modify this, if needed. Probably the contents of the zip file also needs to check that the gloss exists. It looks confusing that there are green check marks, but at least one was a lemma, not an annotation, as mentioned a few comments back.)

susanodd:
@Woseseltops do we not already have the rest framework installed?
The API token  is inside of that.
I though we already had this because @vanlummelhuizen used to talk about it years ago.
Or is that also a programming style? (Make things restful.)

susanodd:
@Woseseltops on dev, using the upgraded Has Video and the upgraded display of the Zip Archive contents, it looks like fewer videos were imported than are in the zip file.
(Like 100 fewer.)

I shall try to figure out why that is so. Maybe the server gets too many requests?
@rem0g suggested earlier to import the videos in batches.
It looks like the entire zip file is uploaded and expanded okay.
I'll try to see how it goes with using the url to import them instead of the Dataset Media View.

The ajax calls inside the template javascript are set up inside a function. There are data arguments on the okay files in the imported videos (middle section, waiting to be imported). Those are passed to the javascript function in a loop.

I suspect the server can't handle all the requests. But I don't know yet how to resolve this.

susanodd:
> Hey @susanodd , great work with the wiki page. I've restructed a few things to make it more readable, and during this process, I noticed a few missing bits of info, which I've marked with `todo`... do you know what should be in these places?
> 
> Furthermore, some thoughts:
> 
> * The naming of the endpoints seems to be inconsistent: `get_gloss_data` and `get_field_data` has a verb, but `package` and `info` do not. I personally feel the option without verb is descriptive enough :)
> * For the section 'updating_data', the standard in the world of APIs is to use `POST` instead of `GET`, to indicate that you're not reading but actually changing things. You could also consider `PUT` when creating new things. Using something other than `GET` has the added advantage that you can send a file along as an attachment, so you don't have to upload it to another server first ;) . See https://stackoverflow.com/questions/3938569/how-do-i-upload-a-file-with-metadata-using-a-rest-web-service
> 
> Let me know what you think!


I'm replying to this here.

Regarding uploading videos in the API.
I purposely made it so it's a zip file first, then that gets unzipped first in the `TEMP` location, then into the `import_videos` folder, then into the glosses.
Because, this conflicts with the videos model and the version management.
Because , the real gloss videos are in `protected_media` which is not visible as a folder. Django protects this.

susanodd:
@Woseseltops I created the api url, plus a way to test it in PyCharm since PyCharm runserver does not have CORS.
I made an extra template and an extra url for using it within signbank.
This is on branch csv_enhancement, because branch api_interface is already 1449 lines of new code to review.
This issue should be merged into master and put on signbank before the rest can be put on dev.

susanodd:
@Woseseltops here is some info about the signbank-dev server, when entering the venv:
```

your server socket listen backlog is limited to 100 connections
your mercy for graceful operations on workers is 60 seconds
```

susanodd:
@Woseseltops on signbank-dev, when importing from the very large zip file, in the developers window, it has a lot of messages that it can't find protected_media. It looks like the paths are incomplete for some reason.

susanodd:
@Jetske and @susanodd found the error that was causing the new videos not to be shown.

They were in fact being uploaded, but if the path to the video of the gloss differs to what is stored in GlossVideo, it ends up not finding it, even if the stored path does not exist.
Now the GlossVideo `videofile.name` is changed to that of the newly imported video.

Plus, some paths weren't being expanded correctly because of `protected_media` being expanded to something totally else.

susanodd:
1. I made it so the videos in the import_videos folder are removed after they are imported to glosses.
2. The import button has been changed to "next 10" instead of all of them, so as to not bog down the server. The user sees the feedback per gloss quicker. Because of how I designed the interface, the filename to be imported is stored in the listing of import_videos (to be imported videos), then the actual import ajax call uses this, then puts the success message in thst same row.)
3. The videos of @ocrasborn were all uploaded, but the "GlossVideo" had references to "lemma" (the pre-dataset "idgloss" field that used to be a gloss field before class Lemma and Lemma Translation were added). Because of this, the video display and video search code was still using a wrong video path, not the path to the new video. So I modified GlossVideo to have the correct filename (the Annotation text).
4. So in fact, the GlossVideo path is actually some kind of path, not just a string representation.

Issue #1187, opened 2024-03-14
Implement authentication token for the API


Woseseltops:
Related to #1152 

Discussion with @susanodd today. We decided on these things:
* One user will be able to generate multiple tokens
* For each token, they can specify the scope, which can be one or multiple datasets
* We want the tokens to expire after a while
* There should be a button in the profile. Once you generate it, it should say 'copy paste, you will never see this token again'
* There should a maximum of tokens the user can generate (maybe a setting)

Woseseltops:
@susanodd, to make a view accessible as API endpoint:
* Use the Django decorator `@csrf_exempt`
* Get the token out of the request header

susanodd:
@Woseseltops I'm having trouble creating the tokens. I need an appropriate example. 99 percent of all the web documentation turns out to be the Rest framework. And the searching gets worse when you add "without rest framework" or "not using rest framework". I tried adding Python 3.12 to it, since it's not supported by rest. I've been searching for 2 hours now.

susanodd:
@Woseseltops this is separate from the previous paragraph.
I won't migrate the database using the auth_token.
Does this not work with Python 3.12 either?

`django.db.utils.DatabaseError: database disk image is malformed`

susanodd:
@Woseseltops (read the above first). How to solve this?
Can we put the IP address of UvA in allowed servers in the apache settings?

The alternative is to implement the tokens from scratch.

susanodd:
[OLD SCHOOL]

The token will be created by hand using old methods of creating keys for this.

I'm working on the session variable for dark mode (#1166).

I'm wondering:

When we create such a token for the API, isn't the actual "token" going to be visible in the calls to signbank from outside of signbank? (Session communication with the server.)
What's to stop spies from just listening in and grabbing the token?

This is what I was trying to ask on Friday @Woseseltops : is the "streaming" done on the (Rest) tokens done in order to make  the communicated tokens not be "visible" in their entirety to aliens trying to monitor communication to Signbank?

(Or do we need to encrypt them as well? Like for passwords.)

(This is not a nonsense question, the Google analytics records the communication, see the old issue about website statistics.)

susanodd:
I made a first go at this, by creating a class for the tokens and accessing them in the User Profile.

But I am not the right person to code the creation and employment of these keys in the API communication.

@Woseseltops according to @Woseseltops has lots of experience using API tokens.

Woseseltops:
[...]

PR #1197, opened 2024-03-21
User profile to include affiliation tags and api token

susanodd:
Branch `api_interface` is running on `signbank-dev`.

Affiliation tags have been added so as to create Affiliations and assign them to Users and Glosses.

Ultimately, the users may be associated with an Affiliation in the Admin.
This then shows up in their profile.
Glosses can be assigned an affiliation in the Gloss Detail View.
This will also be added to the glosses when they are created, so as to keep track of whether they are RU, UvA, or Corpus NGT. Plus other projects, etc, that was to tag things. These tags can be given a colour in the admin.
(At the moment, affiliation tagging is not done automatically yet, since this is a pull request. @uklomp wants to tag things per UvA user.)

The User profile has also been made ready to include the API tokens as discussed in #1187.

But how to implement this exactly has not be done yet.


Woseseltops (approved):
Great usage of template tags
Great usage of guard clauses

Issue #1212, opened 2024-04-01
JSON Download, Phonology in Dutch?


rem0g:
We would need to have JSON file in Dutch language, since we set phonology in dutch parameters. 

An other option would be to have translation file with bounded translation between the parameters.

susanodd:
Would a parameter to indicate the language suffice? This has to work for all the (current) interface languages.

English has been used (for CSV) because some languages are not filled in and so remain English. Moreover, some languages are right to left. (The browser sometimes goes berserk. Safari adds a "ltr" or "rtl" tag to input fields internally.) The API routines will need to work for all languages once such an option is made available.

The translations of all the "interface text" is in the ".po" files. The field choice choice fields, and handshapes use "model translations".

@vanlummelhuizen is the one who initially set up the model translations.

The ".po" files need to be compiled and are not dynamic. The model translations are dynamic. Thus model translations are used.

susanodd:
@rem0g do you want the language code in the url itself? Or as a parameter to the api called by the url? Or obtained from the request?

susanodd:
Update. It's not going to work to let Django do this automatically. It basically ignores the NL in the request and seems to arbitrarily use or not use it based on control flow. It looks like it needs to be a parameter the the method in the Gloss model that retrieves the data.

susanodd:
This is working now.
I did the new api functions. Not the package zip one. Do you need the zip also?
I didn't do the zip one yet because it may surprise people if it all of a sudden is translating stuff.
The zip package has existed for at least 8 years.

The only weirdness is that it is translating "Video" to (lower case) "video".

susanodd:
This is available on signbank-dev now!

rem0g:
Ok great, now i need instructions updated in Wikipedia to obtain the Dutch JSON. I assume this is for dictionary/pakcage?

susanodd:
My browser did this automatically. Based on what interface language I had set in Signbank.
It gets the LANGUAGE_CODE from the request.

I'll update the Wiki. I'll ask @Woseseltops. He's going to do the API tokens.

rem0g:
Using "dictionary/package/" there is no translation of phonology parameters and it also contains a lot of less information in comparison to english JSON (1MB versus 2 MB) on signbank-dev.

susanodd:
I didn't add it to the package url yet. I shall do that. (Because the package function is very old, I did not know what to do with it qua translation, since they use it at MPI.)

[...]

PR #1213, opened 2024-04-03
#1212: Return gloss fields and field values in Dutch

susanodd:
Or whatever the language code is in the request, if available, otherwise English (default).

This is on signbank-dev.

The API function retrieves the user and the language code from the request.


Woseseltops (approved):
Great work

Issue #1207, opened 2024-03-26
Gloss Create API with Senses as an extended field


rem0g:
Since it is not possible to update Senses via CSV, we would need to create Senses via Create Gloss API immediately.

susanodd:
Yes, this is possible!

See the Notes on the CSV Import Update Gloss pages. This is written there.

There are also some "radio buttons" on that page to turn on and off regarding what to do with the columns.
But as noted, you are not including those columns.

I needed to mention the senses because the import update CSV basically ignores Senses columns if the gloss already has senses. (Nothing will show up in the "Changes to Perform" you are shown before you apply the changes.

susanodd:
You probably need a separate Create Senses for Gloss API.
(The glosses must exist first!)
The senses model also involves all the translation languages for the dataset. Just like the other fields.
Perhaps @Jetske can respond here.

It will still be a problem with keywords with commas in them. (This is ambiguous.)

Jetske:
@rem0g what would the sense input (for both creation and update) look like? eg
```
"Senses: Dutch" : [["trefwoord", "trefwoord, gescheiden door komma's"], ["meer", "trefwoorden"]] 
"Senses: English" : [["keyword", "keyword, separated by commas"], ["more", "keywords"]] 
```
or
```
"Senses": {"Dutch" : [["trefwoord", "trefwoord, gescheiden door komma's"], ["meer", "trefwoorden"]], "English" : [["keyword", "keyword, separated by commas"], ["more", "keywords"]]}
```

rem0g:
It can be done both, it's up to your preference then i can adapt my code to that.

susanodd:
What about when the user omits translations, or the number of things does not match up? Empty lists?
Not offering translations is handled by the Gloss Edit Senses interface. They don't have to match up.
And how to match up the new data to the existing data in order to update, instead of replace.

Jetske:
The senses absolutely have to match up, but that can be done by adding empty arrays in case there is no translation for a sense. It would be good to add a check for whether the sense arrays have the same length in every language. 

Other than that, for updating this would be similar to how update works in the gloss edit senses interface. An update will replace the old senses. But by then the existing senses are already loaded in your system @rem0g, right? So you don't start editing from scratch.

rem0g:
Yes and for the senses english/dutch I can add array.length as verificaion before sending it.

susanodd:
[...]

PR #1215, opened 2024-04-05
Senses added to gloss create api #1207

Jetske:


Jetske:
@susanodd I've tried it, built in some error checks, and it seems to work ok. 
Can you check again though before deploying this?

susanodd:
@Jetske yes, I will check this. Thanks for implementing it.

susanodd:
This works okay. See comments about Wiki and error message feedback in the issue.

Issue #1198, opened 2024-03-22
Pages links at bottom of Lemma Search repeats pages


susanodd:
I have no idea what is causing this bug. The Django paginator seems to think there are thousands of pages when there should only be two pages.

Is this possible? None of the python files for Signbank import the Paginator

From the Django website, instructions:

`from django.core.paginator import Paginator`

A new branch has been created for this issue. It is non-trivial to repair. The queryset is an annotated queryset that includes a count of the matching glosses. This is messing up the Django Paginator.

susanodd:
This has been deployed on signbank.

PR #1218, opened 2024-04-11
#1198: Rewrote queryset code of LemmaListView

susanodd:
Moved annotated query to model method to count number of glosses. Removed extra annotated queryset that could not be paginated correctly. Now it uses object_list and calls the new num_gloss method in the view.

BACKGROUND
There used to be two get_queryset methods in the LemmaListView. One provided an annotation column with the number of glosses for the lemma.
The pagination did not work for this setup and all the lemmas kept appearing on Page 1, and every page thereafter.
To repair this, the "num_gloss" was put in a LemmaIdgloss method instead of being used as an annotation on the queryset.
The two queryset methods were merged in a sense into a single method.

The code is a bit obtuse. The actual pagination links in the template are difficult to read.

This branch is deployed on signbank-dev in order to make reviews more convenient.

### DIFF
--- a/signbank/dictionary/adminviews.py
+++ b/signbank/dictionary/adminviews.py
@@ -1,6 +1,7 @@
 
 from django.views.generic.list import ListView
 from django.views.generic.detail import DetailView
+from django.core.paginator import Paginator
 from django.views.generic.edit import CreateView, DeleteView, UpdateView
 from django.db.models import F, ExpressionWrapper, IntegerField, Count
 from django.db.models import OuterRef, Subquery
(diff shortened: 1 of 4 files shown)


Woseseltops:
Apart from 2 forgotten log statements, the code changes look fine to me. I can't easily verify the bug is fixed though... can you provide a url on prod where we can view the bug in action?

PS
Thanks for the commit naming, now I've got that song in my head for the rest of the day 🎵

susanodd:
> Apart from 2 forgotten log statements, the code changes look fine to me. I can't easily verify the bug is fixed though... can you provide a url on prod where we can view the bug in action?
> 
> PS Thanks for the commit naming, now I've got that song in my head for the rest of the day 🎵

:)

I was a bit out of it after all the etc stuff.

You can see the bug in action on the normal Signbank site. Just go to Search Lemma's for NGT and e.g., search on Lemmas without glosses. There are hundreds of pages, all of them identical.

susanodd:
There is something weird going on with this code in the unit tests.

It looks like the control flow of the test shows that they are deleted, but then the lemmas that were successfully deleted in the tests appear again.

Is this caused by the Django code? It looks like inside the test, this is reverting to before the commands inside the test was performed.

`return HttpResponseRedirect(reverse('dictionary:admin_lemma_list'))`

Issue #1219, opened 2024-04-13
how to change phonology/morphology on SignBank


rosestamp:
[new_ISL_lemma_updates (1).csv](https://github.com/Signbank/Global-signbank/files/14968708/new_ISL_lemma_updates.1.csv)

I am trying to upload this CSV in order to update existing entries on our ISL dataset with new morphological and phonological information.
When I try, using "Import CSV Update Existing Glosses", it states: Attempt to update Lemma translations. Use Import CSV Lemma Update instead.
When I do what it suggests using Lemma Update, it states:
The header row of the csv file looks like this: Handedness, Strong hand, Weak hand, Strong hand letter, Contact type, Location, Movement direction, Movement Shape, Relation between Articulators, Handshape Change, Repeated movement, Alternating movement, 42719, ISL, DUCK, 2s, B, no, , neutral space, to and fro, straight, next to, Yes, yes, 42687, DEER, W, beak2_open_spread, initial, forehead, forwards, arc, No

I do not understand what is wrong with my file. 
Please let me know. Thanks!

susanodd:
@rosestamp the easiest is to remove the columns you do not need to update from the CSV file.
Update only needs the Signbank ID column and the specific columns you are updating.

If you remove the Lemma columns it won't tell you to use the Lemma Update.

(More later.)

susanodd:
IMPLEMENTATION CODE COMMENTS
@Woseseltops any ideas here? This is the first time this comes up.

https://github.com/iamoracle/django_case_insensitive_field

The Tags model (prefab) is case sensitive as well. If you create new tags that differ in case, they are different tags.
In this issue, the case of the field choice values does not match those in the database.
The easy fix is to just convert either side to all upper or all lower. The icontains match may find substrings that match.

susanodd:
Hmmmm. The `iexact` will need to be on a model translation multilingual `name` field for the FieldChoice model.

susanodd:
I revised the code to use iexact.
There are much fewer field choices that don't match.
But this one needs to be changed:

@rosestamp:
`For DUCK (42719), could not find option next to for Relation between Articulators`
(There are about 800 rows with this error.)

There is a field "Next-to" as a choice. It requires the hyphen. (You can use next-to now, without the cap. But the code is not yet live.)

I'll put the revision up asap.

susanodd:
@rosestamp here's another one:

`For CORRECT (42658), could not find option downwards + contralateral for Movement Direction`

[...]

PR #1222, opened 2024-04-16
#1219: Case insensitive CSV input field choices, handshapes, semantic…

susanodd:
… fields,

dialects.

Made field choice value lookup case insensitive for CSV import update.


Woseseltops (approved):
Code looks okay to me... but if you have some time, perhaps you can use this issue to isolate a few more methods in `tools.py` related to this functionality, and put them in a separate file? Like discussed before `tools.py` is becoming a very large pile of unrelated functions ;)

Issue #1202, opened 2024-03-23
Update Gloss ID via API


rem0g:
Now we have Gloss Video Update and Gloss create API, we have to yet have Gloss Update API. 

It should have those fields and linked to Gloss ID. 

{
  "fields": [
    "Lemma ID Gloss: Dutch",
    "Lemma ID Gloss: English",
    "Annotation ID Gloss: Dutch",
    "Annotation ID Gloss: English",
    "Senses: Dutch",
    "Senses: English",
    "Annotation Instructions",
    "Word Class",
    "Handedness",
    "Strong Hand",
    "Weak Hand",
    "Handshape Change",
    "Relation Between Articulators",
    "Location",
    "Relative Orientation: Movement",
    "Relative Orientation: Location",
    "Orientation Change",
    "Contact Type",
    "Movement Shape",
    "Movement Direction",
    "Virtual Object",
    "Phonology Other",
    "Mouth Gesture",
    "Mouthing",
    "Phonetic Variation",
    "Repeated Movement",
    "Alternating Movement",
    "Strong Hand Letter",
    "Strong Hand Number",
    "Weak Hand Letter",
    "Weak Hand Number",
    "Weak Drop",
    "Weak Prop",
    "Semantic Field",
    "Derivation History",
    "Named Entity",
    "Valence",
    "Iconic Image",
    "Concepticon Concept Set",
    "In The Web Dictionary",
    "Is This A Proposed New Sign?",
    "Exclude From Ecv"
  ]
}

susanodd:
@rem0g where are you getting the input for all of these fields?

They need to correspond to actual possible values for the Field Choices.

The basic functionality will be part of #1162 .
It still needs extensive checking on the input fields that they all exist.

Are you planning on using the "output" from the other API as input to this one? (For example, a giant dictionary with lots of empty fields?)

Is it possible to modify your function to call each "changed" field separately? (So split the API into calls for each field as a separate API call instead of a huge dictionary.)

(That is basically how the normal Gloss Detail Edit works, each field update is a separate ajax call. Otherwise the system would start choking because multiple users work at the same time. Originally, the system did not use ajax, say 8 years ago.)

The code is going to be locking the database while the entire command runs because it will need to check all the fields.
It will also generate errors if there is a problem with the input.
That's why the CSV Import is in two steps. First an error checking step. Then the actual updates.

The CSV Import can do the individual rows sequentially. The error checking also checks that rows do not conflict with each other.
[...]

PR #1226, opened 2024-04-22
Update senses field added to update gloss API #1202

Jetske:
@susanodd When given input for field 'senses', all existing ones are removed from the gloss (for every language) and deleted (if not in another gloss) and the new senses are created. So it doesn't check for each translation/keyword separately. 

Input is in the form of 'senses': '{"en": [["s1w1", "s1w2"],["s2w1"]], "nl":[["s1w1"],["s2w1", "s2w1"]]}"

Can you check if you agree with this implementation?

### DIFF
--- a/signbank/abstract_machine.py
+++ b/signbank/abstract_machine.py
@@ -20,7 +20,7 @@ def convert_string_to_list_of_lists(input_string):
     try:
         list_of_lists = ast.literal_eval(input_string)
     except (ValueError, SyntaxError):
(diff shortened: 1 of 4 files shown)


susanodd:
I put it in the test template to check if the field is empty. (The javascript that sets up the call.)
(You can see that in the javascript that sets up the "update" dictionary. It skips empty fields.)
So it could be because you're using a list format the code thinks it has contents.

susanodd:
Okay, I'm on your branch on my office computer.

I tried the update on a gloss with only the Senses filled in with exactly the example you show above.

It got this runtime error:

`sqlite3.IntegrityError: UNIQUE constraint failed: dictionary_translation.gloss_id, dictionary_translation.language_id, dictionary_translation.translation_id, dictionary_translation.orderIndex`

Jetske:
I see, that's actually an error in my example input because it tries to add "s2w1" twice to one sensetranslation. 
Good catch though, I'll make it check for duplicates in each list.

Issue #1178, opened 2024-03-04
Panel "Voorbeeldzinnen"


uklomp:
We would like a panel to show example sentences (as discussed with @Woseseltops and @Jetske). (see also issue #1168 for the location of the panel).

The title of the panel should be "Voorbeeldzinnen" / Example sentences. 
We want to be able to upload a video of a sentence in this panel. The sentence will exemplify the use of a sign. There should also be a possibility to upload a linked ELAN file, because we want to show a gloss-by-gloss translation of the video in the sentence. And there should be Dutch translation underneath as well. Something like this: 

![image](https://github.com/Signbank/Global-signbank/assets/147159596/04d3723f-527c-44ed-b2d7-2af53f3edafa)

Additionally, we would like the gloss in the gloss-by-gloss translation to light up / change colour when this specific sign is shown, something like this: 

![image](https://github.com/Signbank/Global-signbank/assets/147159596/7263707c-cf46-4eb9-98c5-5f5e32b0a630)

Last but not least, it would be good if the sentences are able to appear for every gloss that they include - BUT for some glosses, this will automatically lead to many connected sentences and some of them may be good examples for this specific gloss but others might not. Wessel & Jetske and me therefore thought of a functionality to be able to categorize an example sentence as 'representative yes/no' for a specific gloss. If yes, it will appear in the top three or so. If not, it is shown as a written option in the panel of that specific gloss, but not immediately shown as a video. Or something like that. Up for discussion.

Jetske:
Models:
![IMG_20240305_165851732~2](https://github.com/Signbank/Global-signbank/assets/10157482/415938f5-88b4-4704-b0da-34295dd07355)

Woseseltops:
About the upload panel:
* The connection between videos and the glosses in that video will be made with the eaf file, right? Not sure if it makes sense to have the upload panel on the detail page of an individual gloss then?
* How would setting `isRepresentative` work? Perhaps, after uploading, it shows the end result with the highlighted glosses already, with an extra row/box to uncheck glosses that are not representative?

Jetske:
Discussed with @uklomp :
The input will be a video of a sentence, with two tiers that should be read from the eaf file: 'Glosses' and "Sentence". The glosses are read as they are currently. The sentence is added to a textfield (only Dutch for now) that can be edited during upload. Another textfield is added to fill in 'context'. This is shown underneath the video just before the sentence translation.

susanodd:
[...]

PR #1229, opened 2024-04-24
Annotated sentences panel added #1178

Jetske:
This is ready to test on signbank-dev


Jetske:
@susanodd the settings for temporary files is now added to default.py so that can be deleted from the server specific settings

Woseseltops (commented):
Great work @Jetske. I wasn't aware that there was an `Eaf` object in the `pympi` library, cool. 

Two things to consider:

* I'm a bit confused by `object_type=a`... is there a reason to not just make it more descriptive (`object_type=annotated_sentence`?)

* In the Javascript part, I notice you build some html from strings. This is done in more places within Signbank, but I believe a cleaner way to do something like

```javascript
tableHtml += '<thead><tr><th style="width:200px">Gloss</th><th style="width:200px">Representative</th><th style="width:200px">starttime</th><th style="width:200px">endtime</th></tr></thead>';
```

would be something like

```javascript

    // Create a jQuery object for the table
    var table = $('<table>');

    // Create a jQuery object for the table header section
    var thead = $('<thead>');

    // Create a table row for the header
    var headerRow = $('<tr>');

    // Create and append table header cells to the header row
    var glossHeader = $('<th>').text('Gloss').css('width', '200px');
    var repHeader = $('<th>').text('Representative').css('width', '200px');
    var startTimeHeader = $('<th>').text('starttime').css('width', '200px');
    var endTimeHeader = $('<th>').text('endtime').css('width', '200px');

    // Append each header cell to the header row
    headerRow.append(glossHeader);
    headerRow.append(repHeader);
    headerRow.append(startTimeHeader);
    headerRow.append(endTimeHeader);

    // Append the header row to the table header section
    thead.append(headerRow);

    // Append the table header section to the table
    table.append(thead);
```

It takes way more space though, so I'm not 100% sure. I see this table is built using data from a POST request; another approach could be to built the table server side using the Django template language.

Jetske:
@Woseseltops Agreed :) 

The object-type first had only 2 options and then was extended. There is no good reason not to write those names out in full. 

And I added a template to create the annotations table. To be honest I find it less readable, but it is probably more conform the Django rules to have it separated like that.

Woseseltops:
> To be honest I find it less readable, but it is probably more conform the Django rules to have it separated like that.

For completeness, as long as it's a conscious decision, I'm totally okay with not conforming to Django rules; a comment along the lines of 'adding html as string for readability purposes' would also have been okay. However, now that I see the template approach you just added, I personally do prefer it, so thanks :).

susanodd:
The html needs to be constructed as @Jetske did with jquery, otherwise the nesting structure is incorrect in some browsers. Safari can't parse things with non-structured html.

PR #1230, opened 2024-04-25
Api interface API token

susanodd:
This is on signbank-dev


Woseseltops:
I just tried this functionally, and the authorization works! Awesome B) . I noticed that when I used incorrect credentials, it refused to give me access, but the HTTP status was still 200... is there a way to tell Django to make it 403? 

Furthermore, if you're working on error messages, I got the (html for the) Django error page a few times. Perhaps use a `try except` to just return the error message instead of the whole page? :)

PS
Can you update the example on the wiki to something that is actually allowed? It complains about the senses now. :P

susanodd:
For the test page, it doesn't clear the form after submission, so the error messages kind of stay there.
I will ask @Jetske about the senses.

There are numerous checks on things, so it was trying to be helpful messages.

You want it to return 403 instead of error messages?
At the moment, the error messages are feedback about what is wrong. (Like as it's done for CSV, it doesn't crash, it gives feedback.)

susanodd:
> I just tried this functionally, and the authorization works! Awesome B) . I noticed that when I used incorrect credentials, it refused to give me access, but the HTTP status was still 200... is there a way to tell Django to make it 403?
> 
> Furthermore, if you're working on error messages, I got the (html for the) Django error page a few times. Perhaps use a `try except` to just return the error message instead of the whole page? :)
> 
> PS Can you update the example on the wiki to something that is actually allowed? It complains about the senses now. :P


@jetske can you check the wiki for the senses?

Jetske:
Yes, I added some explanation for the update gloss api to the wiki including senses

Issue #1243, opened 2024-05-22
Signbank API: Allow to update gloss/lemma when there is max one gloss connected


rem0g:
As we want to change glosses because of wrong order in A,B,C,D suffix we want to fix the order via Signbank API, only allow glosses name/lemma name to be changed when there is not more than one gloss connected. If there is more than one, then the update should be done at Signbank website.

susanodd:
Do we want this to ignore case? Or allow to create with different cases?

susanodd:
This is almost done. It ended up needing three different kinds of lookups for the language fields, since they are referred to by different kinds of keys.

susanodd:
@rem0g this is live on signbank-dev now.

The language fields for Annotations and Lemmas are there (using the same keys as for creating a gloss).

Will you try out some of your updates on `signbank-dev` ?
If a lemma is shared it won't work.
There are additional checks if the language fields are already in use.
It will report errors in the returned value as before.

It needs to be tried out some more to make sure all potential constraint violations are caught and not executed.
Kindly see if the new fields work as you want them to.

susanodd:
This is on signbank now.

PR #1249, opened 2024-05-24
API update lemma and annotation fields

susanodd:
This is live on signbank-dev !!!!

This checks for database conflicts before allowing to update.

- already used annotations or lemmas for the dataset and language
- shared lemmas cannot be updated

Also for the API update, modification of strings stored in Gloss Revision History for Senses.

The history has some quirks because the senses updates are done as a whole.
This has been done using a loop over the "before" senses and in turn the "after" senses, using the `__str__` method of the Sense class and a join using '`|||`' to separate. These are both modular, but the strings are stored in the history when saved.

The update to the lemma and annotation fields includes mappings between three different "labels" all available in interface languages.

- the display column human-readable header (used in templates)
- the API retrieval JSON header (used in '`package`', it avoids excessive parentheses)
- the internal label with underscores used in form fields

The two first two are multilingual and mapped by model translations. This allows the API user to provide field choice values for e.g., Dutch phonology. For the language fields for lemma and annotations, (2) has been implemented due to the different syntax for these in the API package that already existed.

### DIFF
--- a/signbank/abstract_machine.py
+++ b/signbank/abstract_machine.py
@@ -82,7 +82,6 @@ def create_gloss_columns_to_value_dict_keys(dataset, language_code):
     required_columns = dict()
     required_columns['Dataset'] = 'dataset'
     dataset_languages = dataset.translation_languages.all()
-    lang_attr_name = 'name_' + DEFAULT_KEYWORDS_LANGUAGE['language_code_2char']
     for language in dataset_languages:
         required_columns[gettext("Annotation ID Gloss") + " (" + language.name + ")"] = (
                 'annotation_id_gloss_' + getattr(language, 'language_code_2char'))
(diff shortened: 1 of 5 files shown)


susanodd:
Some of the code is a bit hard to read and some print statements are still there.

Because this is a high priority issue (#1243) the coding was written with priority on asap instead of style.

Woseseltops:
I'll do the code review part... but about the functionality: as this was requested by @rem0g , would it make sense if he tests on signkbank-dev.cls.ru.nl whether this works as he needs?

rem0g:
Yes i can test that at signbank dev

Woseseltops (commented):
Looks good to me. My only complaint is that it's unclear to me what some of the functions do, like `internal_language_fields`, `revision_history_senses`, `check_value_dict_update_gloss_language_fields` or `gloss_update_fields_check` and `gloss_update_typecheck`. 

Giving some sort of hint of the output already makes a function more descriptive, for example `collect_revision_history_for_senses` or `detect_type_related_problems_for_gloss_update`. Not sure if my proposals actually cover the content, it's just an example of what I feel would be easier to maintain.

susanodd:
Okay, thanks! I will take that approach. (I implemented it as if bailing out water, so I just kept adding code. It's confusing for me to read, too.)

PR #1254, opened 2024-05-30
Api interface Update Morphology endpoint

susanodd:
A new endpoint for updating morphology.
This was decided upon because morphology is all relations rather than fields.
The input values mimic those for CSV update.
Already existing routines from the CSV are called here. Some bugs in that code have been repaired.


Woseseltops (approved):
I tried but I have nothing to complain :P

Issue #1234, opened 2024-05-15
Annotated sentences improvements


Jetske:
The following edits will be made:

- Ability to edit which glosses are representative
- Add that you can download an eaf file and upload an improved one on the edit annotated sentence page
- Underneath the video show the annotation gloss id rather than the lemma (e.g. PT -> PT-1 hand)
- Add the source of the sentence (that is, which corpus)

uklomp:
It should be possible for us to upload multiple sentences under one gloss.

Jetske:
@uklomp Uploading multiple sentences under one gloss is already possible, or what do you mean?

uklomp:
Ah, sorry! You're right! Ignore this.

Another point: what did we decide again on the left hand glosses? When there's overlap, only the right hand is selected, but otherwise the left hand is also included?

Jetske:
Yes, all glosses from the right hand tier are added and only those from the left hand tier if they don't have overlap with a gloss on the right hand tier

susanodd:
This is deployed on signbank

uklomp:
Refreshing doesn't work. Uploading it again also doesn't work. I can send you an email with two .eafs that I could upload today and two that I couldn't, maybe you can spot the difference?

Jetske:
Yes please do!

Jetske:
Looks like the error was caused by a "_" that should have been "__" in my code... anyway I hope this solves it!
Also made small improvements to the feedback in case it finds annotations that are not in the dataset.

susanodd:
@Jetske are you able to add the gettext to the feedback so that it will be added to the PO files for translation? (Eventually.)

uklomp:
Ok, I will try again soon.

I have a new request (please let me know if you want me to make separate issues for these things, now or in the future); 
namely (actually like was done for the old example sentences connected to senses), if we could provide information if it's in question or a sentence containing negation. So, in the upload process, I'd like two boxes we can tick, one with 'question' ( 'vraagzin')  and one with 'contains negation' ('bevat ontkenning'). If these boxes are ticked, I'd like this information to appear on signbank as well.

And a question: I see that there's a field now, where I think it is possible to say something about the corpus / source, was it the idea that we put the link there? And should we do this manually for now, or were you maybe still looking into a way to do this automatically?

uklomp:
.eaf 43 and 72 still don't really upload so I'm afraid the bug isn't solved yet!

Jetske:
[...]

PR #1262, opened 2024-06-11
Annotated sentence improvements #1234

Jetske:
@susanodd  I just added the necessary tags for translating the pages, however I have never done this before and don't know how to make new .po files. Can you check my last commit? It that looks ok this is ready to merge.


susanodd:
> @susanodd I just added the necessary tags for translating the pages, however I have never done this before and don't know how to make new .po files. Can you check my last commit? It that looks ok this is ready to merge.

They're ready to translate, so that's the main thing. I'll see what happens with the feedback before generating the PO files again, since they get file names and line numbers in the PO, so the feedback can be translated in the existing file, or if it needs new English in the templates. But the main thing is that all text can be translated. So that's good!

PR #1263, opened 2024-06-12
Non manuele elementen

susanodd:
This is on signbank-dev

NME videos added as subclass of GlossVideo to allow in same table.
Necessary inheritance added to methods.
NME videos appear as for '`_small`' videos, with a suffix: '`_nme_N`' where the user provides N which is the order of appearance.

The unused Video class from original signbank has been removed.

For the GlossVideo, the `upload_to` field is now filled in by an `__init__` method. This is needed to support inheritance.

A languages field has been added to the overloaded `VideoUploadForObjectForm` class init method to support extra fields.


Woseseltops (commented):
Not much to complain, looks good @susanodd . The only thing I'm confused about is how descriptions are separate objects... couldn't this just be a textfield?

susanodd:
> Not much to complain, looks good @susanodd . The only thing I'm confused about is how descriptions are separate objects... couldn't this just be a textfield?

They need to be per language of the dataset. So the language needs to be paired with the text.

Jetske:
This could be a problem on my side, but I'm trying to run it and it doesn't work. I did checkout to your branch, ran 2 migrations, and then runserver. For some reason uploading a video does not work (not glossvideo, not NM video, nothing) but it also doesn't provide an error. When I revert the migrations and go back to my own branch it does work. 

The code looks alright but I'm going to look at it in more detail to find out what is wrong.

susanodd:
Did you try it on signbank-dev ?

Jetske:
@susanodd Yes, I have the same problems on signbank-dev. It just redirects to the page without giving any error. The video is not added in the file system either (at least locally). Same problem for all types of video. It looks like it did not try to convert it either.

Jetske:
The form is not valid in video.views.addvideo

susanodd:
That is very weird?!!

Did you completely close all the windows and try again? Maybe it has old code in your browser?

susanodd:
Can you try an mp4 file?

My browser only allows upload/opens video files it can upload.
(Remember when it wouldn't upload videos from your camera?)
I've given up on trying to do stuff with the conversion code.

Can you try with a file that you download from signbank, so it's already in the correct mp4 format?

Then I can see if it's the type of video.

Jetske:
So when I add 'required=False' to the new charfields in the class VideoUploadForObjectForm(forms.Form), then it all works. I find it a little suspicious that the glossvideo looks like it's loading when uploading a nme video, but it does not seem to give a problem.

susanodd:
> So when I add 'required=False' to the new charfields in the class VideoUploadForObjectForm(forms.Form), then it all works. I find it a little suspicious that the glossvideo looks like it's loading when uploading a nme video, but it does not seem to give a problem.

I can add that. Maybe it's a browser thing?

susanodd:
The addvideo method actually does nothing if the form is not valid. (It's always been that way.)
So I just added the required=False (being deployed now). So that would explain why it would just do nothing. (But it's always been coded like that if the form is not valid.)
Hopefully the nesting structure of the page didn't mess up the variables.
Maybe another form variable isn't being filled in?

Jetske:
Okay the reload of the glossvideo is caused by `onsubmit="return reloadVideo()` in the NME upload form. This may be deleted.

susanodd:
Okay, the repairs are deployed now on signbank-dev

Jetske:
Looks good. Ready to merge :)

Issue #1259, opened 2024-06-05
variants not for everyone visible after search query


uklomp:
When I look for a sign on Signbank, and then click one of the results, the other results are visible in a row at the top of the page. This is very convenient, but my colleagues do not seem to have this. Is it correct that this might be related to certain rights? Or is something else going wrong? In the screenshot below, I looked for 'zoek' and clicked the second hit. 

![image](https://github.com/Signbank/Global-signbank/assets/147159596/b1b8cb20-e916-41fb-a891-7bab6875e04e)

susanodd:
Everybody should be able to see this.
These are in a session variable and are passed around while you navigate to different pages. We normally call this the scroll bar.
The main problem is that it ends up hidden behind other things on the page if the screen is too narrow.
The Gloss Edit view is susceptible as it contains many things in the header, including the buttons for editing and space to show feedback messages if things fail.

Some pages do not show the search results. These are typically "helper" pages, like for adding a video or editing a lemma.

The scroll bar also takes time to "load". Sometimes it appears after some delay.

The "height" of the "header" is computed dynamically based on what is displayed. So sometimes this is incorrect.

@Woseseltops is the one with the most CSS experience. Hopefully he can take a look to see if there's something obvious that's wrong.

The code makes use of many different languages to generate the page. JQuery, javascript, HTML document entities, plus the Django markup for templates. The global "header" with all the datasets is also computed dynamically. The interactions of all these things kind of mismatch sometimes. (That it is visible sometimes and not other times.)

The specific code that "displays" the scroll bar is written in javascript using JQuery. All of the "items" are HTML elements that are generated dynamically. And this is supposed to "be inserted" into the labelled HTML element of the Django template. (So very many things can go wrong.)

susanodd:
@Woseseltops can you take a glance at the header and "affixed" css code at the top of Gloss Detail view and see if something is messed up with the dynamic calculation of the "top" area that is supposed to be always at the top?

Another thing about it is that it's transparent. So if you have some of the panels open and scroll up, you can see them "behind" the buttons and scroll bar area.

I made an issue about this (#1224) because the code in this view is totally unreadable.

Is it possible to modify the setup of the "fixed" things on the page? (The upper area below the menu bar and pull downs.)

It would be best if the separate parts could be in separate templates, but the translation does not work anymore if this is done. I previously tried to put modals into separate files, but they do not get translated anymore.
There are 1000 lines of Senses code inside one table row.
[...]

PR #1267, opened 2024-06-19
Crowded view

susanodd:
Changes for #1259 

This is on `signbank-dev`

Revised base layout header and menu bar so menu bar wraps and stays inside same element

This keeps the header bar elements from floating over the scroll bar and query button area.

To solve, only the datasets viewable by the user are shown in the header.
There are too many datasets to fit in the header on small screens. This causes things in the header to float around and interfere with other things.
The RU logo image path has been moved / put inside the html code instead of inside CSS where it was.
This to allow its dimension to be available and to put it on the right.
The background page color was removed from the "hands" logo SVG code. This was causing it to have dimension.

The scroll bar is now 100 % visible on an 11 inch Macbook Air. No need to keep reloading the page.

This branch will need to be tested  by colleages with different browsers and computer screens.


susanodd:
Reviewers are asked to help fix things that don't work on their browser......

susanodd:
The very last commit that makes `edit_text` become `edit_area` is not yet on `signbank-dev`. It's been tested on Safari with a small screen and fixes the remaining edit layout problems with width in gloss view.

Woseseltops:
Looks good @susanodd . The only thing I don't understand is what is the function of the `gloss skeleton` in 2-3 sentences?

susanodd:
> Looks good @susanodd . The only thing I don't understand: what is the function of the `gloss skeleton`, in 2-3 sentences?

No function. It was an early working file. To gather the overall nesting structure of the html elements.  I think it would help if the structure blocks could be made clear. There are references to specific element ids that give the structure. But in some places, these were no longer the blocks on the page.

(I will delete this. But the problem behind still remains wrt code maintenance. Header, Menus, Scrollbar, Affix Header, Navigation Panels, Left Column, Right Column, etc. None of this pops out.)

Issue #1130, opened 2024-01-24
Review /dictionary/info endpoint


vanlummelhuizen:
The `/dictionary/info` endpoint returns a json with a list of datasets the user has `change_dataset` permission for. When there is no dataset for which the user has `change_dataset` permission, a json with settings.LANGUAGE_NAME and settings.COUNTRY_NAME is returned.

This endpoint is used by the Signbank extension in ELAN. The second json as describe above is unusable in ELAN. 

We could/should do two things:
1. Make the necessary permission `view_dataset`.
2. Return an empty list if there is no dataset to choose from.

Any thoughts, @Jetske @Woseseltops @susanodd ?

susanodd:
New users only are given view permission (it used to be to NGT) and put in group Guest.

Do they need to be able to change things when they use ELAN?

There are zillions of permissions that aren't being tested for. Can a permission specific to ELAN be made and granted (to who? to group Guest? to Anonymous users?)

susanodd:
view_dataset seems the most fitting, since anonymous users can also view glosses that are public.

But maybe a second option for users that need advanced info about the glosses in the datasets?
(Is a distinction between view and change sufficient?)

Jetske:
I don't really know how this extension works but if users can make changes to the Signbank datasets through this extension in ELAN, option 2 sounds as the most logical/safest solution. Otherwise, option 1 would be ok I think.

susanodd:
@vanlummelhuizen see #1149  (scroll down a lot to the end where the most recent is shown)

I made it so if the user is anonymous, it provides in NGT glosses that are public.

I revised the fields returned in glosses.json

There were some bugs in the original fields. Also the API setting is obsolete.
I made it so the fields of Public View / Gloss List View are available.
I made the Translations be Senses (language name).
Do we need the Lemma fields too?

Should the info command of this issue also show the available fields?
Should they be available in Dutch as an option?

The package command is deployed on the development server as well as on signbank-dev.
(But dev has no videos, so the zip output on signbank-dev is probably dodgy for images and videos.)

susanodd:
I added some info to the wiki about this endpoint.

This endpoint probably needs to be expanded to show the available corpora for the dataset.
The dataset ID and corpora ID should also be provided to allow seamless integration with the new API routines.
The original package routine also needs to specify from what corpus the zip files are needed
and the others extended to retrieve from a particular corpus.

susanodd:
Can this be closed?

[...]

PR #1269, opened 2024-06-21
Dictionary info endpoint: permission change rollback

Woseseltops:
Debugging together with Divya, we've figured out that setting the guardian permission to `view_dataset` always results in just `[NGT]`, even though she has view permission on several datasets (which I can verify). If I make it `change_dataset`, like it used to be, the problem goes away. 

Divya has a deadline, so I propose:
* Rollback the situation, so she continue her work AND
* Figure out what's really happening here (is the permission named differently? is there something else happening with view permissions?)

If bullet point 2 takes too long, we can make it a separate issue.

### DIFF
--- a/signbank/dictionary/views.py
+++ b/signbank/dictionary/views.py
@@ -2310,7 +2310,7 @@ def package(request):
 
 def info(request):
     import guardian
-    user_datasets = guardian.shortcuts.get_objects_for_user(request.user, 'view_dataset', Dataset)
+    user_datasets = guardian.shortcuts.get_objects_for_user(request.user, 'change_dataset', Dataset)
     user_datasets_names = [dataset.acronym for dataset in user_datasets]
 
     # Put the default dataset in first position


susanodd:
Yes, we know this problem. I arose when Django was first updated from 1.11 to 4.x. The names of the Django guardian permissions the Django generates changed. The code was then changed to check for both view names.

Probably one checkpoint slipped by.

susanodd:
@Woseseltops the package info url is the original API.
@vanlummelhuizen was working on exactly this issue before #1130 

It's not clear to me what this url was originally supposed to do or who was using it.
At the time I started on the API, this is what we had.

Maybe check what the users expect this to do?
The easiest solution is also to grant the view_dataset permission to the user, that they have both can_view_dataset and view_dataset. We did not do this because it is overwhelming. It also wasn't clear how users originally got the permissions. There used to be a group NGT that gave users permissions. But somebody deleted this group.

susanodd:
@Woseseltops I think neither the change_dataset nor view_dataset should be needed for the info, since the existence of the datasets doesn't reveal any sensitive data.
That the change_dataset is used and works (?) is weird. I suspect the permission is computed incorrectly for this method.
The change_dataset should only be used in combination with a dataset. It's used for write permission on the dataset.
But if you only check for change_dataset, then it's going to show you (in this case, this url) it's going to show you all the datasets even if you can only change a couple.

(The original code did not compute this correctly, but it was written for Django 1.11.)

It needs to be changed for view_dataset and can_view_dataset. (Either of these.)
Or maybe the permission needs to be removed.
Or maybe checked which datasets have public glosses, then reveal those datasets?

(It's not clear to me what should be shown.)

susanodd:
@Woseseltops I had to solve this a different way.

It was not possible to deploy since you edited the file on the server.

I gave Divya view_dataset permission in the admin.

That the error arose in spite of having change_permission may have something to do with group membership.
Most users are in group Editor. But Divja is only in Researcher and Publisher.
Probably more long time users also need the permission added.

PR #1276, opened 2024-06-27
Move Keywords to Different Senses

susanodd:
This is live on signbank-susan

Part 1. Matrix modal converted to panel.
Part 2. Language modals converted to panels.

- Removed modals. Removed "keywords" column that was on the old page. Put placeholders that show the language name in empty fields. (The original version is visible on signbank.)

- One thing I am not able to do is due to the design of the signbank Language model. The languages need to be consistently sorted by ID in order to maintain the same sorting order. They cannot be sorted by name here because that depends on the interface language.

What is a good title for this page?
I like any: 

- Move Keywords to Different Senses
- Split Up Keywords into Different Senses
- Batch Edit Senses

This is intended so users do not need to do "cut and paste" and don't need to retype keywords in a different sense.
This helps the kind of RSI due to typing text.


susanodd:
This is live on signbank-susan

You need to do a query on one dataset, then go to Analysis -> Batch Edit Senses
or Analyse -> Batchbewerking betekenissen


Kindly give suggestions for improving the layout. (It's a bit flat looking, the layout elements don't pop.)

susanodd:
I need some help with this effing CSS. I can't get rid of the horizontal lines.
Moreover, inside the loop showing different glosses, the same exact code is arbitrary widths of the "row lines" in the Senses area.

css-row-lines-different-lengths

susanodd:
I tried to sort the languages by reverse of ID in order to put Dutch first instead of English.
This does not work. (It partly works but it seems that Django messes this up intermittently because of internal Django stuff and that the Languages model is sorted by ID. It can't be sorted by NAME of the language because the sort order changes based on what the interface is set to.
I gave up trying to put Dutch first. It's just sorted by language ID now.
(To avoid bugs in processing updates that make use of the language ID, so that the dynamic html elements are associated with the correct language. This view makes heavy use of the IDs of things inside the jquery processing.
Django only uses the language objects when first generating the page. After that the language references are actually the object IDs. So places that refer to languages are actually language ids.)

susanodd:
Here's the other panel,

split-senses-duivel

Woseseltops:
You've got my approval code wise, but before we put this live I'd like to know how Ulrika (as user representative) feels about the functionality. Okay with you @susanodd ?

susanodd:
> You've got my approval code wise, but before we put this live I'd like to know how Ulrika (as user representative) feels about the functionality. Okay with you @susanodd ?

Yes, I agree.
(It's basically the same code functionality as before, just different layout.)

uklomp:
I commented on the issue.

susanodd:
I'm going to merge this with master on Monday.

Issue #1278, opened 2024-06-27
The dictionary/info/ API endpoint does not check for the API token in Authorization:"Bearer ..."


vanlummelhuizen:

vanlummelhuizen:
> can we do this everywhere?

Yes, I already did in https://github.com/Signbank/Global-signbank/pull/1279/commits/cf9b1693daaa4e3499c41cc29a7e889930099548 which is now included in the PR.

> * let's make sure to also support the 'old' way of authenticating to this endpoint, so we're not suddenly forcing ELAN to make the switch.

To be sure I just tested the dictionary/info endpoint from my browser and it seems to work correctly.

> * Related recent pull request

@Woseseltops Just a reference or do you want me to look into that PR?

Woseseltops:
Just a reference!

vanlummelhuizen:
I am closing this issue because the solution was merged in #1279.

PR #1279, opened 2024-06-27
fix: Add a decorator to put API user in request #1278

vanlummelhuizen:
I tried to create a solution that is elegant and reusable.

### DIFF
--- a/signbank/abstract_machine.py
+++ b/signbank/abstract_machine.py
@@ -12,7 +12,7 @@ from signbank.settings.server_specific import LANGUAGES, LEFT_DOUBLE_QUOTE_PATTE
 from signbank.dictionary.update_senses_mapping import add_sense_to_revision_history
 from django.http import HttpResponse, HttpResponseRedirect, HttpResponseForbidden, HttpResponseBadRequest, JsonResponse
 from guardian.shortcuts import get_objects_for_user
-from signbank.api_token import hash_token, put_api_user_in_request
+from signbank.api_token import put_api_user_in_request
 
 
 def get_interface_language_api(request, user):
--- a/signbank/abstract_machine.py
+++ b/signbank/abstract_machine.py
@@ -12,7 +12,7 @@ from signbank.settings.server_specific import LANGUAGES, LEFT_DOUBLE_QUOTE_PATTE
 from signbank.dictionary.update_senses_mapping import add_sense_to_revision_history
 from django.http import HttpResponse, HttpResponseRedirect, HttpResponseForbidden, HttpResponseBadRequest, JsonResponse
 from guardian.shortcuts import get_objects_for_user
-from signbank.api_token import hash_token
+from signbank.api_token import hash_token, put_api_user_in_request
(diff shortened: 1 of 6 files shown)


vanlummelhuizen:
A remark: as it is now implemented there are no errors that could be passed to the client like what happens in https://github.com/Signbank/Global-signbank/blob/fb97037616378827661b13c5530d282184b4a13c/signbank/api_interface.py#L100

I think I can implement it so that errors could be passed. I will work on that next week.

Woseseltops (approved):
Fancy walrus operator

vanlummelhuizen:
> Fancy walrus operator

Nice that you noticed. Unfortunately (?) I removed it in my latest changes. 

I think this is ready to merge now, if approved of course.

vanlummelhuizen:
> I think this is ready to merge now, if approved of course.

Hmm, I keep finding things to improve in the API. Perhaps keep this PR open until I have finished checking all API endpoints.

susanodd:
Looks good to me! Much cleaner.

vanlummelhuizen:
@Woseseltops @susanodd @Jetske 

I have gone over all endpoint as documented in https://github.com/Signbank/Global-signbank/wiki/Signbank-API and added the `put_api_user_in_request` decorator and changed some code where applicable.

**Note:** The following function seems to need some work. For instance, the `errors` variable is a string that is never send back to the user. https://github.com/Signbank/Global-signbank/blob/fb97037616378827661b13c5530d282184b4a13c/signbank/api_interface.py#L468-L505

vanlummelhuizen:
Ready to merge afaic

Jetske (approved):

susanodd:
> @Woseseltops @susanodd @Jetske
> 
> I have gone over all endpoint as documented in https://github.com/Signbank/Global-signbank/wiki/Signbank-API and added the `put_api_user_in_request` decorator and changed some code where applicable.
> 
> **Note:** The following function seems to need some work. For instance, the `errors` variable is a string that is never send back to the user.
> 
> https://github.com/Signbank/Global-signbank/blob/fb97037616378827661b13c5530d282184b4a13c/signbank/api_interface.py#L468-L505

Hmmm. I see that. It looks like it might be testing things that are already tested at a previous step.
Sometimes, to make type check etc it is needed to check for things that don't happen.
(Not related, but an example of, when we know something is unique, but use "first" anyway and ignore the case when there could have been multiple objects. Then test if first is empty. There is lots of code that does this for the dataset languages.)

But could it happen I guess if somebody knew about the url and didn't follow the previous steps.
But it looks like I just ignored these, but tested anyway. (Remove code? Or give a Http warning and crash out?)

susanodd:
I actually don't know how to "approve" on GitHub.

Woseseltops (approved):
Improves maintainability

susanodd (approved):

susanodd:
> Ready to merge afaic

Okay, it's probably time.

Issue #1270, opened 2024-06-21
Cannot create gloss for fresh install


Woseseltops:
This is a fresh install of Signbank https://signbank4.vlaamsegebarentaal.be/

If you go to the admin, you can log in with these temporary test credentials: 

user: divya
pwd: divya!23

When you try to create a gloss, you get an error, though... any idea what's going on there @susanodd ?

susanodd:
I'll look at it.
I know when signbank was new there needs to be some already existing data in some tables.
I will look at my notes from the install up signbank-susan.
Is there an "empty" signbank database anywhere?
It won't work to start with the "original" signbank since the migrations have points where they can't migrate from the previous. (That has to do with the populate methods that relied on old fields, then mapped the data to new fields, the removed the old fields. Namely the "name_english" etc fields. So the migrations can't be applied.)

I will check it out.
That's off the top of my head.

susanodd:
First, it gives a message that there are no pages, login to define some.
(This should not happen. Most of the pages have templates. Only the About pages and Copyright do not, and are visible for anonymous users.)

susanodd:
I changed the site name in the "sites" in the admin.
But still when I then type the page after the site name in the browser, to my surprise I get a message to contact Onno (old email).

susanodd:
In the admin, I added a page `/about/copyright/` with some text

But there are no menus.
To see it, you need the url

https://signbank4.vlaamsegebarentaal.be/about/copyright/

susanodd:
Maybe make a create or insert sql with the required contents for the tables?

susanodd:
It works to add feedback (by typing in the page in the url by hand, since there are no menus)

https://signbank4.vlaamsegebarentaal.be/feedback/overview/

It looks like in our Admin python code we did not implement methods for adding glosses or lemmas. (There are no `+add` in the column next to the table.)
To create a gloss, you also need to have a lemma, but it's not shown to add those either.

Normally, because creating a gloss involves lots of steps, also some fields need to be filled in before it can be saved.

susanodd:
The settings file is not properly defined.

I typed in this url (by hand since there are no pages):

https://signbank4.vlaamsegebarentaal.be/signs/search/

It complains here:

` language = Language.objects.get(**DEFAULT_KEYWORDS_LANGUAGE) `

this setting does not exist.

Here,

```
DEFAULT_DATASET	
'Your Dataset'
DEFAULT_DATASET_ACRONYM	
[...]

PR #1282, opened 2024-07-03
Api perm check 1281

vanlummelhuizen:
Did some work to get rid of the can_view_dataset permission. 

When running tests, only signbank.dictionary.tests.ManageDatasetTests.test_User_is_dataset_manager fails, but that seems a bug in that test because it is trying to change a Dataset but it expect a message about whether the user can view a dataset.


vanlummelhuizen:
I hope this PR does not bite PR #1279 ...

susanodd:
> Did some work to get rid of the can_view_dataset permission.
> 
> When running tests, only signbank.dictionary.tests.ManageDatasetTests.test_User_is_dataset_manager fails, but that seems a bug in that test because it is trying to change a Dataset but it expect a message about whether the user can view a dataset.

I can look a the tests. They're mainly helpful for debugging code.
The tests look at the feedback that is put in messages, because there is no other way to find out if something actually worked.
So it could be the message that got changed then the test will fail. We started doing more things with translations. But the tests need to keep using English for the messages testing. (So a test would fail if you're running in NL.)

susanodd:
So what is the procedure for deployment?
We can do it first in signbank-dev.

Get and Pull the branch (or just merge it first), then execute the command in the venv, then create a new test db.
Then it's okay to run the deploy.sh and the tests.
The code needs to be updated on the server and the database modified and a new test database created before the normal deploy command will work. (Chicken or the egg going on.)

Woseseltops:
Thanks for cleaning up, @vanlummelhuizen :) . Two thoughts after reviewing the code:

* The management command that you added is only temporary, right? If so, it might be good to add that to the docstring, so that we are not scared to remove it if we encounter it later

* Very good to use our unit tests to verify this works. I'd like to manually test this with Divya's specific usecase to ensure we're not breaking anything for her (given her deadline) before we put it live, though... Let me know if this is on signbank-dev. In case somebody else wants to test, this is what I would do:
  * Create a test user with Divya's roles
  * Add and remove viewing permissions for various datasets
  * See if the `dictionary/info` endpoint responds with the correct datasets

susanodd:
At the moment signbank-dev has the branch  batch_edit.

The most recent updates to batch_edit to address the permissions (feedback of @vanlummelhuizen) are not live yet.
Probably best to do this branch before?

So it's okay to put this branch on signbank-dev.
@Jetske is also wanting to use signbank-dev but is getting a new container.

susanodd:
@vanlummelhuizen is there a migration missing?
The meta of dataset has been changed by removing the can_view_dataset.

susanodd:
I'm working on this locally to add the migration and fix the test.

susanodd:
I figured out what is wrong with the test.
The test dataset is set to NGT.
But the test in question is supposed to make sure the user has view permission before being granted change permission.
For NGT this won't work because NGT is a public dataset, so the user always has view permission.

The test used to make a new dataset and grant and revoke permissions to it. (NOT use NGT.)

susanodd:
@vanlummelhuizen @Woseseltops you can inspect the permissions now in the Admin on signbank-dev.

It is this branch.
The command was run.

But if you look at the permissions for NGT, the users that have change permission do not have view permission anymore!
(Visible in Datasets -> NGT -> Object permissions)

For example, Casper can change the dataset but cannot view the dataset. (This is wrong.)

susanodd:
> Did some work to get rid of the can_view_dataset permission.
> 
> When running tests, only signbank.dictionary.tests.ManageDatasetTests.test_User_is_dataset_manager fails, but that seems a bug in that test because it is trying to change a Dataset but it expect a message about whether the user can view a dataset.

The tests for Dataset Manager ought to make use of this dataset:

`settings.TEST_DATASET_ACRONYM`

instead of NGT.
Because NGT is a public dataset so it isn't possible to remove the view dataset permission.

susanodd:
> @vanlummelhuizen @Woseseltops you can inspect the permissions now in the Admin on signbank-dev.
> 
> It is this branch. The command was run.
> 
> But if you look at the permissions for NGT, some users that have change permission do not have view permission anymore! (Visible in Datasets -> NGT -> Object permissions)
> 
> For example, Casper can change the dataset but NOT view the dataset. (This is wrong.)
> 
> Divya can change AND view the dataset.
> 
> More comical, @uklomp can CHANGE dataset Oefen, but she can't view it! (She's the only one that can't view it.)

Okay, for Casper and @uklomp the permissions were also that before the new code. Neither had view permission.

Probably we ought to make sure the original permissions are correct before applying the command.

Note: The command was applied to signbank-dev first, before the migration and before the deploy (with the code updates that check the permissions.)

susanodd:
I updated the perms by hand for a couple of datasets in the admin.
But as this branch does, it is less work to use a command.

But I think there needs to be a second command.

- For all the datasets, for all the users:
if the user has (exact match, not "any") permission can_view_dataset
then grant the user also permission view_dataset
if the user has (exact match, not "any") permission change_dataset
then make sure the user also has view_dataset permission and grant this if the user does not have this

Then both kinds of view permission columns will be filled in in the Object Permissions table for the Dataset in the Admin.
And any user that has permission to change the dataset will also have permission to view the dataset

It's important to use explicit match, not "any" match.
The idea of the dataset permissions is that the dataset manager has granted them.
Not that the user is somehow else getting them.
This should be visible in the Admin table.

The main use of "any" is that staff and superusers can have extra permissions that are not shown in the table.
For other permissions, the users get this via the Group, e.g., Editor, like for change_gloss, etc

vanlummelhuizen:
> * The management command that you added is only temporary, right? If so, it might be good to add that to the docstring, so that we are not scared to remove it if we encounter it later

The alternative is to change it from a _command_ into a _data migration_. This makes sure it is run once. After that we could create a migration that removes the can_view_dataset.

About creating view permissions for users that have change permission: I think it does not matter if it is done before or after the command/data migration and the migration removing can_view_dataset, for the simple reason that the command/migrations do not touch upon change permissions.

susanodd:
> > * The management command that you added is only temporary, right? If so, it might be good to add that to the docstring, so that we are not scared to remove it if we encounter it later
> 
> The alternative is to change it from a _command_ into a _data migration_. This makes sure it is run once. After that we could create a migration that removes the can_view_dataset.
> 
> About creating view permissions for users that have change permission: I think it does not matter if it is done before or after the command/data migration and the migration removing can_view_dataset, for the simple reason that the command/migrations do not touch upon change permissions.

Okay, that sounds good.
Via a data migration. That is safer.

susanodd:
No the info endpoint is a read only access, so the user should not need change permission.

@vanlummelhuizen is going to make a migration to fix the permissions, so the table contents are changed at the same time as the model. Otherwise it doesn't work to deploy because once the model is changed, the code can't be run anymore because it references old preferences. So @vanlummelhuizen is going to write a migration that maps the values.

susanodd:
@vanlummelhuizen I removed the migration that was needed but unwanted.

This branch has been merged with all the recent updates so there are no conflicts.

I put the "delete gloss" branch onto signbank-dev because the others were all busy.

When you want to you can use it again for this request.

I replaced the database with a new one that does not have that command applied.

susanodd:
signbank-dev is now showing branch master

It can be switched to this one in order to develop the migration (to replace the command)

susanodd:
NEW UPDATE

I created an empty signbank (see #1316) 

If you browse the changes to the code there are numerous changes to the permissions being tested or assigned.
The code passes all of the tests.

I did not delete the old permission can_view_dataset, but it is not used anymore.

We still need to migrate the master signbank database.

Because the permissions differ in #1316, I'm not sure what is going on, why there are different names for the permissions. To get the code to run, I kept putting print statements to inspect what permissions were being found and then adapt the code and make sure the tests match. (So both sides work.)

I'll try the revised code on a non-empty database, too.

Then we still need a migration with data mapping from this branch.

susanodd:
This branch should be completed before the empty database branch is delivered so the old permissions are removed via migration. (#1316 #1270 )

vanlummelhuizen:
This is now done as far as I am concerned. @Woseseltops @susanodd

vanlummelhuizen:
See also https://github.com/Signbank/Global-signbank/pull/1316#issuecomment-2324070002

susanodd:
Merged with master, all the tests succeed!

Issue #1283, opened 2024-07-04
User has problems creating annotated sentences


Jetske:
One user has reported that they are unable to create annotated sentences. Everyone else who has tried has succeeded, so it's hard to say what is the problem. I'm thinking that the video (this could be the videofile and/or eaffile) never ended up as InMemoryUploadedFile/TemporaryUploadedFile correctly, so it was lost before processing the request. 

![image](https://github.com/Signbank/Global-signbank/assets/10157482/b1c3f391-782b-42f2-85db-fb8107f8683a)

Woseseltops:
The first step is (nearly) always that we are able to reproduce... in the past, users that create videos of their problem have been really helpful, perhaps this user is willing to create one too?

Jetske:
@susanodd this was just solved while calling and editing live on signbank-jetske, thanks for your help to set it up!!

PR #1285, opened 2024-07-09
Remove annotated sentence file checks #1283

Jetske:
For one user the video file did not pass these tests so the file was not in the expected temporary location or something like that. I guess it's ok to remove them, if you agree. The fields are required already, so it can only go wrong if the eaffile is not actually an eaffile but that would be incorrect use of the upload function and we can check that in other ways if necessary :)

### DIFF
--- a/signbank/dictionary/models.py
+++ b/signbank/dictionary/models.py
@@ -3994,13 +3994,7 @@ class AnnotatedSentence(models.Model):
         """Add a video to the annotated sentence"""
         from signbank.video.models import AnnotatedVideo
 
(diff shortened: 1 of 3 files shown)


susanodd:
If you look at the code of the various "other media" there are some extra checks on file type that inspect the head of the file to determine its type (python magic).
It was written to make sure people didn't put files by changing the extension. Normally the code only looks at the file type in the file header.
We had some cases where the file could not be recognized.
So some of that code for inspecting the actual type might be of use.

PR #1296, opened 2024-07-24
Delete gloss api

susanodd:
Implemented the achived field on glosses.

The various queries over Gloss all make sure archived is set to False before inclusion in the results

The various Relation and Morphology classes all exclude archived glosses from the results.

The archived glosses are stored in DeletedGlossOrMedia as before and can be viewed in the Admin for this class.
This allows the deletion to be undone.

A restore has been implemented. On Signbank, it is available via the Lemma Detail view.

OPTIONAL: a hard delete on very old archived objects.


susanodd:
This branch is on `signbank-dev` now.

The "delete" button of Gloss Edit (lower left in template when Edit is active)
now sets the gloss to archived.

It should not show up anywhere anymore.

Feel free to assist in testing whether this works everywhere.

susanodd:
I guess all the explicitly coded `Gloss.objects.get(id=` and `get_object_or_404` calls outside of views also need checks. Hmmm.

Woseseltops (commented):
As expected, this touches a lot of lines of code, well done. The only thing I wonder: shouldn't there be a migration file to add the `archived` field?

susanodd:
> As expected, this touches a lot of lines of code, well done. The only thing I wonder: shouldn't there be a migration file to add the `archived` field?

Yes, I already added the archived field via a migration. That's the first thing I did. The master and all the development servers (except @Jetske) have this migration applied.

susanodd:
I'm working on the patterns `Gloss.objects.get(id=`

There are quite a lot of these, but the code as it exists makes no allowance for a gloss not being found:

https://github.com/Signbank/Global-signbank/blob/7eb63385c50c5655d30f4b49b5fab45f370a7cee/signbank/dictionary/update.py#L331-L334

I could rewrite the code so it queries on the gloss id instead of retrieving it first?
(But there are lots of these.)

When can we assume that the "gloss" being handled by methods isn't an archived gloss?
Presumably the caller of this method has also checked?

Woseseltops:
I don't think adding `archived=False` everywhere would be worth the time investment; even if the user somehow finds a way to do make changes to an archived gloss... doing so wouldn't actually break anything, right?

susanodd:
> I don't think adding `archived=False` everywhere would be worth the time investment; even if the user somehow finds a way to do make changes to an archived gloss... doing so wouldn't actually break anything, right?

Yes, that's true. It's basically made invisible, and other glosses that referred to it via relations and morphology don't see it anymore.
Okay, that makes sense.
I just need to check the API functions, since that is the point that those should not see the archived ones.
Yes, it seems excessive if all the internal functions called by Gloss Edit are also checking.

susanodd:
The most recent commits are not deployed yet to signbank-dev

susanodd:
Another detail about lemmas.
In the Lemma List, I modified the Num Glosses to show the count without the archived glosses.
However, this will be a problem.
You are not allowed to delete lemmas, except for when they do not have glosses.
This is because glosses must have a lemma. That is where they get their dataset info from. The dataset used to be stored in the gloss, but then it was moved to the lemma.
So for archived glosses, the user also can't delete the lemma because then the glosses won't be in any dataset.

So I can make an additional column in the lemma list to show "archived glosses" so that the count makes sense.
Then in the Lemma View (only visible via a link in the lemma list), it can also show archived glosses.
(Because the user will want to know what's going on.)

susanodd:
This branch needs the bug fixes and implementation modifications done in the last five weeks in order to implement the delete.
This branch has older versions of code because bugs have been fixed in the batch edit and split senses branches.
I will not keep fixing code that has already been fixed in other branches. The other branches need to be merged into this branch. It is blocking this branch because it needs newer code from those branches.

susanodd:
> I don't think adding `archived=False` everywhere would be worth the time investment; even if the user somehow finds a way to do make changes to an archived gloss... doing so wouldn't actually break anything, right?

After much more functionality, new response

The places where the archived gloss would be seen (via related glosses and morphology), it is now hidden.

One place it is visible in in Lemma Detail / Update (via the Lemma List / Search)

Because the gloss has been archived, it still shows up in the count of the lemma group.
This needs to be there because although hidden, the gloss still affects database constraints.
So here it shows up in the counts of glosses in the lemma group. An extra column has been added for "Archived Gloss Count".
Then in the Lemma Detail, the archived glosses show up in the group, but without a link and with "(Archived)" instead.

A gloss can be restored via the API. This is meant to assist in automated API routines. Once restored, the delete and restore will be visible in the restored gloss's revision history.

A logical place for a Restore button for a gloss would be in the Lemma Detail where it is visible as being archived.

Woseseltops (approved):

PR #1302, opened 2024-07-31
Non manuele elementen API

susanodd:
This on `signbank-test`

Base64 encoding is used for the video because
1. that's what javascript uses once put in a buffer
2. the file is passed inside of a json data structure which needs to be JSON.stringified
3. this is working

TO DO

Add to API Wiki

### DIFF
--- a/signbank/api_interface.py
+++ b/signbank/api_interface.py
@@ -81,6 +81,8 @@ def api_fields(dataset, language_code='en', advanced=False):
         for field in gloss_fields:
             api_fields_2023.append(field.verbose_name.title())
 
+        api_fields_2023.append(gettext("NME Videos"))
+
     return api_fields_2023
 
 
--- a/signbank/dictionary/models.py
+++ b/signbank/dictionary/models.py
@@ -1531,18 +1531,39 @@ class Gloss(models.Model):
 
         sequential_morphology_fieldname = gettext("Sequential Morphology")
         sequential_morphology = self.get_hasComponentOfType_display()
-        if sequential_morphology:
+        if sequential_morphology_fieldname in fieldnames and sequential_morphology:
             fields[sequential_morphology_fieldname] = sequential_morphology
 
         simultaneous_morphology_fieldname = gettext("Simultaneous Morphology")
         simultaneous_morphology = self.get_morpheme_display()
-        if simultaneous_morphology:
+        if simultaneous_morphology_fieldname in fieldnames and simultaneous_morphology:
             fields[simultaneous_morphology_fieldname] = simultaneous_morphology
 
         blend_morphology_fieldname = gettext("Blend Morphology")
         blend_morphology = self.get_blendmorphology_display()
-        if blend_morphology:
+        if blend_morphology_fieldname in fieldnames and blend_morphology:
             fields[blend_morphology_fieldname] = blend_morphology
+
+        nme_videos = gettext("NME Videos")
+        gloss_nme_videos = self.get_nme_videos()
+        if nme_videos in fieldnames and gloss_nme_videos:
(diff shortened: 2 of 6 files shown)


susanodd:
I'm still having trouble "making json" of the POST fields to send in the API.

I can't find any examples of python request either. This is glossed over and only examples of normal dicts with strings are shown in the documentation.

We have

```
        ....
        type: 'POST',
        headers: { 'Authorization': 'Bearer XXXXXXXXXXXXXXXX',
                         'Content-Type': 'application/json',
                         'Accept-Language': 'en' },
        data: .......,,
        datatype: "json",
        async: false,
        success : .......
```

The arguments are taken apart in Python:

`    post_data = json.loads(request.body.decode('utf-8'))
`

The body needs to be used because of what Django does (undocumented, this is because we include headers now for the API token)

The "data" needs to be stringified.
But for the NME videos, the data includes "normal" string fields, plus it needs a "file" field.

JSON.stringify(
{```
 'Description (Dutch)' : 'blah blah',
  'Description (English)' : 'blah blah',
 'Index' : '2',
 'File' :  ?????????????
```
})

The File field needs to be "loaded as data" not a link to something.
I have tried various formats inside a promise in javascript for making sure the contents of the file is obtained.
But on the Python side it can't process this.
We need like nested stringify and encoding/decoding because the "file" is also encoded, then encoded again inside the json.

**Comment by Woseseltops** [comment]
Since you added a 'draft' label, I also officially set it to 'draft', so it's not showing up as something to be reviewed :)

**Comment by susanodd** [comment]
Sure, I added DRAFT because it's not finished yet.
But that doesn't mean I don't need help.
We're supposed to collaborate on this.

**Comment by susanodd** [comment]
I confirmed that all three operations work correctly on signbank-test.

I added the create to the Wiki (with signbank-test for now)

**Review by Woseseltops — COMMENTED** [review]
Three things:

* I'm not sure about this bit when processing the uploaded file

```python
        except (OSError, EncodingWarning, UnicodeDecodeError):
            pass
```

If something failed, shouldn't that be returned to whoever is calling this API endpoint?


* I like f-strings (introduced in 3.6, see https://docs.python.org/3/tutorial/inputoutput.html#tut-f-strings). Something like

```python
gettext("Description") + " (" + language.name + ")"
```

becomes

```python
gettext("Description") + f" ({language.name})"
```


* Minor thing, I guess `%s` doesn't add much here ;)

```python
info_field_name = gettext("Description") + ": %s" % language.name
```

susanodd:
Good point!!

I'm actually working on this for the Revision History. (The stuff isn't being displayed in any useful way at the moment.)
(This is because something needs to be saved, and then later decoded to display what was saved.)

susanodd:
For this part,

        except (OSError, EncodingWarning, UnicodeDecodeError):
            pass

There can be an error in any of the commands.
The field isn't set and it reports back that the file is missing.
The error can arise when it's creating an NME object as well.
It doesn't catch video format errors (if the file isn't actually mp4).
I tried to make an all encompassing error catcher.

If the file was not put in a buffer on the client side, then it ends up empty.
It seems impossible to catch all possible kinds of errors.

The file is uploaded to "writable/tmp" and then that file is saved in the new GlossVideoNME object
(using the already existing code for the NME object).

The "getting the data out of the json data" is very "finicky" and Django doesn't let you access it at other places, only right after you get the post data.

I previously tried in other API routines to elegantly split up the routines that process the data, but it kind of disappears and can't be fetched out of the POST except when it's first accessed.
So the "saving the file to tmp" is done right away.
Then the NME object is created at the same time. (so it doesn't need to clean up after itself if something goes wrong. It knows immediately because it gives the except error.)

I didn't consider yet how to return errors in the same function that retrieves the json data.
It's basically only meant to construct a dict of the post data.

PR #1314, opened 2024-08-19
AnnotatedGloss list view

susanodd:
COMPLETED
-  group by annotated gloss
GROUP BY is complete. The sorting is by gloss annotation in either the interface language or the default language of the dataset. [CODE] the nature of sorting to allow group by in the template makes use of ids (gloss id, then sentence id for that gloss). The outer level sorting on the gloss is done after the query results are retrieved. The objects in the results need to maintain the grouping in order for regroup to work in the template. The query results span multiple pages. Sorting on other columns interferes with the group by. In other views, the arrow buttons evaluate the query again to complete the sort. This does not work here because of the group by, so the interface language is used for sort.

- added two fields to search on "is representative" and "annotated sentence contains" in Query form
- videos for sentences in list (it's just the path now) (I'm going to leave it as just the path, I added the video for the gloss)
- a Detail Annotated Sentence view has been implemented; an Edit button links to Edit Annotated Sentence (of @Jetske)
- the scroll bar is maintained in all the views where the scroll bar is visible
- the scroll bar shows the first 5 words of the sentence
- compatibility with Query View (for the Gloss query parameters)

TO DO

- merge with the branch of @Jetske Manage Sentences (that branch needs to be merged with master, it contains a hidden page that will be made visible when merged, we implemented two aspects of the issue on separate branches.) The pages need to be linked using url parameters that let the user return to either the list of annotated glosses, the manage sentence page, or the gloss detail page. This needs to be passed to the "save" url in order to return correctly using the context. Previously (on master), the Edit Annotated Sentence was only available via Gloss Edit.

OPTIONAL

- [optional] query parameter and search results passed to Search / Manage Annotated Sentences (of @Jetske)
- [optional] shorthand for showing a sentence abbreviation in the scrollbar of search results [discussion needed the user would need to specify this and perhaps add a field]


susanodd:
@Jetske I updated the code for this.

It shows duplicates of the annotated gloss in the same sentence. I made screenshots [see the two below].
This must be because of the 'starttime" and "endtime" fields?
I needed to create a detail view for an annotated sentence. I put the search results in the scrollbar (of type annotated sentence -- for now). Eventually the results can be passed around because they are in the session variable search_results.

Duplicate Annotated Gloss IDs for the same combination Gloss and Sentence.
search-annotedgloss-list-duplicates

The Annotated Sentence Detail View showing the scroll bar and the (duplicated annotated glosses).
annotatedsentence-detail-view-duplicates

TO DO
add the start time to where the gloss occurs in the sentence?

NOTE
this looks nothing like your display in Gloss Detail View.

susanodd:
Questions:

How useful is such a scroll-bar?
It can basically be anything. The individual items link to annotated sentences.
However, because the items also contain references to glosses, there end up being duplicates.

I was thinking that it's possible to access "different" attributes of the scrollbar to allow the items to be used / referred to in different other views and as query input to other views.

Jetske:
For the Annotated Sentence Detail View wouldn't it be easier to just reuse the edit annotated sentence page or does that serve another purpose?

susanodd:
> For the Annotated Sentence Detail View wouldn't it be easier to just reuse the edit annotated sentence page or does that serve another purpose?

It took me a while to "hide" (remove) the things specific to editing. There are many additional fields there.

For the "detail view" it does not have anything related to editing. I can put the things back after I get other things working.

This may seem backwards, but the dark option has made me think about the code differently. Those "edit edit" things with pull-downs need lots of extra code. The Gloss Edit view was really involved to get all the parts to work in dark mode.

I can work on it more. For now, I was just trying to get the "scroll bar" to work.
The search results types didn't include anything appropriate, so I made a new type annotatedsentence. But some additional functionality is needed to allow the user to also click on glosses. (You can see the gloss in the scrollbar when in the annotated sentence view when the search was via the annotated gloss search, but the scrollbar will disappear if you go from the glosses (click on glosses visible inside the annotated sentence view) to gloss view (then there is no scroll bar, which is confusing for the user.)

susanodd:
@Jetske is it possible to have an abbreviated form for the annotated sentence? I made it so the first 20 characters show in the scrollbar. But for some this isn't very helpful, it would be cool if it was some buzzwords or key part.

The scrollbar works now! you can switch between the annotated sentence (after searching on annotated glosses) then switch to Gloss view and you keep seeing the scroll bar.
(It passes the search results that are in the session variables, until you do a new search or search on say handshapes instead.)

susanodd:
@Jetske in the first two sentences (they are sorted by sentence ID at the moment, need to add sorting still)

There are glosses "palm up" is this one of those NME things? Is it morphology ? It doesn't seem like that's the same as a word.
Should there be "extra stuff" on the annotated glosses? (when shown in order for the sentence.)

Jetske:
@susanodd awesome! 

For the scrollbar you can maybe show the representative glosses (or their lemma) to have a short version of the sentence?

susanodd:
> @susanodd awesome!
> 
> For the scrollbar you can maybe show the representative glosses (or their lemma) to have a short version of the sentence?

Are there multiple representative glosses per sentence?

Jetske:
Yes, I think many sentences will have multiple representative glosses. The combinations won't necessarily be unique but I think they will be the most important glosses that distinguish the sentences.

susanodd:
@Jetske if you search on "praten" you get as a result 11 annotated glosses.

You can see the IDs of the Annotated Gloss and the ID of the annotated sentence.
Some of the sentences look identical but have different sentence ids.

annotated-gloss-search-praten

susanodd:
I'll see if I can do something about the duplicates. They're not actually duplicates wrt to the sentences because of position, but it looks that way in the other detail views.

susanodd:
Any idea why there seem to be duplicates of the sentence text in the sentences? Is one a concatenated version?

susanodd:
@Jetske the "sort arrows" in the list views actually force a new query, they don't merely rearrange what is displayed.

I made it so the Annotated Gloss list view sorts by the interface language or the default language of the -- first selected -- dataset (if the interface language is not available).

When I wanted to use the "arrows" to update the sort, it totally doesn't work.
The grouped results end up not grouped anymore.
It's also not possible to sort by columns within the "grouped" part because they can't sort as a list.

I removed the arrow buttons and leave it now as sorting per interface language.

The sorting with the groups is also made weird because of pagination.
Using the Django regroup it works and the pages also show correctly.
But using the sort buttons messes up the pagination as well as the groups.

To avoid excessive waste of time, I'm just going to omit the arrows.

susanodd:
@Woseseltops read the "issue" description above for the summary; ignore the discussion blog above.

susanodd:
I've merged this branch with the work of @Jetske for managing annotated sentences.

It looks like the "redirect" is properly implemented and should return the page to the calling page.
However, since I don't know anything about updating an annotated sentence (it includes a video that can be trimmed as well as an eaf file), I am not able to test the "save" of an edited annotated sentence to see if the redirect to caller works.

We'll wait and see if users experience bugs.

PR #1316, opened 2024-08-22
#1270: Signbank empty database.

susanodd:
Modified permissions. Command to empty tables, command to remove users except for developers.

[ORIGINAL, SKIP THIS]
The empty database is created with the same name as the test database, it does not overwrite the original database.

Then to use the new empty database instead, you need to copy it to overwrite the real signbank.db (on signbank-test this was done.) It is intended that other developers (the VGT-ers will start out with the empty database).

Then a new test database need to be made (the create_test_db throws away more table contents.)

This is a TEST DATABASE

This is on signbank-test with an EMPTY database. (Well, nearly empty, the interface translations are still there..)

ALL TESTS SUCCEED!!

I was able to create a gloss in tstMH.

NEXT

Remove interface languages. This will require a migration, so this branch won't be compatible with other branches.
The settings for this server will need to change the model translation settings.
I'll see if this can be a script so it can be used optionally.

Some of the permissions probably need to revert to what they were. It was a problem at first because there was no data. The Dataset pages did not show any of the datasets and the original permissions did not work.

Specific to VGT, see if they want to retain the VGT dataset glosses.
This would facilitate sharing in the future since the gloss IDs would be retained.
One of the issues of @ocrasborn was to add extra relations to other sites (#93)

[READ THIS, UPDATE PER 02-09-2024]
After the new permissions (#1281, #1282), the empty database has zero permissions to choose from in the Admin.
All the "change_dataset, view_dataset, etc have disappeared from the permissions. It's not possible to grant users permissions in the admin because there are no permissions.
Does the "vacuum" command do this? Or are too many tables erased, but not shown in the log of tables erased?


susanodd:
@Woseseltops this is on signbank-test.

If you look at the admin you can see what the contents is.

I need to make a migration to remove the extra interface languages.
(Go back to just EN and possibly NL. Or what to do here?)
The end developer should be able to decide what languages to include.
English must be there because the templates are in English and that is the language from which the other languages are translated in the PO files.

All of the Field Choices and Handshapes are in the database. It won't work without them. Or at least some. (They can't be empty.)
(At the very least they need machine values 0 and 1 as constants.)

susanodd:
@vanlummelhuizen can we make a conditional migration to remove the model translations that are not needed?
Then the users that start with an empty database can keep up to date with the master branch.
Of course, they could add their own model translation languages, so then they'd have extra migrations on their installation.

(I write this in general although this is for the VGT site. #1270 )
Can we have Django set the migration to "applied" if the model translations don't match (either direction)?

susanodd:
@vanlummelhuizen I needed to alter the permissions in the Dataset Select page when I started using the empty signbank. It did not want to display anything.

Then I went ahead and changed them elsewhere to be the same. But then the tests didn't work, so I changed the ones in the tests (manage dataset tests). (Although I had to remove the checks on "messages" returned because it wasn't showing any messages. So it merely checks the status of the operation now.)

Perhaps you can check out the permissions?

I removed users in a command, and I left more tables in the database (create_empty_db), based on create_test_db but it didn't work with the users removed, since I couldn't login anymore, so I left the development team, plus the future user of the VGT site.) I removed datasets by hand. It didn't work to remove users by hand. (Some error about a content type on some users, internal to Guardian.) The removal of the datasets will be needed to be added to the create_empty_db command. Maybe we just leave the test dataset? Plus there likely need to be some "start" data in the database because it's not possible to check that everything works without any glosses. It would be good if we had some "representative data" in the test dataset. I made the one gloss I created in tstMH be public.

susanodd:
I left the Handshape, Field Choice, Semantic Field, and Derivation History choice tables in the database.
Although other sites may want to change these or add translations, leaving these objects would facilitate sharing data between signbanks, at least for the phonology. Although new gloss ids would be needed. Or some configuration of gloss ids.

susanodd:
[CODE DETAILS RELATED TO PERMISSIONS]

In the `context_processors.py` file:

```
permissions_for_dataset = get_user_perms(request.user, dataset)
print(permissions_for_dataset)
```

results in this kind of permission:

`<QuerySet ['add_permission', 'change_permission', 'delete_permission', 'view_permission']>`

NOT the permissions that look like "change_dataset".

susanodd:
One of the tests that FAILS is the test that creates a gloss. The CRUD test.

I have no idea how to get it to work.
It's not creating a gloss.

However, it is possible to create a gloss in the interface.

I don't know if the error that the test does not work is related to permissions or not.

susanodd:
@Woseseltops 

ALL OF THE TESTS ARE SUCCESSFUL !!!

susanodd:
@vanlummelhuizen Yes, it looks like that.

But some of the code does not run at all unless the kind with `_permission` is used.

On the initial startup, it would not show any datasets at all in the Select Dataset page. I had to change it to the `_permission` one in order for it to show anything.

vanlummelhuizen:
@susanodd I can only imagine that this is coincidental, probably because the current user is a `staff` member or `superuser`. The problem that is does not show anything lies elsewhere.

Permissions with `_permission` in the codename are not a *kind* of permission but rather a normal permission that says what users may or may not do with objects of the **`Permission`** model, and those users are usually staff or superuser.

susanodd:
> @susanodd I can only imagine that this is coincidental, probably because the current user is a `staff` member or `superuser`. The problem that is does not show anything lies elsewhere.
> 
> Permissions with `_permission` in the codename are not a _kind_ of permission but rather a normal permission that says what users may or may not do with objects of the **`Permission`** model, and those users are usually staff or superuser.

I still don't know why the select database page was not working.
Did some "data" or "specific permissions like owner" get thrown out when users were deleted?

I need to figure out if too many things were deleted to make it empty, causing it to not work.
There are some commands that fall back to the default dataset. But NGT is empty now.
Maybe like that other example (the signlanguage query issue) maybe if the dataset is empty some things don't work?

susanodd:
> @susanodd I can only imagine that this is coincidental, probably because the current user is a `staff` member or `superuser`. The problem that is does not show anything lies elsewhere.
> 
> Permissions with `_permission` in the codename are not a _kind_ of permission but rather a normal permission that says what users may or may not do with objects of the **`Permission`** model, and those users are usually staff or superuser.

What about the Manage Dataset page?
Those needed to be changed also in the tests.

susanodd:
> @susanodd I can only imagine that this is coincidental, probably because the current user is a `staff` member or `superuser`. The problem that is does not show anything lies elsewhere.
> 
> Permissions with `_permission` in the codename are not a _kind_ of permission but rather a normal permission that says what users may or may not do with objects of the **`Permission`** model, and those users are usually staff or superuser.

Okay, so I switched to this branch with a "normal" database.
Guess what? The select dataset page shows no datasets.

I'll set the permissions back to what they were.
Hopefully it is something coincidental. Or related to the datasets being empty.

susanodd:
I set the permissions back, although they now match the other issue about the permissions.
(By accident since I was inspecting code to see what permissions there were available to know what could be checked for.)

susanodd:
@Woseseltops since you have experience with a totally different signbank branch, ASL, perhaps you can decide what to do for VGT.

Once they start modifying the settings, it ought to remain possible to pull from the master branch.
They will probably also want to keep the po file up to date. Probably a 'be' po file.
But then that will need to be filled with thousands of translations because they start out empty.
Or they could copy the 'nl' file and edit that to start.
But they will need to keep doing this in order to keep their local signbank up to date.

Also the migrations for the model translations will need to be conditional, and also on master.
We could create a migration to add 'be' be have it be conditional and not add it to global signbank.
Then it would also be on master.

Woseseltops:
> @Woseseltops can you check that there isn't anything signbank-specific to Radboud in the empty database?

I wouldn't really know how to approach that, but I also don't think it would be much of a problem if something 'stays behind' by accident :).

> If you use the new Amazon-tool, can it be with the empty database to check it out?

No, or at least not at the moment: it always 'plugs in' the same static database I have stored on AWS (Amazon cloud) as well.

> since you have experience with a totally different signbank branch, ASL, perhaps you can decide what to do for VGT.

I think this goes for a lot of applications: as long as you don't make changes to the code, and you don't track any changes made to config/translation with git, you should be able to deploy (`git pull`) more recent versions.

--

About reviewing: I'll review once @vanlummelhuizen accepts (so after the permission thing is solved!

susanodd:
> 
> > since you have experience with a totally different signbank branch, ASL, perhaps you can decide what to do for VGT.
> 
> I think this goes for a lot of applications: as long as you don't make changes to the code, and you don't track any changes made to config/translation with git, you should be able to deploy (`git pull`) more recent versions.
> 


Do they know this in Belgium?
If we make conditional model translation migrations and a po file for be, then it would just be in the settings.

vanlummelhuizen:
> About reviewing: I'll review once @vanlummelhuizen accepts (so after the permission thing is solved!

I am a bit confused: there is PR #1282 in which I copy all can_view_data permissions to view_dataset (without can_) permissions and remove the can_view_dataset permission. This is now done in a data migration. As far as I am concerned, this is done. So it can be reviewed and merged. Preferably not by me since I coded it. 

I understand that this PR (#1316) should be merged into master after #1282 is merged (https://github.com/Signbank/Global-signbank/pull/1282#issuecomment-2316917293). But there are overlaps in the code so I predict that it is going to be a challenge, especially because master was merged into #1282. So the current diff between the two PRs is huge.

My assessment of the situation leads to the following procedure:

1. After reviewing, merge #1282 into master + Test, commit, deploy
2. After reviewing, merge #1316 into master
    * Solve merge conflicts (probably numerous, but hopefully not too difficult)
    * Test, commit, deploy

So, first is reviewing #1282. Who is up for that? @Woseseltops @susanodd @Jetske

susanodd:
I'll see about merging master in #1282 

I did it once for dark mode, I think. (Yes, it was time consuming.)

I reduced the permissions differences in #1316 (reverted back so don't use _permission anymore), so hopefully it is more similar now. The main thing was the tests. They all work now.

susanodd:
> > About reviewing: I'll review once @vanlummelhuizen accepts (so after the permission thing is solved!
> 
> I am a bit confused: there is PR #1282 in which I copy all can_view_data permissions to view_dataset (without can_) permissions and remove the can_view_dataset permission. This is now done in a data migration. As far as I am concerned, this is done. So it can be reviewed and merged. Preferably not by me since I coded it.
> 
> I understand that this PR (#1316) should be merged into master after #1282 is merged ([#1282 (comment)](https://github.com/Signbank/Global-signbank/pull/1282#issuecomment-2316917293)). But there are overlaps in the code so I predict that it is going to be a challenge, especially because master was merged into #1282. So the current diff between the two PRs is huge.
> 
> My assessment of the situation leads to the following procedure:
> 
>     1. After reviewing, merge [Api perm check 1281 #1282](https://github.com/Signbank/Global-signbank/pull/1282) into master + Test, commit, deploy
> 
>     2. After reviewing, merge [#1270: Signbank empty database. #1316](https://github.com/Signbank/Global-signbank/pull/1316) into master
>        
>        * Solve merge conflicts (probably numerous, but hopefully not too difficult)
>        * Test, commit, deploy
> 
> 
> So, first is reviewing #1282. Who is up for that? @Woseseltops @susanodd @Jetske

Okay, I did 1) and 2) except did not deploy 2 (this branch) yet.
All of the tests pass on this branch, too.
It looks like the only difference is that I repaired one of the tests. It was commented out on master.

susanodd:
I'm going to empty the tables in a different way.
It's not clear how they are related behind the scenes and what the cascade actually does.
I suspect it's deleting too many things with the command employed by create_test_db (from which the new command is derived.)

susanodd:
Ahah!

This is going wrong:

`save_info_about_deleted_gloss`

Django is trying to save information when deleting things. This also happens in venv commands.

susanodd:
i undid all the things I previously did with the permissions in order to get the tests to work on an empty database.
But now the tests don't work anymore (with the code for permissions matching master.)

Perhaps a table is being removed that is needed?
Or the cascade delete is removing extra things?
I'm having trouble with the datasets.
Also, removing users ends up removing the owner field and the manager field.
It seems like the constraints are failing on am empty dataset/database.

Suggestions?
What dataset should remain in the empty database? There needs to be at least one.
It needs to be public so the site still works for anonymous users.
There needs to be at least one gloss otherwise some queries end up empty and some try except things fail.
This also applies to the setup of the languages and language settings.

I can just leave tstMH and change things so the settings match it. But it needs to be public.

vanlummelhuizen:
> There needs to be at least one.

> There needs to be at least one gloss

Ideally, the code does not fail (or fail gracefully, potayto/potahto) when there is no dataset or only datasets without glosses. This goes for any object for that matter. 

The app could then present the public user with the message that there is no data to view, and present a logged in user with options to fill the database.

That said, for a more quick and dirty solution, you could add a dataset and gloss with names that reflect their default nature, e.g. FIRST-DATASET and FIRST-GLOSS.

susanodd:
> > There needs to be at least one.
> 
> > There needs to be at least one gloss
> 
> Ideally, the code does not fail (or fail gracefully, potayto/potahto) when there is no dataset or only datasets without glosses. This goes for any object for that matter.
> 
> The app could then present the public user with the message that there is no data to view, and present a logged in user with options to fill the database.
> 
> That said, for a more quick and dirty solution, you could add a dataset and gloss with names that reflect their default nature, e.g. FIRST-DATASET and FIRST-GLOSS.

Okay, that last sentence sounds good.

I just made the empty dataset again. First a copy of the newest signbank. Apply migrations. (I copied it before the perms migration.) Then apply the "create empty db" command.
The users and datasets are all there, but no glosses.
no-glosses-no-permissions



Moreover, I do not have any permissions at all for any dataset. (I left the users table. I made a command to delete users, but I'm trying to debug.)
I don't know why I don't have any permissions. Did I delete too many tables?
Or does the vacuum command also delete Django stuff.
Or is it because of cascade delete?

I have no permissions showing in my profile anymore. (But I had them before I made it empty.)

susanodd:
I found the bug!! Here, this is the admin, showing available object permissions for FIRST-DATASET.
Change dataset isn't there to choose.

admin-object-permissions-first-dataset

susanodd:
To create the empty database, make sure the perms migration was applied to a new copy of the master db.
In virtual environment:

1. python bin/develop.py migrate
2. python bin/develop.py create_empty_db
3. cp test_database.db signbank.db
4. python bin/develop.py delete_datasets_not_development
5. python bin/develop.py delete_users_not_development
6. python bin/develop.py create_test_db

susanodd:
@vanlummelhuizen please help.

I've done step 1, 2, 3 on signbank-test

(You can try using it and see that it doesn't work. All glosses have been removed but it won't let me create any.)

I can't do anything at all. I can't create a new sign because it doesn't show the list of datasets to choose.

There are no permissions on Dataset objects.
What went wrong?

This is why I changed all the permissions in the code before because it wouldn't let me do anything.
I'm not going to repeat that because then it can never be used with the master branch.

susanodd:
I modified the `server_specific.py` file to set the DEFAULT dataset acronym to FIRST, to match the new name of tstMH.

However, it looks like the code is using the "base" settings rather than the server specific and keeps using NGT.

https://github.com/Signbank/Global-signbank/blob/6c277dc43495cc3664f71aa9bd8708aa9a86ec8c/signbank/dictionary/views.py#L38-L39

https://github.com/Signbank/Global-signbank/blob/6c277dc43495cc3664f71aa9bd8708aa9a86ec8c/signbank/dictionary/views.py#L485-L486

susanodd:
Using settings.server_specific makes the bug go away.
But it's still not possible to create a gloss because all the dataset permissions (in the choice of permissions in the admin) are gone.
Somehow a table that has permissions has been erased, in spite of not showing up in the log of tables that have been erased.

susanodd:
From bad to worse. I tried to deploy and it said there that migrations needed to be created (False?)
Then I created a migration. (It sets the dataset options to empty.)
Now tried to deploy again and this is the error:
`
django.db.utils.IntegrityError: The row in table 'auth_permission' with primary key '336' has an invalid foreign key: auth_permission.content_type_id contains a value '83' that does not have a corresponding value in django_content_type.id.`

It looks like some tables are being emptied that are not shown in the log of tables that are being emptied.
This is Deja vu.

susanodd:
@Woseseltops @vanlummelhuizen please take over for a while.
I'm going round in circles trying to debug this.

vanlummelhuizen:
> Now tried to deploy again and this is the error:
> ` django.db.utils.IntegrityError: The row in table 'auth_permission' with primary key '336' has an invalid foreign key: auth_permission.content_type_id contains a value '83' that does not have a corresponding value in django_content_type.id.`

Sometimes it is good to dive straight into the database. 

On the test server the content_type with id 83 is gone:

```sql
sqlite> SELECT * FROM auth_permission WHERE id = 336;
id|content_type_id|codename|name
336|83|add_annotatedsentencecontext|Can add annotated sentence context
sqlite> SELECT * FROM django_content_type WHERE id = 83;
sqlite> 
```

Not so on the live server:

```sql
sqlite> SELECT * FROM auth_permission WHERE id = 336;
id|content_type_id|codename|name
336|83|add_annotatedsentencecontext|Can add annotated sentence context
sqlite> SELECT * FROM django_content_type WHERE id = 83;
id|app_label|model
83|dictionary|annotatedsentencecontext
```

There are more things that are messed up on the test server. Permissions for QueryParameter also has 'dataset' in the codename. That seems to be incorrect. Look at the following:

```sql
sqlite> SELECT * FROM auth_permission WHERE codename LIKE '%dataset%';
id|content_type_id|codename|name
148|46|add_dataset|Can add dataset
149|46|change_dataset|Can change dataset
150|46|delete_dataset|Can delete dataset
151|46|view_dataset|Can view dataset
368|46|can_view_dataset|View dataset
493|3|add_dataset|Can add dataset
494|3|change_dataset|Can change dataset
495|3|delete_dataset|Can delete dataset
496|3|view_dataset|Can view dataset
497|3|can_view_dataset|View dataset
sqlite> SELECT * FROM django_content_type WHERE id IN (3, 46);
id|app_label|model
3|dictionary|dataset
46|dictionary|queryparameter
```

No problem on the live server:

```sql
sqlite> SELECT * FROM auth_permission WHERE codename LIKE '%dataset%';
id|content_type_id|codename|name
148|46|add_dataset|Can add dataset
149|46|change_dataset|Can change dataset
150|46|delete_dataset|Can delete dataset
151|46|view_dataset|Can view dataset
368|46|can_view_dataset|View dataset
sqlite> SELECT * FROM django_content_type WHERE id IN (46);
id|app_label|model
46|dictionary|dataset
```

There are more things wrong in the database on the test server:

```sql
sqlite> SELECT * FROM django_content_type WHERE id IN (SELECT DISTINCT content_type_id FROM auth_permission WHERE codename LIKE '%permission%');
id|app_label|model
1|auth|user
4|auth|permission
13|guardian|groupobjectpermission
14|guardian|userobjectpermission
44|dictionary|derivationhistory
45|dictionary|derivationhistorytranslation
```

What is *user* and *derivationhistory(translation)* doing there?

On the live server:

```sql
sqlite> SELECT * FROM django_content_type WHERE id IN (SELECT DISTINCT content_type_id FROM auth_permission WHERE codename LIKE '%permission%');
id|app_label|model
1|auth|permission
44|guardian|userobjectpermission
45|guardian|groupobjectpermission
```

Perhaps it is a good idea to take a fresh copy of the live database and see what happens. (Backup the old DB first, just in case.)

susanodd:
@vanlummelhuizen thanks for investigating.
I did make a fresh copy, using the master database.

The previous database was an empty one (the first empty one) I had made on my local machine. But after the permissions were fixed, I wanted to start with a new one from master. Then that didn't work. I created it on the test server, not locally.

(Perhaps the original one I had created locally kept things because it was older and had been used a lot. Sometimes, like with the model translations, they didn't appear right away in the database, several years ago. Like Django evaluates things lazily.)

I will try again using those commans you used to figure out what is going wrong. It's taking way too much time.
Originally I tried to remove the users in the admin, but that did not work at all. It kept getting consistency violations.
So I moved that to a command, but that also had problems, so now it also deletes the profiles before deleting the users.

Maybe the cascade in the models is causing extra things to be deleted. (When things are run as a command, using python, not sql, it executes the signals and stores stuff about deletes.)

susanodd:
Are some migrations missing from the guardian model?
Those extra permissions that get created when a new model is added?

vanlummelhuizen:
I just ran the *create_empty_db* command on a fresh copy of the live database backup and I can confirm that ContentType with id 83 is is delete by the command while the auth_permission (with id 336) is kept. So, I think `create_empty_db.py` still need some work and testing.

vanlummelhuizen:
First problem: table `django_content_type` is emtpy. The command throws it away because it is not in `keep_tables`.

vanlummelhuizen:
> First problem: table `django_content_type` is emtpy. The command throws it away because it is not in `keep_tables`.

`django_content_type` is called `dictionary_content_type` by mistake in both *create_empty_de* and *create_test_db*. (What is/should be the difference between those commands anyway?)

vanlummelhuizen:
> (What is/should be the difference between those commands anyway?)

The only difference is the list of tables that is to be emptied and a print statement. Could we please make this thing generic and have the list of tables to be emptied as a variable/argument is some way (if that is even necessary)? DRY: Don't Repeat Yourself.

susanodd:
> > (What is/should be the difference between those commands anyway?)
> 
> The only difference is the list of tables that is to be emptied and a print statement. Could we please make this thing generic and have the list of tables to be emptied as a variable/argument is some way (if that is even necessary)? DRY: Don't Repeat Yourself.

This was meant to be a one-off to make an empty database. (This is not something we need to keep doing.)

I just copied the create-test-database to start. The command is dangerous, so it just outputs to the same filename as the test database. (Not the same name as signbank.db)

The create test database erased too many tables. We need to keep users, for example or it's not possible to use the test database in the interface.

susanodd:
@vanlummelhuizen see the issue (#1270). I have no idea where they got the initial database that didn't work. Their pages table was empty, so no menus showed.

susanodd:
> First problem: table `django_content_type` is emtpy. The command throws it away because it is not in `keep_tables`.

But it's not shown in the "print" statement log of things that are thrown away either.
I ran the command many times without deleting anything in order to see what it wanted to delete compared to what is needed for it to run.

vanlummelhuizen:
Well if I run _create_empty_db_ it does say so:

```
❯ bin/develop.py create_empty_db
Making backup of old test database
Copying database file
Emptying tables, for faster tests
...
Delete contents from table:  django_content_type
...
```

susanodd:
Strange.

susanodd:
For this one:

```
sqlite> SELECT * FROM auth_permission WHERE id = 336;
id|content_type_id|codename|name
336|83|add_annotatedsentencecontext|Can add annotated sentence context
sqlite> SELECT * FROM django_content_type WHERE id = 83;
id|app_label|model
83|dictionary|annotatedsentencecontext
```

I haven't been able to test the command that uses this because I don't have any test data.
This class was added very recently to Signbank.
There are no tests for annotated sentences.
There are no videos or eaf files on the servers except for on master signbank.

susanodd:
I see you are working on this on a new branch, @vanlummelhuizen

vanlummelhuizen:
As an attempt to approach this problem from the other end, I tried to create an install script that prepared the database from scratch. The result can be found in branch [make_installable](https://github.com/Signbank/Global-signbank/tree/make_installable).

For `server_specific.py` I used the code below. You probably need to change ROOT, BASE_DIR and SECRET_KEY. Also, you need to create the *writable* directory structure first. You don't have to create a SQLite database file; that is done from the script [INSTALL.sh](https://github.com/Signbank/Global-signbank/blob/make_installable/INSTALL.sh).

@susanodd @Woseseltops @Jetske , could you try it out?

```python
from signbank.settings.server_specific.default import *

ROOT = '/home/micha/docs/ru/signbank/global-signbank/'
BASE_DIR = 'repo-global-signbank/'

ADMINS = [('admin', 'admin@example.com')]
LANGUAGE_NAME = 'Global'
COUNTRY_NAME = 'Netherlands'
SIGNBANK_VERSION_CODE = 'global'
URL = 'http://localhost'
ALLOWED_HOSTS=["127.0.0.1","localhost"]

EAF_FILES_LOCATION = 'corpus-ngt/eaf/'
METADATA_LOCATION = 'CNGT_MetadataEnglish_OtherResearchers.csv'


LANGUAGES = (
  ('en', gettext('English')),
  ('nl', gettext('Dutch')),
  ('zh-hans', gettext('Chinese')),
  # ('he', gettext('Hebrew')),
  # ('ar', gettext('Arabic')),
)

LANGUAGES_LANGUAGE_CODE_3CHAR = (
    ('en', 'eng'),
    ('nl', 'nld'),
    ('zh-hans', 'zho'),
    # ('he', 'heb'),
    # ('ar', 'ara'),
)

INTERFACE_LANGUAGE_SHORT_NAMES = ['EN','NL','官话','עִברִית','عربى']
MODELTRANSLATION_LANGUAGES = ['en','nl','zh-hans',
                              # 'he', 'ar'
                                ]

WRITABLE_FOLDER = 'writable/'

#Database settings
DATABASES = {'default':
                {
                    'ENGINE': 'django.db.backends.sqlite3',
                    'NAME': 'database/signbank.db',
                    'TEST': {
                        'NAME': 'database/test-signbank.db',
                    }
                }
            }

USE_REGULAR_EXPRESSIONS = True

ECV_FILE = ECV_FOLDER+'ngt.ecv'
ECV_SETTINGS = {
    'include_phonology_and_frequencies': True,

    'description_fields': ['handedness', 'domhndsh', 'subhndsh', 'handCh', 'locprim', 'relOriMov', 'movDir', 'movSh',
                           'tokNo', 'tokNoSgnr'],

    # The order of languages matters as the first will
    # be treated as default by ELAN
    'languages': [
        {
            'id': 'nld',
            'description': 'De glossen-CV voor het CNGT (RU)',
            'annotation_idgloss_fieldname': 'annotationidglosstranslation_nl',
            'attributes': {
                'LANG_DEF': 'http://cdb.iso.org/lg/CDB-00138580-001',
                'LANG_ID': 'nld',
                'LANG_LABEL': 'Dutch (nld)'
            }
        },
        {
            'id': 'eng',
            'description': 'The glosses CV for the CNGT (RU)',
            'annotation_idgloss_fieldname': 'annotationidglosstranslation_en',
            'attributes': {
                'LANG_DEF': 'http://cdb.iso.org/lg/CDB-00138502-001',
                'LANG_ID': 'eng',
                'LANG_LABEL': 'English (eng)'
            }
        },
    ]
}

#Settings for the default dataset
DEFAULT_DATASET = 'Your Dataset'
DEFAULT_DATASET_ACRONYM = 'YDS'
DEFAULT_DATASET_LANGUAGE_ID = 1
DEFAULT_DATASET_PK = 1
TEST_DATASET_ACRONYM = 'TESTDB'

FREQUENCY_CATEGORIES = ['Occurences', 'Signers']

FREQUENCY_REGIONS = ['Amsterdam', 'Voorburg', 'Rotterdam', 'Gestel', 'Groningen', 'Other']

TMP_DIR = '../writable/tmp'
FILE_UPLOAD_MAX_MEMORY_SIZE = 5242880 # 5 MB

#temp experiment
SWITCH_TO_MYSQL = False 

if SWITCH_TO_MYSQL:
    DATABASES = {'default': {
        'ENGINE': 'django.db.backends.mysql',
        'OPTIONS': 'mysql.conf',
        'TEST': {'NAME':'sigbank_test'}
    }}
SECRET_KEY="change_me"
CSRF_TRUSTED_ORIGINS = ["https://127.0.0.1","https://localhost"]
```

susanodd:
Which of the development servers are you using @vanlummelhuizen ? signbank-dev or signbank-test?

I prefer not to do this on my local computers because it messes up the normal signbank. (Because of PyCharm runserver, it will get screwed up with different settings.)

susanodd:
Can probably drop the frequency regions. (Or what do we do? Those are tied to the CNGT corpus.)

susanodd:
@vanlummelhuizen the Language model itself is sorted by '-id' (because English has a smaller id)
This is because sorting by 'name' ends up showing the languages in a different order depending on the interface language.
I see above something is also done for ELAN. UvA always wants Dutch (Nederlands) first independent of the interface.
(This won't work for other installations if they prefer something else.)

vanlummelhuizen:
> Which of the development servers are you using @vanlummelhuizen ? signbank-dev or signbank-test?

I do this locally. You could `git clone` the repository again so that it does not interfere with existing local repo's. 

> What we need to make sure is that "whatever" is created is that it still works with the master branch and the tests still succeed. Can you put the created database onto either signbank-dev or signbank-test and change the server specific as shown but not mess up anything related to Apache? (Don't undo what ansible did?)

This branch is branched from master. Good point about the tests. I will look into that. 

I would rather see whether this is a feasible way to go before putting it on any (test or dev) server.

> Can probably drop the frequency regions. (Or what do we do? Those are tied to the CNGT corpus.) For VGT they don't want Chinese. They actually didn't seem to want English either, but it won't work without.

I left all Global-Signbank languages in for now to have a working database. Next step would be to see whether it can be done with more/less/other languages.

susanodd:
This one ought to be done using the new Amazon temporary server pull request setup of @Woseseltops.

susanodd:
I added more tables to the KEEP list as discovered by @vanlummelhuizen. (For some reason this did not affect the tests on the normal database, but they need to be explicitly in the KEEP list.)

Then performed the steps above. (https://github.com/Signbank/Global-signbank/pull/1316#issuecomment-2326018944)

All of the tests work now.
Only the dataset tstMH remains and only the developer users remain, plus Divya.

The empty database is on signbank-test.
The name of the dataset can be changed to something generic. I changed the server specific to have dataset 1 tstMH be the default.

@Woseseltops communicate however you did this before with the empty database.

I like the method of @vanlummelhuizen much better of course. But I'm not able to develop that myself, I think it should be tested on the Amazon temporary servers.

vanlummelhuizen:
> Or onto your tensusers account so that we can copy it elsewhere?

It is here: /vol/tensusers/mhulsbosch/signbank/signbank.db

> I browsed the json files, I suspect that the pages data is going to need the actual "contents" fields for the different languages, albeit empty/single space character.

I don't think so. The handling of pages not in a urls.py somewhere is done by PageFallbackMiddleware in https://github.com/Signbank/Global-signbank/blob/master/signbank/pages/middleware.py: when the normal flow results in a 404 it tries to find the page in Pages.

> Is it wise to edit the early migrations? The migrations can no longer be applied again retroactively.
> (Will this cause any problems with the master database if the migrations that it has applied in the past have been modified?)

This is a good question. I assumed editing early migrations is possible. These migrations are already in the Global-Signbank instance so they will not be applied again there. Also, I tried to make sure the end result is not different from the database we already have in terms of structure/schema. If you find anything in my edits that makes you think otherwise, please let me know.

> The models Handshape, SemanticField, DerivationHistory are all objects and no longer field choices. They will need json files too. And the field choices for them removed. (Otherwise there will be lots of type errors because they need to be objects of these models.)

Good point. I will look into that.

susanodd:
@vanlummelhuizen I just put your signbank.db (empty as on tensusers) onto signbank-dev.

I revised the settings in the server_specific.py file (for the dev server only) as shown above for the dataset settings.

I'm not sure how I can login now.
It's on signbank-dev

susanodd:
Okay, quite a lot of tests fail. (!!)

susanodd:
@vanlummelhuizen could you have a look at the settings files?
(Slightly related to this, I'm adding the basic structure for the animation issue. I've encountered some imports of settings and I'm not sure whether they are correct. For example,

https://github.com/Signbank/Global-signbank/blob/a18988c173e876bc22fd9c96cc19a50575104e4e/signbank/video/models.py#L17

Those things aren't defined in base.

Woseseltops:
Summary of a conversation with @susanodd :
* This functionality is mainly for new Signbank installations by new teams, so the approach where you don't need an existing database will be more useful in most cases
* Various tests (and also functionality?) require at least one dataset, so it would convenient if this tool/command created that

Optional functionality: also have this tool/command create a page with copyright info

vanlummelhuizen:
> * This functionality is mainly for new Signbank installations by new teams, so the approach where you don't need an existing database will be more useful in most cases

Do you mean what is done in [make_installable](https://github.com/Signbank/Global-signbank/tree/make_installable)?

> * Various tests (and also functionality?) require at least one dataset, so it would convenient if this tool/command created that

When creating a database from scratch as in [make_installable](https://github.com/Signbank/Global-signbank/tree/make_installable), a default dataset is created: "Your dataset", YDS. See https://github.com/Signbank/Global-signbank/blob/master/signbank/dictionary/migrations/0027_auto_20180803_1340.py

But it would be best if the tests are creating a test-dataset themselves, independent of the contents of the (test-)database.

vanlummelhuizen:
> Okay, quite a lot of tests fail. (!!)

I did some testing and one problem was a typo in the DEFAULT_DATASET setting of the server_specific.py file I gave earlier. It should be:

```python
DEFAULT_DATASET = 'Your dataset'
```

(lower 'd' of dataset)

susanodd:
Yes, some of the older tests use NGT. (Kind of dangerous.) Some do create a new dataset. I agree that this needs to be for all. I an work on that.
Good catch with the capital d typo!

susanodd:
> To solve quite some tests I add `language_code_2char` values to migration 0053, see [103da9c](https://github.com/Signbank/Global-signbank/commit/103da9c98ab9b2a45f737568a495b5751145bd97)

That will also solve many problems in the templates, since all the little annotation dictionaries map the 2-char to a string.

vanlummelhuizen:
There is one test that still fails: https://github.com/Signbank/Global-signbank/blob/103da9c98ab9b2a45f737568a495b5751145bd97/signbank/dictionary/tests.py#L623

I think this is because the test expects something that will not happen: the test sends a form (stage 2) with changes that were *pretended to be* extracted from a CSV in the stage before (stage 1). However, the thing that is tested (https://github.com/Signbank/Global-signbank/blob/103da9c98ab9b2a45f737568a495b5751145bd97/signbank/dictionary/tests.py#L728) will never happen because in `views.py` the text 'No changes were found.' is only used in stage 1, not stage 2. See https://github.com/Signbank/Global-signbank/blob/103da9c98ab9b2a45f737568a495b5751145bd97/signbank/dictionary/views.py#L1499-L1502 and https://github.com/Signbank/Global-signbank/blob/103da9c98ab9b2a45f737568a495b5751145bd97/signbank/dictionary/views.py#L1489

@susanodd Do you agree and if yes, do you have an idea to solve this?

susanodd:
@vanlummelhuizen @Woseseltops I copied the empty database I had created onto my tensusers account.

I needed to put the animation code on signbank-test because the CORS isn't set up properly on signbank-susan and there is no ajax lookahead. (Some bug it thinks that it is on a different site.)

susanodd:
@vanlummelhuizen that's weird about the test. I will need to check it out. For some reason it works on the normal databases.
Perhaps I modified something in the code that didn't trickle down.
I haven't modified that code in quite a while. But some similar constraints checks for lemmas got implemented for the API, so it could be that I modified ("improved") some code but didn't check it.
I also added some prototype spreadsheet rows to the template, so I could have messed up something.

susanodd:
The test is testing whether there is one translation language. So probably something else is going on.
I don't think it gets as far as stage 2.

https://github.com/Signbank/Global-signbank/blob/a18988c173e876bc22fd9c96cc19a50575104e4e/signbank/dictionary/tests.py#L698-L704


https://github.com/Signbank/Global-signbank/blob/a18988c173e876bc22fd9c96cc19a50575104e4e/signbank/dictionary/tests.py#L723-L728

Yes, something needs to be modified for the tests because it assumes the dataset is NGT and has two languages. (That's not generic enough.)


[BLAH BLAH]
Sometimes is is necessary in the tests to have Follow being True or else False in order for it to "complete" whatever it is doing in the template, or not. I think this is also to do with when the messages are displayed "inside the test" since it's not the same as a real template. 

It could even be that the template area for messages has become hidden or behind other stuff. I could have modified something for dark mode, or else for showing the scroll bar and not noticed. (Those are things that can hide the messages, also the "affix".)

vanlummelhuizen:
I made changes so that creating translation fields is done dynamically: https://github.com/Signbank/Global-signbank/commit/18a97e24c1a00bb5cfa4a10a4e5bd10f8b51325f so that it also works with the following language settings

```python
LANGUAGES = (
  ('en', gettext('English')),
  ('nl', gettext('Dutch')),
  ('zh-hans', gettext('Chinese')),
  ('he', gettext('Hebrew')),
  ('ar', gettext('Arabic')),
)

LANGUAGES_LANGUAGE_CODE_3CHAR = (
    ('en', 'eng'),
    ('nl', 'nld'),
    ('zh-hans', 'zho'),
    ('he', 'heb'),
    ('ar', 'ara'),
)

INTERFACE_LANGUAGE_SHORT_NAMES = ['EN','NL','官话','עִברִית','عربى']
MODELTRANSLATION_LANGUAGES = ['en','nl','zh-hans',
                              'he', 'ar'
                                ]
```

susanodd:
On a normal test database, even including the migrations for animations, this is what that test prints:

`More than one translation language, attempt to update a lemma translation`

That is the first branch of the code. Not the branch that checks for No changes found.
(So something is different in the empty database.)

susanodd:
@vanlummelhuizen Okay, remove the "else" clause of that test, but keep the "if" clause.
```

        response = client.post(reverse_lazy('import_csv_update'), form_data, follow=True)
        if count_dataset_translation_languages > 1:
            print('More than one translation language, attempt to update a lemma translation')
            self.assertContains(response, 'Attempt to update Lemma translations')
```

Remove the "else" where it looks for "No changes found."
But that first test should be being tested. That's what happens on the non-empty database.

You're right, even if there is only one translation language (I tried ASL), it won't work for that. I suspect that is a remnant of old code that did additional tests on the input file. Once upon a time, it kept checking the dataset column again and again because it was allowed to modify multiple datasets.

vanlummelhuizen:
> Can you put the newest empty database on your tensusers again?

Done.

> Is this set?
> 
> `language_name = getattr(language, settings.DEFAULT_LANGUAGE_HEADER_COLUMN['English'])`

Yes, it is taken from `signbank/settings/server_specific/default.py`.

---

> Remove the "else" where it looks for "No changes found."

I did the following:

```diff
diff --git a/signbank/dictionary/tests.py b/signbank/dictionary/tests.py
index e84517d7..64de0252 100644
--- a/signbank/dictionary/tests.py
+++ b/signbank/dictionary/tests.py
@@ -699,9 +699,6 @@ class ImportExportTests(TestCase):
         if count_dataset_translation_languages > 1:
             print('More than one translation language, attempt to update a lemma translation')
             self.assertContains(response, 'Attempt to update Lemma translations')
-        else:
-            print('Only one translation language, no changes found')
-            self.assertContains(response, 'No changes were found.')
 
 
         # Prepare form data for linking to SEVERAL EXISTING LemmaIdgloss + LemmaIdglossTranslations
@@ -723,9 +720,6 @@ class ImportExportTests(TestCase):
         if count_dataset_translation_languages > 1:
             print('More than one translation language, attempt to update a lemma translation')
             self.assertContains(response, 'Attempt to update Lemma translations')
-        else:
-            print('Only one translation language, no changes found')
-            self.assertContains(response, 'No changes were found.')
 
 
     def test_Import_csv_new_gloss_for_lemma(self):
```

and now there is no failure. But then again, I just removed the failing test.

susanodd:
@vanlummelhuizen it fails because it thinks the number of translation languages is not 2.
The normal database does the branch that is still there. You can see in the log for a normal database that it does that branch.
Does the test dataset have two translation languages? They are added in the code with "add". Does the make install do it via the fixtures?

susanodd:
@Woseseltops see here

https://github.com/Signbank/Global-signbank/issues/1270#issuecomment-2464617692

PR #1318, opened 2024-08-26
Annotated Sentence List View

Jetske:
@susanodd can you review this and if you want build into your annotated gloss page as switch (it has a url: /annotatedsentences/show_all)? It is not anywhere in the menu at the moment so it's like a hidden page. 

It's also live on signbank-jetske


susanodd:
@Jetske related to this, I am not able to see any annotated sentence videos on my local server.
I copied the files 1.mp4 and 1.eaf to my own computer
writable/annotatedvideo/tstMH/1.mp4
writable/annotatedvideo/tstMH/1.eaf

This is using a database from 19.08.24

In spite of having the files, the Gloss Detail refuses to show the video for the sentence.
It does print the correct path in the python code but it won't show in the template.

Is there some video javascript that doesn't work on Ubuntu? Am I missing a file?
Are the files in the proper location?
Has the video data been changed since that date?

Jetske:
@susanodd The videos should probably be one more folder down in writable/annotatedvideo/tstMH/1/1.mp4

susanodd:
@Jetske we need an additional URL to enable switching to the Edit for annotated sentence.
The Gloss ID appears in the URL.

susanodd:
@Jetske I can edit the Edit Annotated Sentence template to make the redirect url and the glossid be optional.
The template can make use of the dataset instead of the gloss.
To do this, the reference to 
`url 'dictionary:edit_annotated_sentence'`

will need to be an ajax call instead so parameters can be passed that are not inside the url.

susanodd:
@Jetske I've got the Annotated Gloss List view nearly done. I just figured out how to get the sort working that works with the regroup.

I will merge yours into master first, since it will be easier to link between them if your code exists first. I see you have modified the Edit view, which will make it easier as well using the redirect parameter. At the moment, it's not possible to save an edited sentence because the redirect goes to a gloss detail page, not the sentence detail page where the button was.
So the redirect will fix that.

Jetske:
@susanodd Sounds good! Do you still need answers to your questions? I think I'd prefer option 2, to make it a hidden parameter. 
I think it leaves edit mode when returning to the gloss detail page, but I'm not sure.

Issue #1311, opened 2024-08-09
Increase allowed video upload size


Jetske:
Increase it from 5mb to 10mb.

susanodd:
@Jetske this should not be done in file 

https://github.com/Signbank/Global-signbank/commit/8b9248c3d80d0edfa22d0651b7043a3f77de9176#diff-0a3b59552c5f10906d3787110652f40ee700668282f92e358e0dfcc18633506a

It needs to be done in server_specific / server_specific.py
(on the server)
Otherwise it can't be merged and git overwrites the server specific secret key.

I fixed in on signbank, but beware when doing this on your own server.
Can this be in default instead?
Because we can't publish the secret key, the changes to this file are not actually merged into the server.

PR #1324, opened 2024-09-12
Sentence video upload size increased #1311

Jetske:


Jetske:
@susanodd not relevant for upload size, but anyway I think permissions for annotated sentences need to be added for group 'editor'. Right?

susanodd:
> @susanodd not relevant for upload size, but anyway I think permissions for annotated sentences need to be added for group 'editor'. Right?

I added them (signbank). I don't think it matters unless they are explicitly checked in the code.
For example, none of the permissions for AnnotationIdGoss don't have been added for Editor.

Woseseltops (approved):

Issue #1307, opened 2024-08-07
Possibility to show videos from different perspectives


uklomp:
Hi @Jetske and @Woseseltops 
We talked about the possibility to upload multiple videos for signs and annotated sentences and to show this sign/sentence from multiple perspectives. We record the signs, and a new batch of annotated sentences (that are not from the corpus), with three cameras: one from the middle, like we upload now, one from the front-left and one from the front-right. We would like the following things:
1. A symbol in the video frame that users can click to see one of the other perspectives.
2. You asked before if this functionality of multiple perspectives would only be for single signs or also for annotated sentences --> we indeed also want this possibility for sentences, if it's not too difficult.
--> Related to 2: If it's indeed possible for annotated sentences, the functionality regarding linked glosses should still work the same as it's working now, also when a video with a different perspective is shown. 
3. We discussed what to do with search results, because then a small video from the front is shown: this will stay the same, we don't need the other perspectives here.
4. You asked what to do with deleting, moving, overwriting and you suggested to build two options: upload three videos at the same time, and upload/change only the middle video. We agree with this suggestion. In the second option, potential older videos from the left and right perspective remain, while the middle video is updated.

Woseseltops:
To make this task a bit easier to digest, let's split it up into separate phases:
1. Upload 'other perspective video' to a gloss
2. Display 'other perspective videos', and make it possible to switch between them
3. Delete/overwrite extra videos
4. The relation to annotated sentences

Jetske:
How would you prefer to upload the 3 files? 
- As a zip file that contains 1-3 videos given some name that distinguishes which video is which perspective?
- One by one in three file upload boxes?

susanodd:
Just a request, can this be on a different page rather than on Gloss Detail?

(Gloss Detail has some problems with too much code.)

uklomp:
> How would you prefer to upload the 3 files?
> 
> * As a zip file that contains 1-3 videos given some name that distinguishes which video is which perspective?
> * One by one in three file upload boxes?

I would say the first option, but it might be that @rem0g has different wishes for uploading things automatically?

susanodd:
> > Just a request, can this be on a different page rather than on Gloss Detail?
> > (Gloss Detail has some problems with too much code.)
> 
> Preferably not, especially not for the sentences. For the single gloss videos, I can imagine a pop-out window for the videos, if that would make things better. Going to a completely different page is not really an option.

[...]

PR #1328, opened 2024-09-23
#1307: Model, upload, videos from different perspectives

susanodd:
Coding issues:

1. to upload a file, it needs to be visible to the user and there needs to be an area to upload it
2. the upload file form is "hidden" during normal viewing
3. the perspective files trade places with the main video using the buttons surrounding the video area

[NOT THIS ISSUE BUT CROWDED CODE GLOSS VIEW] The "record" code does not work on Safari nor on Ubuntu, but the html code still appears in the (hidden until in edit mode) browser underneath the normal video (the perspective video code appears below this drag and drop since it is for different files)

The following "now" suffix was removed because this doesn't work in Safari.

`src="{{protected_media_url}}{{gloss.get_video_url}}?v={% now 'YmdHis' %}`


susanodd:
CODE

Because of the above (see issue) I did the following to the Gloss Detail

I removed the "reload videos" code.

I keep the Perspective videos panel (now on the left) visible all the time. Also when not in Edit mode.
(I want to keep the videos loaded so they don't disappear when elements disappear.)
The panel can be opened and closed.
Hopefully it will work to "show" them -- they are already loaded -- in the same place as the normal video.

susanodd:
Next, the "source" videos need to "swap out" the normal video. But they should not be reloaded again.

susanodd:
And after doing a bunch of stuff in a different browser window, the perspective videos have disappeared again from the file system.

susanodd:
Why does the code work when the panel is in the right column, but not when it's in the left column?

susanodd:
I forgot to put in the commit message, I made it so if a perspective video already exists for left or right, the file is replaced with a new upload, rather than deleting the old object and creating a new. It makes use of shutils to move the new temp file to the same location as the already existing.

Woseseltops:
Hey @susanodd , this is an open pull request and my review is requested, but the comments suggest this feature is still being debugged. To avoid confusion, and reviewing code that will later be deleted, would you agree to only create pull requests for features (or if it's really large, parts of features) that are ready from your point of view? Thanks :)

No need to delete this one, but I'll make it a draft. Feel free to 'undraft' whenever you feel it's ready!

susanodd:
https://signbank-susan.cls.ru.nl/dictionary/gloss/36430/

That's the gloss with looping. The play and pause don't work yet.

susanodd:
> Hey @susanodd , this is an open pull request and my review is requested, but the comments suggest this feature is still being debugged. To avoid confusion, and reviewing code that will later be deleted, would you agree to only create pull requests for features (or if it's really large, parts of features) that are ready from your point of view? Thanks :)
> 
> No need to delete this one, but I'll make it a draft. Feel free to 'undraft' whenever you feel it's ready!

I was hoping for some feedback while coding. I didn't delete anything, just repaired code that was broken (elsewhere). Some bugs with video files and except clauses.

I thought helps others to see the code on a server.
The gloss has the perspective videos showing now. (But playing them and switching does not work yet. Probably the controls are still needed on each. It doesn't respond to "play" or "pause".)

susanodd:
This is more or less complete, except files keep disappearing on PyCharm runserver.

Woseseltops:
Converting this to a draft again, since more changes are coming after discussion with @uklomp

susanodd:
> Converting this to a draft again, since more changes are coming after discussion with @uklomp

@Woseseltops the trick to putting the buttons in front does not work for the buttons above the video. (It moves the videos itself or it hides the buttons. The videos end up cropped vertically.)

susanodd:
@Woseseltops THIS DOES NOT WORK ON SAFARI
The buttons show up behind the video, not on top.

susanodd:
@Woseseltops with ordering the buttons, the perspectives are now hard-coded.

Woseseltops (commented):
Hey @susanodd , code looks good in my opinion, except in `GlossVideoPerspective` I see some situations where I think guard clauses would be easier to read (https://en.wikipedia.org/wiki/Guard_(computer_science))

susanodd:
> Hey @susanodd , code looks good in my opinion, except in `GlossVideoPerspective` I see some situations where I think guard clauses would be easier to read (https://en.wikipedia.org/wiki/Guard_(computer_science))

That would be code that originated as a copy from other code. :(

susanodd:
> Hey @susanodd , code looks good in my opinion, except in `GlossVideoPerspective` I see some situations where I think guard clauses would be easier to read (https://en.wikipedia.org/wiki/Guard_(computer_science))

This is kind of annoying because there are numerous methods that do it backwards. Many of the really old methods. The signal methods. I was merely copying the same style and filling in on the signal methods.
Yes, I understand why, mainly because of recursion, so there is nothing left to do when a called method finishes.
That there is code left to do could be a source of some of the python methods that call methods in other files and then need to save stuff to the database afterwards. The atomic block branches over multiple python files.
I keep fixing things when I find them. But since we ignore error reporting this most of the them doesn't help.

I'm fixing lots of these now for perspective videos, but then I also end up editing other code that is not part of this issue.

susanodd:
@Woseseltops can you take a look at the changes in the most recent push. I modified the (perspective) video code to repair the guards and added some error reporting related to the file system. Actually doing anything with the errors needs to be done elsewhere. (See issue about improved error reporting.)

Issue #1330, opened 2024-09-24
Error in changing gloss


uklomp:
When editing a gloss on Signbank, I click 'edit' and write the new gloss, and save it. It often does not save in one time, I often need multiple attempts. Usually after 2 or 3 times, or refreshing the page, it saves, but today it doesn't seem to work - at least not for the English gloss for REPAREREN-A (in Dutch). I get this error: 

<img width="660" alt="image" src="https://github.com/user-attachments/assets/f3b74a2c-8da4-410b-af66-7077ca203310">

susanodd:
Is this for changing the Annotation or the Lemma ?

susanodd:
The permissions have been revised, but that was several weeks ago.

We will need to look at the logs.

susanodd:
@uklomp what do you want the English to be? It's FIX-B right now.
Is that what it's supposed to be?

![ngt-repareren-a-fix-b](https://github.com/user-attachments/assets/af07f138-597d-4c32-87d7-fa9e9088531e)

susanodd:
From the log:
```

Internal Server Error: /dictionary/update/gloss/46332
django.core.exceptions.ValidationError: ["The annotation idgloss translation text 'REPAIR-A' is not unique within dataset 'NGT' for gloss '46332'."]
```

susanodd:
I've got it fixed (locally):


![error-message-annotation](https://github.com/user-attachments/assets/b958131a-c5b9-4683-955f-91701b392f1f)

susanodd:
It's going to be stuck in English for a while.

susanodd:
This is live on signbank.

uklomp:
Ok I get your first questions: Yes, it was about the English annotation ID, and it didn't allow me to save. The other messages I don't quite get but I think you wanted me to try it again. 
So what I wanted to do is change FIX-B to REPAIR-A and then it pretended to be saving, but then the field became empty, so no English annotation gloss at all anymore?? And then I refreshed and the old text FIX-B is back...

susanodd:
> Ok I get your first questions: Yes, it was about the English annotation ID, and it didn't allow me to save. The other messages I don't quite get but I think you wanted me to try it again.
> So what I wanted to do is change FIX-B to REPAIR-A and then it pretended to be saving, but then the field became empty, so no English annotation gloss at all anymore?? And then I refreshed and the old text FIX-B is back...

[...]

PR #1334, opened 2024-10-02
#1330: Report constraint error as alert to user, not as message

susanodd:
Extended feedback message to include id of conflicting gloss.

This branch is not on any server.


susanodd:
@vanlummelhuizen I see in the update_gloss method of update.py there are some already existing "returns" that are inconsistent.
Most are HttpResponse but delete gloss has HttpResponseRedirect. It's not possible to return to Gloss Edit if the gloss has been deleted.

Woseseltops (approved):
Looks good!

I do prefer f-strings though :) . Instead of 

```python
 msg = "The annotation idgloss translation text '%s' is not unique within dataset '%s' for gloss '%s'. Gloss %s also has this text." % (self.text, dataset.acronym, self.gloss.id, gloss_with_same_text.id)
```

you get

```python
msg = f"The annotation idgloss translation text '{self.text}' is not unique within dataset '{dataset.acronym}' for gloss '{self.gloss.id}'. Gloss {gloss_with_same_text.id} also has this text."
```

susanodd:
> Looks good!
> 
> I do prefer f-strings though :) . Instead of
> 
> ```python
>  msg = "The annotation idgloss translation text '%s' is not unique within dataset '%s' for gloss '%s'. Gloss %s also has this text." % (self.text, dataset.acronym, self.gloss.id, gloss_with_same_text.id)
> ```
> 
> you get
> 
> ```python
> msg = f"The annotation idgloss translation text '{self.text}' is not unique within dataset '{dataset.acronym}' for gloss '{self.gloss.id}'. Gloss {gloss_with_same_text.id} also has this text."
> ```

This is to do with already existing translations of the first sentence. I added a second sentence.
The text as a whole is poorly written because it's two sentence now.
But then all the translations will need to be done with some un-handy Python patterns in the translations.
(Those are difficult to translate because the syntax isn't the same in all languages.)

susanodd:
I modifed it as you show.

We were supposed to make it be a link to the other gloss.
But I think that is not wise because the control flow should be to "close" the alert (using browser button in "alert") not to jump out of it to a completely different page.

Issue #1331, opened 2024-09-25
Improve error reporting when an operation fails


susanodd:
Instead of just try-except pass around error messages to report back. Also for transactions that fail. Also for file system errors.
This is non-trivial.

Kindly report use cases (including API) where something is going wrong but this isn't reported sufficiently.

susanodd:
@rem0g are you busy trying to upload a zip file?

In the error log it was giving errors about the api token.

rem0g:
Yes, sometimes video_upload returns whole django HTML template so my script doesnt know how to interprent that. It will keep for looping that same video file, so error reporting should be always in JSON format.

Sometimes i'm getting Failed to upload message, though there is no reason attached as why it happens. 

I will continue to run the script as some videos are uploaded succesfully, but I am not satisfied with how Signbank handles the video upload process so we will need another method to handle the videos as in the future we will have to process thousands of videos every week (happening now). 

I am getting fails with gloss with special chars for example: 
"DISNEY+-A"
"#WDD"

Also Signbank notifies the video is not h264 but yet the video is there on Signbank. 

@Woseseltops i think the video upload process needs a do-over; 
A seperate python file to retrieve video file via POST and process as is, it's not needed to check whether it's h264. If it doesnt play, it wont play and we will know that  one way or another.
JSON file so i can check every moment if the upload and placing the video has been successful.

Jetske:
@rem0g what is a good way to report errors? Returning a JsonResponse with the error? I'm trying to add checks into the APIs for example sentences.

Woseseltops:
Indeed @Jetske , something like this:

```python
return JsonResponse({'error': 'Invalid request'}, status=400)
```

In addition to that, using the correct error code would be even better:
* 400 Bad Request: The request was malformed.
* 401 Unauthorized: Authentication is needed.
* 402 Payment Required: Reserved for future use.
* 403 Forbidden: The server understood the request but refuses to authorize it.
* 404 Not Found: The requested resource could not be found.
* 405 Method Not Allowed: The request method is not supported for the resource.

@rem0g , I'll create a separate issue for your video upload suggestions!

susanodd:
The error reporting I added for gloss updates reports type errors, for example choices that don't exist.

susanodd:
@rem0g have you been able to use the NME API functions all right?

susanodd:
> Indeed @Jetske , something like this:
> 
> ```python
[...]

PR #1335, opened 2024-10-02
#1331: Improve error reporting (phase 1)

susanodd:
in error message feedback

- gloss not in selected datasets => reports dataset gloss is in
- import zipped video archive => does not check h264
- mport zipped video archive => same dictionary for branches in json result
- mport zipped video archive => if video file overwrite existing fails, report error and do not return video url (since not uploaded)

### DIFF
--- a/signbank/api_interface.py
+++ b/signbank/api_interface.py
@@ -187,10 +187,7 @@ def check_gloss_existence_for_uploaded_video(dataset):
                     gloss = Gloss.objects.filter(lemma__dataset=dataset, archived=False,
                                                  annotationidglosstranslation__language__language_code_3char=language3char,
                                                  annotationidglosstranslation__text__exact=filename_without_extension).first()
-                    if gloss:
-                        list_of_video_gloss_status[language3char].append((file, True, gloss))
-                    else:
-                        list_of_video_gloss_status[language3char].append((file, False, gloss))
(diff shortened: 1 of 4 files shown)


susanodd:
This issue is marked as a draft because it is potentially "endless".
But the video upload errors are high priority.

Woseseltops:
Thank you for providing a bite sized pull request AND a list of bullet points with the content, much appreciated :D 

I have nothing to complain code wise, but as for the functionality: I notice there are a few `print()` statements with extra info... is there are a reason why we don't return this to the end user making the requests, instead of just storing it in our own log?

susanodd:
> Thank you for providing a bite sized pull request AND a list of bullet points with the content, much appreciated :D
> 
> I have nothing to complain code wise, but as for the functionality: I notice there are a few `print()` statements with extra info... is there are a reason why we don't return this to the end user making the requests, instead of just storing it in our own log?

The print statements are when one of the temporary unzipped files can not be deleted.
That used to be just ignored. It does not harm, it's just they reside in the import_videos/ folder.
I don't think the user needs this information because nothing went wrong. The import just couldn't clean up after itself.
It would be if the permissions are not set properly on the folder.
Actually some of the tests print messages that the videos could not be deleted.
In debugging the code, I discovered that in the Admin for GlossVideo it's not possible to delete an object unless the file has already been removed.

I don't know whether to leave these print statements or not. It's only useful for development.

Issue #1203, opened 2024-03-23
Extend API fields for multiple videos and example sentences


rem0g:
API should have extended fields for example sentences and accepting multiple videos per gloss. 

To be discussed.

susanodd:
The storage system for videos needs to be extended. @Jetske is working on that in another issue.
This needs to be put on signbank-dev for extensive testing before it can be deployed on signbank.

rem0g:
Up. We will need API for creating, removing, modifying voorbeeldzinnen + upload video attached to voorbeeldzin.

susanodd:
This is not implemented yet.

Jetske:
Just discussed with @rem0g 

Fields for create api:

- "Video_url"
- "Annotations_url"
- "Source"
- "Url" (to be added still)

Fields for edit api:

- "Context (language)"
- "Translations (language)"
- "Source"
- "Url" 
- "Respresentative_glosses": ["gloss_id_1": True, "gloss_id_2": False, "gloss_id_3": False, ...]

(or something like it)

Issue #1332, opened 2024-09-26
Redo the API video upload


Woseseltops:
@rem0g suggested (in #1331) we redesign the video upload. I think the current workflow is:

* Name video files after glosses, store them in a zip
* Upload a zip file to a server somewhere and grab its url
* Call `/dictionary/upload_zipped_videos_folder_json/{datasetid}/` and provide this url
* Call `/dictionary/upload_videos_to_glosses/{datasetid}` to actually link them to the glosses.

I want to replace this with just one step, to be run for each gloss invidually:
* POST `/dictionary/gloss/{gloss_id}/zippedvideo` with just the zipped file in the payload.

I don't think uploading the videos one by one has any real drawbacks in terms of overload, and it does make the whole process easier to follow: you upload a video to a gloss, and if it fails somewhere in the process, it can just return what went wrong for that one gloss.

Also, @rem0g suggests to remove the h264 check:
> it's not needed to check whether it's h264. If it doesnt play, it wont play and we will know that one way or another.`

Woseseltops:
@rem0g , would this make your life better?

@susanodd , can you think of other drawbacks?

susanodd:
I agree the check for h264 is excessive. It was more a check that the user didn't put say, PDF files in with videos.
But if we stick with "try-except" with friendly error reporting for when there was a problem, then this could be programmed away, so not to "check before" but just fail.

The zip uploads them to a different location because the normal upload of a video creates an object at the same time.

My opinion is that we should turn off reversion. (This will create a mess because it does this every time you upload a video and then moves files around.)
I would just remove it entirely and just overwrite an existing video file.
Invest more time in saving the writable archive instead.

For the perspective videos, I made it so it just "overwrites" and existing left or write file and leaves the "GlossVideo object" alone. (So it's link just points to a replaced file.)

Woseseltops:
Summary of @susanodd : there are some extra things we should think about if we want to do this right.

rem0g:
I would rather go this way: 

- Name video file after signbank ID (because of special chars of some glosses)
- Upload the video file to server directly, I don't see good reason for zipping the file besides special chars.
- Signbank will automatically bind the signbank video with gloss. There is no reason this could only be done with call, Python and Django are very well able to process this and it's alot easier to implement.
- Signbank will provide status in gloss query about the status of video (processing,succeed (at which date). This can be checked periodically than continue to upload videos and see what happens with binding forever; thus reducing need for resources and latency.
[...]

PR #1338, opened 2024-10-07
API for create and delete sentences, URL field added #1203

Jetske:
This includes 
- Addition of URL field for annotated video (migration!)
- A correction for the gloss create API in swagger
- New endpoints for get (through glosses), create, update, and delete for annotated sentences + added to swagger


susanodd:
> @susanodd do you have time to review/+merge this? Note the migration :)

I think @Woseseltops ought to look at this as well as the perspective videos pull request, since both have migrations of the video models.
Esp, since there is a new issue about the videos.
It's not clear to me if @Woseseltops is going to modify the video models as well. (#1332)
I fixed some other bugs with the videos on my branch. But @Woseseltops has not looked at the code yet.

Woseseltops (approved):
Code looks okay to me (apart from a few cases of camelCase where we use under_score stuff in the rest the code)!

susanodd:
@Jetske I'll look at this on Monday.
I deployed the fixes to the zip import #1331 #1340 . This issue #1341 the videos seem to have been uploaded.
But the file permissions are all over the place. So perhaps something else is going on.

susanodd:
@Jetske we have that same inconsistency problem again. #1342 
That means that migrations cannot be applied right now.

PR #1340, opened 2024-10-10
Error code improvement

susanodd:
Revised missing videos view url to show videos that have a file but no GlossVideo object.

This branch is being used to detect where the errors arise in video import.

This is working code in a pull request so others can contribute or make sure their suggestions are being followed up.

The Manage Media works on signbank-test again. (This branch.)

For some reason the settings file on signbank-test had TWO different secret keys in it.


susanodd:
I'm just going to merge and deploy this on signbank.
It works fine on signbank-test.
It only affects video upload of zip files and importing those zip files.
We need this in order to debug #1341 with real video files and @rem0g does not want to repeat the commands on signbank-test.

@Woseseltops if you're reading.

PR #1348, opened 2024-10-17
Redo video upload

Woseseltops:
An attempt to make the video upload more elegant, to make the life of @rem0g easier, as discussed in #1332. This can be extended later (with other video types, for example), but I thought this part could already be reviewed.

The endpoint can be used like this:

```python
# The URL where the video file will be sent
url = 'http://35.159.203.216/dictionary/gloss/12/video'

# The path to your video file
video_file_path = 'a_video.mp4'

# Auth
headers = {'Authorization': f'Bearer {BEARER_TOKEN}'}

# Open the video file in binary mode and send it in a POST request
with open(video_file_path, 'rb') as video_file:
    files = {'file': video_file}
    response = requests.post(url, files=files, headers=headers)

# Print the server's response
print(response.status_code)
print(response.text)
```

PS
It also contains a cleanup of the imports in `api_interace.py`, which you can ignore for this review.


susanodd:
Need to check:
Gloss has a lemma
Gloss is not archived
May want a try-except on the add_video method since the save can fail

Jetske:
I added the code for swagger and found that the url pattern wasn't right. May I push that to your branch @Woseseltops ?

Woseseltops:
> @Woseseltops kindly look at the issues related to file permissions, folder creation, exceptions raised. #1356 #1352 #1355

Explanation for myself: these three issues contain stuff related to videos, so should be read before finishing this functionality

Issue #1361, opened 2024-10-29
Manage dataset video storage


susanodd:
This can be previewed on `signbank-susan `under Manage Datasets (button links to new page)

Does anybody else recall what we think the backup filenames are supposed to look like?
1. .bak.bak.bak ... ?
2. .bakNNNNN where NNNNN is an ID ... ?
I find both in GlossVideo paths I vaguely recall that we switched from one to the other.

[DISCUSSION]
I made a Dataset Manager page to view the video paths of the video objects. See the screenshots below.
At first, this was going to be an overview to check constraints and search for bugs in video file paths.
In the screenshots below, there is one gloss that has two GlossVideo objects with version 0.
In the paths that don't end in ".mp4" you see those glosses and paths. (For paths that have MP4 instead of mp4 this could cause bugs with other code.)

You can also see the various backup video paths. As you can see, there is a logic error in the reversion. The file extension has been replaced by ".bak" sequences and the video file extension is gone. (Hmmm. Was that an "mp4" video? That's why there is the ensure_mp4 method. Except that fails on some formats. It's been turned off per request of the API.)
(Uploading via the zip archive just overwrites an existing video, it does not employ reversion. This was intentional to avoid moving a bunch of older files around.)

![manage-dataset-videos](https://github.com/user-attachments/assets/094b61d2-6e38-4333-a77c-909dafe3cac0)

![manage-dataset-videos-2](https://github.com/user-attachments/assets/56659475-a41d-4b84-8406-da752ff85b33)

For completeness, here is showing the weird characters that appear in the filename (dataset tstMH) (see #1356)

![manage-dataset-videos-extra-characters](https://github.com/user-attachments/assets/887a4cae-3b97-46ae-9950-84eb42e1e0ae)

Questions:

- Is this a sufficient feature, or does it need expansion? It has a simple implementation right now. But I could make it a real List View with search capabilities on the Gloss and the specific video file checks.
- (positive or negatve) The entire table includes all the glosses on one page. But there are no videos shown. You can just use the browser search to find strings. It's done by Python not a query, so the displayed data is computed internally, not via a query (as opposed to the List Views, which use ajax calls for each row.)

susanodd:
@Woseseltops what is your opinion about video reversion?
Due to a setting, it's not actually possible to delete the video files. So then the GlossVideo objects cannot be deleted.
The reversion "backs up" a previously uploaded video.
[...]

PR #1362, opened 2024-10-30
#1361: View video paths dataset manager

susanodd:
- This is on signbank-susan

- It is accessible from the Dataset Manager page as a button for the dataset which goes to the new page

### DIFF
--- a/signbank/dataset_checks.py
+++ b/signbank/dataset_checks.py
@@ -1,0 +1,133 @@
+
+
+from signbank.dictionary.models import *
+from signbank.video.models import GlossVideo, GlossVideoHistory
+from django.views.decorators.csrf import csrf_exempt
+from django.http import JsonResponse
+
+
+def gloss_annotations_check(dataset):
+
+    translation_languages = dataset.translation_languages.all()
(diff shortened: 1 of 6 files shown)


Woseseltops:
Note, we need to tell @rem0g once this is live.

Woseseltops (commented):
This code looks fine to me. One thing I'm unsure about this, which I see a few times:

```python
return JsonResponse({})
```

Would it make sense to include an error message in the `JsonReponse`, like we've done elsewhere? Or perhaps this code can just raise an error?

susanodd:
> This code looks fine to me. One thing I'm unsure about this, which I see a few times:
> 
> ```python
> return JsonResponse({})
> ```
> 
> Would it make sense to include an error message in the `JsonReponse`, like we've done elsewhere? Or perhaps this code can just raise an error?

Oh, that function isn't being used at the moment. I was initially planning to make it fetch from the template so it would not be so slow. I can either remove it, or make it actually return information as a json response.
(That would depend on whether @rem0g would like to retrieve this kind of data.)

Issue #1325, opened 2024-09-13
Change ECV descriptions


Jetske:
Remove the phonology from the description and make sure the senses are included correctly

PR #1364, opened 2024-10-31
Only add senses to ecv descriptions, not phonology #1325

Jetske:
@susanodd @Woseseltops @vanlummelhuizen question: MPI want to only have senses in the ECV gloss descriptions, not phonology or other information. Would this be ok to put that in the default settings? Does anyone else use these ECV files?

### DIFF
--- a/signbank/settings/server_specific/default.py
+++ b/signbank/settings/server_specific/default.py
@@ -204,9 +204,9 @@ ECV_FILE = 'myecv.ecv'
 
 #What do you want to include in the ECV file?
 ECV_SETTINGS = {
-    'include_phonology_and_frequencies': False,
+    'include_phonology_and_frequencies': True,
 
-    'description_fields' : [],
+    'description_fields' : ['handedness', 'tokNo'],
 
     # The order of languages matters as the first will
     # be treated as default by ELAN
--- a/signbank/settings/server_specific/default.py
+++ b/signbank/settings/server_specific/default.py
@@ -204,9 +204,9 @@ ECV_FILE = 'myecv.ecv'
 
 #What do you want to include in the ECV file?
 ECV_SETTINGS = {
-    'include_phonology_and_frequencies': True,
+    'include_phonology_and_frequencies': False,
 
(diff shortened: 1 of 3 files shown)


Woseseltops:
When say MPI, you mean ELAN, right? I think ELAN is the only user of the ECVs indeed, but I'm not 100% sure. The fastest way to find out is to just change it and listen for complaints :).

Also, do note that we override this setting in `server_specific/global_sb_new_aj.py`!

Woseseltops (approved):

Jetske:
Yes, I mean ELAN :)
Okay so then I made the changes in server_specific/global_sb_new_aj.py instead of default.py

ocrasborn:
Hi @Jetske @Woseseltops ! :-)

> When say MPI, you mean ELAN, right? I think ELAN is the only user of the ECVs indeed, but I'm not 100% sure. The fastest way to find out is to just change it and listen for complaints :). 

Perhaps 'MPI' means the users of the Corpus NGT there, I'm thinking of PhD candidate Jenni Sander in particular, who works on NGT child data and may refer to Signbank in their own annotations as well. 

Anyway, I just thought I'd add some background here should people wonder in the future how the ECV came to be what it is now: before the Signbank Lexicon Service feature in ELAN existed, annotators adding glosses to their timelines had a hard time differentiating glosses that are very much alike (think of HOUSE-A and HOUSE-B, for instance). We added an extract of the Signbank phonology to the description field at the time, so that annotators would know which gloss refers to which phonological form. (Truth be told, in practice most deaf and hearing annotators found this info hard to read and use, but researchers found it very handy.) Now that the semantic description of signs has taken better shape in the form of ordered senses, AND now that the Lexicon Service in ELAN can pop up a video of the sign form when selecting or inspecting a gloss, it makes total sense to remove the phonology from the description field. I would check with @uklomp at UvA though whether that doesn't interfere with any workflow they have.

susanodd:
@Woseseltops do you want me to merge and deploy this?

Jetske:
@susanodd Yes!

Issue #1356, opened 2024-10-25
File storage system Django (underscore random 7 characters before file extension)


susanodd:
[UPDATE: See the solution in the next comment]

The extra characters on the filename before the extension cause problems with Signbank file naming (see Django implementation code below). This has something to do with obsolete or old video files not being deleted or moved properly by Signbank. The fact that some files are named with these random characters causes *other* to code go wrong. For example, retrieving the images.

https://github.com/django/django/blob/25e724a5d6e331d2d73050d6dcdf2e8593c3aebf/django/core/files/storage.py#L63-L74

```

>     def get_available_name(self, name, max_length=None):
>         """
>         Return a filename that's free on the target storage system and
>         available for new content to be written to.
>         """
>         dir_name, file_name = os.path.split(name)
>         file_root, file_ext = os.path.splitext(file_name)
>         # If the filename already exists, add an underscore and a random 7
>         # character alphanumeric string (before the file extension, if one
>         # exists) to the filename until the generated filename doesn't exist.
>         # Truncate original name if required, so the new filename does not
>         # exceed the max_length.

```

Example:

```
Glosses with Extra Characters in Video Filenames

3 results found

Lemma (Nederlands)	Glos-ID	Annotatie (Nederlands)	Annotatie (Engels)	Video Paths
NOORD-KOREA-B	48577	NOORD-KOREA-B	NORTH-KOREA-B	0: glossvideo/NGT/NO/NOORD-KOREA-B-48577_99B3ELt.mp4
NORMANDIE	48403	NORMANDIE	NORMANDY	0: glossvideo/NGT/NO/NORMANDIE-48403_bEDDTHD.mp4
WIPSTOEL	48743	WIPSTOEL	ROCKING-CHAIR	0: glossvideo/NGT/WI/WIPSTOEL-48743_ANOpcjh.mp4
```

susanodd:
See patch in #1371

Issue #910, opened 2023-03-13
Uploading videos of glosses with a slash creates a subdirectory


susanodd:
The upload video code is creating extra folders underneath the `<first two characters>` folder when a gloss lemma has a `slash` in it.
The slash is being interpreted as a path slash rather than a slash in the lemma text.
This is even worse when there is also a space as the folder and file names end up with spaces in them.
E.g., a subfolder that ends in a space, with files underneath that start with a space.

susanodd:
See #911 

(The creating of subfolders is being done by Django.)

susanodd:
@vanlummelhuizen The problem I've been having with the FileField is that it actually refers to the physical file.

One of the GlossVideo files was without an extension, so it wasn't being displayed.
But the actual physical file had an extension.
It was not possible to modify the file in GlossVideo to refer to the correct file.
I ended up solving this by copying the file to ponyland, downloading it to my computer, removing the GlossVideo entry and uploading it again. (This was in the Chinese dataset.)

Another problem was when I wanted to convert the files again that were in the wrong format.
GlossVideo refers to the actual file and the convert functions are methods on the FileField files.
So it's not possible to "overwrite in the file system". The method has to be applied to the file.
So I had to copy them (because this worried me since the conversion is messed up) and then just apply the method to the GlossVideo object. (But the convert method is broken.)

This seems to be causing errors now with Delete. It's getting messages that the ".bak" is missing when you try to use the Delete video button. (Because of trying to fix the actual files. Django seems to know that something has been done outside of Django.)

I don't like the FileField type.

There is actually no contents in the Video class in the database.
And the setting refers to '/media' instead of protected_media. (The media setting was being used by Admin. That's why I modified it to show the actual path including writable.)

susanodd:
Is this a problem, or a feature?

susanodd:
Check if this is still a problem now that the API also exists for videos.

vanlummelhuizen:
The problem seems to still be there. Some statistics of the situation on the live server:

Of a total of 21299 GlossVideo instances, there are 562 that refer to video files that don't exist. 259 of them end in `.bak`.
There are 20737 GlossVideo instance that refer to existing files. 1549 of those end in `.bak`.

<details>
  <summary> Python code used...</summary>

  ```python
from signbank.video.models import *
non_existing_files = []
existing_files = []
for gv in GlossVideo.objects.all():
  path = os.path.join('/var/www/writable/', gv.videofile.name)
[...]

PR #1371, opened 2024-11-06
#1356: Patch to prevent 7-char sequence and missing objects

susanodd:
Remove an existing gloss video file that does not have an object; lazy creation of gloss video object if file does not have it.

Solution thought up by @Jetske and @susanodd when looking at the video storage.

This patch includes two parts.

- if a Gloss Detail page is visited and there is a video file but there is no GlossVideo object pointing to it, an object is created for it.

- if a new video is being uploaded but there already exists a file with this name, but no GlossVideo object, the file is deleted so that the new file can have a correct path, without the 7-char code added to it.

Based on recent video creations, this is happening, so this is necessary.

If you want to try this out on your local signbank, then 

- set DEBUG_VIDEOS to True.
- Then for a particular gloss, remove its videos in Gloss Edit.
- Then on the file system, copy a video file to the "normal" location for this gloss's video file. (I.e., put back the video file, but on the file system, not using the Form upload.
- Then try either "renewing the page" (the new GlossVideo object will be created and you will see messages in the venv log.
- Or immediately upload a new video without reloading the page. Then the "add video" will erase the file before creating a new object. This prevents a weird file name from being created.

TO DO (different branch)
Part 3, if there already is a file with the weird filename, the object and the file both need to be fixed.
This has recently happened with NOORD-KOREA-B. It already had a video, but no GlossVideo object. So the new video has a 7-char sequence.

We don't know how it happens that a GlossVideo object is not created. Perhaps the transaction does not complete or there is a file system error, or the database is locked.

### DIFF
--- a/signbank/dictionary/models.py
+++ b/signbank/dictionary/models.py
@@ -2329,12 +2329,24 @@ class Gloss(models.Model):
         return escape_uri_path(image_path) if image_path else ''
 
     def get_video_path(self):
-        from signbank.video.models import GlossVideo
+        from signbank.video.models import GlossVideo, get_gloss_video_filepath
         try:
             glossvideo = GlossVideo.objects.filter(gloss=self, glossvideonme=None, glossvideoperspective=None).get(version=0)
             return str(glossvideo.videofile)
         except ObjectDoesNotExist:
-            return ''
+            relative_path = get_gloss_video_filepath(self)
+            if relative_path:
+                if settings.DEBUG_VIDEOS:
+                    print('get_video_path: already existing file without GlossVideo object found: ', relative_path)
+                # there is a video file but no GlossVideo object
+                video = GlossVideo(gloss=self, glossvideonme=None, glossvideoperspective=None, version=0)
+                video.videofile.name = relative_path
+                video.save()
+                if settings.DEBUG_VIDEOS:
+                    print('get_video_path: GlossVideo object created for already existing file: ', relative_path)
+                return relative_path
+            else:
+                return ''
         except MultipleObjectsReturned:
             # Just return the first
             glossvideos = GlossVideo.objects.filter(gloss=self, glossvideonme=None, glossvideoperspective=None).filter(version=0)
@@ -2362,7 +2374,7 @@ class Gloss(models.Model):
 
     def add_video(self, user, videofile, recorded):
         # Preventing circular import
-        from signbank.video.models import GlossVideo, GlossVideoHistory, get_video_file_path
+        from signbank.video.models import GlossVideo, GlossVideoHistory, get_video_file_path, get_gloss_video_filepath
 
         # Create a new GlossVideo object
         if isinstance(videofile, File) or videofile.content_type == 'django.core.files.uploadedfile.InMemoryUploadedFile':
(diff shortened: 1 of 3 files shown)


vanlummelhuizen:
> Based on recent video creations, this is happening, so this is necessary.

Have you seen filenames with the 7-char sequence being created? It would be nice to know in what situation it happens.

I think it is good if we could also do this in batch mode, e.g. with a Django command. If we do this, and clean up other things (see also https://github.com/Signbank/Global-signbank/issues/910#issuecomment-2459751097) we avoid that the clean up is only done when Signbank is used.

susanodd:
> > Based on recent video creations, this is happening, so this is necessary.
> 
> Have you seen filenames with the 7-char sequence being created? It would be nice to know in what situation it happens.
> 
> I think it is good if we could also do this in batch mode, e.g. with a Django command. If we do this, and clean up other things (see also [#910 (comment)](https://github.com/Signbank/Global-signbank/issues/910#issuecomment-2459751097)) we avoid that the clean up is only done when Signbank is used.

Yes, we investigated this. There was a recent one for NOORD-KOREA-B. So we looked at the files and at the admin.
There already existed a file for the gloss, but there was no GlossVideo object for that file. The file sizes were identical. The old file was from the past. The new file ended up with the 7-char sequence. And that was also inside its path in GlossVideo.
So it creates this sequence if there already exists a file with the desired name. (Django does this. It's that Django implementation  function in the issue.)

I suspect it's something to do with the database getting locked that it hasn't finished making the object and then ends up locked but saved the file but didn't complete.

susanodd:
All the GlossVideo objects with non-existent file paths have been deleted.
I did this in the admin by hand. That way I could confirm every removal.

Woseseltops (changes requested):
Thank you for providing bite sized pull requests :).

About the code:
* I think the changes in ``signbank/dictionary/models.py`` are a good opportunity for a guard clause :).
* I think the function name `get_video_path()` is a bit misleading if that functionality will also correct missing connections. If we don't want to rename the function to better reflect that logic inside, perhaps we can 'communicate' this extra behavior by making it optional, and including a parameter for it? Something like `get_video_path(fix_videos_without_glossvideo_object=True)`
* I think this is easier to maintain with f-strings:

```python
name = str(self.examplesentence.id) + ': ' + str(self.action) + ', (' + str(self.datestamp) + ')'
```

```python
name = f"{self.examplesentence.id}: {self.action}, ({self.datestamp})"
```

About the functionality: creative solution to 'fix' the connections when they are visited in the UI... but wouldn't it make more sense to have a management command that can check and fix all of Signbank? I see @vanlummelhuizen also suggests this. Perhaps good for a separate issue?

susanodd:
> Thank you for providing bite sized pull requests :).
> 
> About the code:
> 
>     * I think the changes in `signbank/dictionary/models.py` are a good opportunity for a guard clause :).

Okay.

>     * I think the function name `get_video_path()` is a bit misleading if that functionality will also correct missing connections. If we don't want to rename the function to better reflect that logic inside, perhaps we can 'communicate' this extra behavior by making it optional, and including a parameter for it? Something like `get_video_path(fix_videos_without_glossvideo_object=True)`

I'll modify the name. The function is called a lot. It can't be easily called in the templates if there are parameters. (Yes, then to use a vertical bar as is done for other functions that required the language. (There are actually about 5 different functions that kind of overlap. It used to be an option on some functions to check whether the file exists. But if that's not checked, it causes other problems.)

I can't really see why we would want to "not" use it. When we switched to field choice objects, there were some functions that were changed in order to support the "change" since it was difficult to find all the places in the code. Then that code was evolved away after it was all working.

This is meant as a patch since we can't control the API transactions from not completing.

>     * I think this is easier to maintain with f-strings:
> 
> 
> ```python
> name = str(self.examplesentence.id) + ': ' + str(self.action) + ', (' + str(self.datestamp) + ')'
> ```
> 
> ```python
> name = f"{self.examplesentence.id}: {self.action}, ({self.datestamp})"
> ```

That's @Jetske's code. Yes, the f-strings are cool.

> About the functionality: creative solution to 'fix' the connections when they are visited in the UI... but wouldn't it make more sense to have a management command that can check and fix all of Signbank? I see @vanlummelhuizen also suggests this. Perhaps good for a separate issue?

Yes, I am making a management command. Except we don't actually know how these file names arise. (If a transaction does not complete because the database gets locked. That's what I suspect.)

susanodd:
Another question, what do we think about non-mp4 videos?

Woseseltops (approved):

Issue #1373, opened 2024-11-07
Repair videos with incorrect backup filenames


susanodd:
- This is about files and gloss video objects with "old style" backup paths. These contain sequences of "bak" at the end, in accord with the version number of the gloss video.
- The code was revised in the past to use "bak" followed by the ID of the gloss video object instead.
- But some of the old code still uses the obsolete format. This issue is to remove the old format from the code in the creation of new video objects.

The filenames can be fixed by a command in #1412

<details>

Reminder of filenames existing in the code:

The new format leaves the "mp4" in the filename. So the old format files with 'bak bak" sequences are missing the video format. I was always "mp4" since we used to use "ensure_mp4" on (Signbank uploads, not API). But we don't know this before because there is still code that mentions the `(version * ".bak")` suffix. (See #1374)

susanodd:
@vanlummelhuizen about "reverse renaming the bak bak files". Do we just assume they are mp4 ?

vanlummelhuizen:
> @vanlummelhuizen about "reverse renaming the bak bak files". Do we just assume they are mp4 ?

There are files that are not MP4. The listing below shows the files in `glossvideo`  that end in `.bak` and do not have the string 'MP4' in the file type.

```sh
root@signbank-new:/var/www/writable/glossvideo# find . -type f | grep -P '\.bak$' | xargs -i file {} | grep -v MP4 | less
./NGT/ON/ONE-AND-A-HALF-B-40012.bak17345.bak.bak: ISO Media, Apple iTunes Video (.M4V) Video
./NGT/ON/ONE-AND-A-HALF-B-40012.bak17346.bak: ISO Media, Apple iTunes Video (.M4V) Video
./NGT/ON/ONE-AND-A-HALF-B-40012.bak.bak.bak.bak.bak: ISO Media, Apple iTunes Video (.M4V) Video
./NGT/ON/ONE-AND-A-HALF-B-40012.bak17344.bak.bak.bak: ISO Media, Apple iTunes Video (.M4V) Video
./NGT/BL/BLIKJE-A-36667.bak.bak: ISO Media, Apple iTunes Video (.M4V) Video
./NGT/BA/BACTERIE-A-40006.bak13572.bak: ISO Media, Apple iTunes Video (.M4V) Video
./NGT/te/testlemmaidglosstranslation6-3729.bak.bak: ISO Media, Apple iTunes Video (.M4V) Video
[...]

PR #1377, opened 2024-11-08
#1373: Imperative function with side effects to rename video files

susanodd:
when they have the wrong bak bak names.

this will end up in a different place in the code. It's not clear how much feedback will be needed or how long it will take to run.

### DIFF
--- a/signbank/dataset_checks.py
+++ b/signbank/dataset_checks.py
@@ -27,6 +27,61 @@ def gloss_annotations_check(dataset):
     return results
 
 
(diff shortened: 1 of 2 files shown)


susanodd:
I needed to create a pull request to save my branch in order to switch to the other pull requests. Thus it has the label Draft.
Although I would welcome feedback from others since all the branches and video issues are interrelated.

susanodd:
This command is for the virtual environment. Only backup videos with the wrong kind of filename are renamed.

Issue #1350, opened 2024-10-22
Link Wordnet to Signbank


Jetske:
Add "Wordnet synsets" under senses

https://www.sign-lang.uni-hamburg.de/easier/sign-wordnet/index_ngt.html

E.g. 
https://www.sign-lang.uni-hamburg.de/easier/sign-wordnet/sign/ngt.1863.html
this page has a link to signbank -> do this in reverse in signbank, but instead don't link to the gloss but to every synset of that gloss

With a list like this:
![image](https://github.com/user-attachments/assets/c2c1f113-383c-41f7-a689-95ddb704c426)

Jetske:
![image](https://github.com/user-attachments/assets/b3be28cd-d21e-4f74-a7bf-887f20687df7)

Okay I fixed the missing lemmas and descriptions with web scraping the urls that do work. It works, but is probably not really ideal as it takes quite long and there's a risk of too many requests. 

@Woseseltops @vanlummelhuizen if you have ideas let me know

vanlummelhuizen:
@Jetske First a small technical remark/question about your command. In method `download_links_csv` you start a context with `with ...`:

https://github.com/Signbank/Global-signbank/blob/2c32b7f9377ac2bbc9da401f08c3fe57a50c7643/signbank/dictionary/management/commands/update_wordnet_links.py#L41-L48

It seems to me you end it prematurely on line 48. 

---

If I correctly understand what you are trying to do, I think it boils down to two questions:

1. When should the script run? Once a day, every hour or triggered by an event?
2. What should be done and what not? Can the script skip parts if they are not necessary at that time, so that it runs faster?

Depending on the answers we could find a better solution.

Jetske:
@vanlummelhuizen Yes, that can probably be done without the 'with'

I would say ideally once a day. Everything should be done, as it is meant to update the synsets according to new versions of the different files that are downloaded. Each of these files may contain changes.

Jetske:
Yes, to prevent too many requests it's probably better to check which links are updated and only edit the synsets in case of changes, rather than what is done now (delete everything and make again).

Jetske:
@vanlummelhuizen done, now it just deletes the links between glosses and synsets and creates them again. The synsets themselves are not deleted, except if they are not used anymore. Only for new synsets it makes a web scrape request, if it could not be found in wordnet. That should make it much more efficient.

Woseseltops:
Sam Bigeard emails me about this

[...]

PR #1382, opened 2024-11-08
Add synset wordnet links to glosses

Jetske:
Waiting also for approval of @uklomp in #1350 

This is a command that adds/updates the wordnet synset information. Note: this only works for NGT. It takes a little while to run it. I wonder if this is the best way, for example, it is also possible to add a button "Update synsets" to the manage datasets page. But I think it's better to run the command for example once a week to update it automatically. 

The first time the command is used, it creates all synsets, finds missing wordnet information, and links them to glosses. After that, when updating synsets, it will use those that are already there, create new ones and deletes the ones not used anymore.

To do:
Add your wordnet login credentials to server specific settings
migrate


susanodd:
@Jetske can you add the verbose names to the fields of the new model? That is used elsewhere, for example in queries if they are included in search fields.

Jetske:
@susanodd adding verbose names would be redundant as they would be the same as the field names, or not?

susanodd:
> @susanodd adding verbose names would be redundant as they would be the same as the field names, or not?

The verbose names will be put in the translation files. Then if the fields are used in search forms, they will be retrieved that. The query parameters also make use of these.

vanlummelhuizen:
Thanks @Jetske 

One important thing: nltk and bs4 are not in requierments.txt.

Question: the Synset model now links to Glosses using a ManyToManyField and to Dataset using a ForeignKey. Can Dataset be derived from Glosses, making the ForeignKey to Dataset unnecessary?

You now download the data and save it in a file on disk. Would it be feasible to keep the data in memory and use it from there? It would avoid potential file system problems.

Some minor remarks below.
<details>

Some methods in `update_wordnet_links.py` are quite long and some have quite some branching (for, if). If possible (time wise) you could see if you can make them shorter and simpler.

In lines 57-79 you are doing the same thing twice with a minor difference: 'links' and 'signs'. It would be a candidate to extract a new method from.

In `get_links_data` there are two nicely separated parts. Getting the signs from the CSV could be done in a separate method.

Note that these remarks may interact with the thing I said about saving data to disk.
</details>

signbank/dictionary/management/commands/update_wordnet_links.py
```diff
@@ -21,22 +22,36 @@ class Command(BaseCommand):
...
+    def download_csv(self, session, download_url, csv_type):
+        """ Download the WordNet links and signs CSV file from the server. """
+
+        download_payload = {
+            "submit": csv_type
+        }
+        download_response = session.post(download_url, data=download_payload)
+        if download_response.status_code == 200:
+            csv = download_response.content
+            print("Links CSV downloaded successfully.")
```
vanlummelhuizen, line 44: `print("Links CSV downloaded successfully.")` should probably be `print("{csv_type} CSV downloaded successfully.")`

vanlummelhuizen:
Good busy!

I saw you made the variable names for the urls lower case. I thought upper case was a good choice since they are more like settings/constants. If anything, I would probably move them to the top of the file so that anyone reading this code instantly knows we are dealing with external urls.

Jetske:
@vanlummelhuizen yes, that was too quickly changed to lowercase. Should all be done now :)

Issue #1375, opened 2024-11-07
archived glosses or deleted glosses


uklomp:
I noticed that glosses aren't deleted anymore but now archived. This is great, as a colleague accidentally deleted a gloss and I was able to retrieve it. However, sometimes we really want to delete a gloss and/or lemma because we want to add a complete new entry with the same gloss and/or lemma (and that would be easier than adjusting the whole entry). I think for the deletion of a lemma, the deletion of the glosses is first needed. How do we now delete the glosses?

susanodd:
> I noticed that glosses aren't deleted anymore but now archived. This is great, as a colleague accidentally deleted a gloss and I was able to retrieve it. However, sometimes we really want to delete a gloss and/or lemma because we want to add a complete new entry with the same gloss and/or lemma (and that would be easier than adjusting the whole entry). I think for the deletion of a lemma, the deletion of the glosses is first needed. How do we now delete the glosses?

This probably needs to be done in the management section. This is because there are lots and lots of other objects that potentially refer to the "to be deleted gloss".

We need to discuss this on our side. There are also a bunch of "signals" sent to other code when a gloss is deleted.
The archival was implemented in such a way so that nothing was done with the video files. They are just not shown. But if a gloss really is deleted, there is an avalanche of other objects being deleted. For example, senses, lemmas, lemma translations, sentences, gloss video (objects).

uklomp:
So... in short: delete does not delete, and you need to talk about it on your side? let's talk about it the next meeting.

susanodd:
Normally, if an object is deleted, all the related objects are also deleted. Related objects are previewed in the "Delete popup" of Gloss Edit if you choose the delete the sign.

Multiple choice

- just go ahead and delete all the related "cascaded delete" objects as well
- the user first deletes related objects by hand prior to deleting to make sure there are no surprise side effects
- since most relations are symmetric, delete the relations, make sure the related gloss doesn't need this
- the user deletes videos if they exist
- make sure deleting morphology where the morphemes are not used by other glosses does not delete the morphemes

Woseseltops:
After discussion with @susanodd , could 'cleaning' an existing gloss also solve your problem? So an existing gloss looks identical again to a fresh gloss, as if you just made it.

susanodd:
Teams Update

A Delete option will be offered in the Summary of the Lemma where you the glosses that share the lemma are shown.
That is where you can Restore a gloss. So it will also be possible to Trash a gloss.

[...]

PR #1391, opened 2024-11-15
#1375: Trash button in Lemma Update for gloss deletion.

susanodd:
The trash icon shows for archived gloss in the "update lemma" page accessible via the "show all lemmas"

trash-button-next-to-archived-gloss-lemma-view


Woseseltops (approved):

susanodd:
@vanlummelhuizen are you there?

When a gloss is really deleted rather than archived, the video files are not deleted.

It's because of this setting:

DELETE_FILES_ON_GLOSSVIDEO_DELETE  (which is False)

And the files are merely unlinked from the objects.

Issue #1392, opened 2024-11-15
Video Still Choices When Creating Poster Image


susanodd:
As discussed in Team meeting.

@uklomp This is available on `signbank-susan`, as a button in Gloss Edit, next to the Create ("middle frame") button that already exists.

(Screenshots: https://github.com/Signbank/Global-signbank/pull/1393#issuecomment-2494277118)

See if the number of images at 15 fps is sufficient to choose from.

<details>
Since it takes time to generate the images, and they need to be viewable in the browser, there are two buttons inside the pop-up. One to generate the images and one to show the images. Then you can hover over them to choose.
A button at the bottom of the popup is for saving your selection. If you click Dismiss, and exit the modal, you can see the images when you open the modal again. They're deleted after you choose.

susanodd:
@uklomp this has been deployed to signbank

uklomp:
I think we agreed on not having the multiple buttons (generate still), and also on having less stills, right? 
so: when I click 'choose image' I already expect the stills to be there. It's okay if they still need to load (although it would be good to have the text 'images are being loaded' or something like that) but it's not necessary to then still click 'generate' in my opinion.
Then: I asked for 3 stills, you gave me 15 first, but now there are 45. That's too much, because now I need to scroll to see them all and there are too many options. Let's stick to 9 for now, so I can see them in a 3 by 3 display.

susanodd:
> I think we agreed on not having the multiple buttons (generate still), and also on having less stills, right?
> so: when I click 'choose image' I already expect the stills to be there. It's okay if they still need to load (although it would be good to have the text 'images are being loaded' or something like that) but it's not necessary to then still click 'generate' in my opinion.
> Then: I asked for 3 stills, you gave me 15 first, but now there are 45. That's too much, because now I need to scroll to see them all and there are too many options. Let's stick to 9 for now, so I can see them in a 3 by 3 display.


The "choose image" is the original button. "Choose still" opens a model. There is only one button inside of the modal.

The number of stills is dependent on the length of the video. If the video is long there are more stills.

(The computer can't know which are the interesting images. If one tries IJSBEER there probably aren't enough since the middle bit is the part between IJS and BEER.)

I will let @Woseseltops decide. He approved it as it is now.

Woseseltops:
[...]

PR #1393, opened 2024-11-20
#1392: Generate still images for video, choose modal

susanodd:
The code shows the radio bullet and outlines the chosen image in red inside a modal pop-up.

(Radio bullets are shown in the display of still images. On Firefox, it works to hide the input bullet, but that does not work on Safari, so the bullets are visible.)

A frame rate of 15fps is used to extract the images. The first second is ignored from the video as per feature request.

It may be necessary to show more images than 15 fps. But that would be an awful lot of images. See how it goes in usage.


Woseseltops (changes requested):
Hey, I'd like to test this functionally, but from the information I see in the pull request or the issues, I don't understand where exaclty on `signbank-susan` I can find this... can you explain this to me in the form of a screenshot?

Already a technical remark: I'm confused by the function naming here. I see `gloss_video_stills_ajax`, `create_stills`... another `create_stills`, `make_thumbnails`. I think more descriptive names would make this easier to understand what is the difference between all of these.

susanodd:
It's in Gloss Edit, next to the Create Citation Image

choose-still-image-button-gloss-edit

highlighted-still-image-in-pop-up

susanodd:
> Hey, I'd like to test this functionally, but from the information I see in the pull request or the issues, I don't understand where exaclty on `signbank-susan` I can find this... can you explain this to me in the form of a screenshot?
> 
> Already a technical remark: I'm confused by the function naming here. I see `gloss_video_stills_ajax`, `create_stills`... another `create_stills`, `make_thumbnails`. I think more descriptive names would make this easier to understand what is the difference between all of these.

I copied the naming that was used in similar functionality urls. I prefer the name stills but other functions also use citation image or thumbnail. There are already many similar functions and methods so it's difficult to identify them from each other. Unless we go for really long phrases with underscores? Is that better for being clear? Or worse for being long-winded?

Woseseltops (commented):
Just had a look, I think the function names are way better now, thanks! I was also able to find the button to test it functionally now (again, thanks :) ). 

One remark: there are separate buttons now for 'generate stills' and 'show stills', but I think this is too low level for the end user... the end user just wants to choose still. I propose to always do 'generate' and 'show' each time this modal is opened, and hide the buttons for the user.

susanodd:
choose-still-one-button

susanodd:
The changes were made yesterday as requested. The code works fine. No timeout necessary.

Since you have limited time to review, I would prefer that you spend that time reviewing the "reversion bug fix" pull request.

Issue #1400, opened 2024-11-27
Retrieve more signs when working with function 'split up senses'? 


uklomp:
Would it be possible to retrieve / analyse about 500 glosses in the function "Split up senses", instead of max 100? 
If the answer isn't easy, let's talk about it tomorrow.

Woseseltops:
Proposal: just try to make it longer. If our browsers explode (the page might to heavy to everything at once) we move to an Ajax-based approach like @susanodd pioneered herself within Signbank.

susanodd:
This is on `signbank-susan` now.

I can't tell whether the browser is choking on the amount of code. The "debug" of Firefox won't show all the nodes.

Can you try it out @Woseseltops ?

susanodd:
@uklomp to clarify, do you want 500 search results, or 500 results per page?

What would be ideal?

It's on signbank-susan now so you can test it.

uklomp:
Ideally, we get all the results from the search query, and still have the option to show a limited number per page - let's say 50 or 100 per page.

susanodd:
> Ideally, we get all the results from the search query, and still have the option to show a limited number per page - let's say 50 or 100 per page.

Okay, that is good to hear for the results per page. That should be possible.

For a technical question about the browser truncating data, I need to ask @Woseseltops. It's a problem with the amount of data per page.

susanodd:
This is live now on Signbank.

PR #1404, opened 2024-11-28
#1400: Unlimited retrieval and 500 glosses per page on Split Senses

susanodd:
This change is trivial.

- @Woseseltops this is on signbank-susan now.

- You do a Search on the dataset (NGT since that's what we're testing)

- Then go to `Analysis -> Split Up Senses`  (it shows them for the most recent query results)

- The "browser debug window" has trouble showing all the nodes!!!!

- But it does indeed work!!!

- But I suspect it might not be showing all the results????
- In the python code, I set it to display 500 results per page.
- But there are only 2 pages shown.

- Can you experiment with this yourself to see what the browser is doing? I'm using Firefox so I can't test it for Windows

- I also don't know whether Firefox is just throwing stuff away because it's too much???

### DIFF
--- a/signbank/dictionary/adminviews.py
+++ b/signbank/dictionary/adminviews.py
@@ -6656,7 +6656,7 @@ class KeywordListView(ListView):
 
     model = Gloss
     template_name = 'dictionary/admin_batch_edit_senses.html'
-    paginate_by = 25
+    paginate_by = 500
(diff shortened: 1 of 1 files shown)


Woseseltops (approved):
Took some time to load, but other than that everything is fine

PR #1414, opened 2024-12-05
#1392: Removed another button from Choose Still

susanodd:
The disadvantage of this modification is that the images are generated every time the user clicks on Choose Still.

This is because the operation needs to clean up after itself in case the user has uploaded a different video, etc.

If the user decides to choose a still image, but does not actually select any image (chooses "dismiss") then the images are not cleaned up from the server.


Woseseltops (approved):
Thanks for removing this button as well :)

Issue #1403, opened 2024-11-28
Make more information of gloss publicly accessible


rem0g:
For example https://signbank.cls.ru.nl/dictionary/gloss/729 only show video and senses, it's too short of information for non logged in users.

The page should display relation with other glosses, such as variants (METRO-A and METRO-B). 
Phonology
Voorbeeldzinnen
NMM videos
And finally 3-perspective video

@uklomp let me know if you agree with this.

uklomp:
There is indeed a difference between the public version and the gloss-detail page (last one only visible for logged in users). As I understand it, you don't want this difference anymore Gomer? I'd like to know why this difference was there in the first place, maybe there was a good reason for this? @ocrasborn

susanodd:
It's fairly easy to add the requested information.

Since the non-NGT datasets have their own permissions, and since each gloss can be visible or not (In Web setting), then the information that's shown can be controlled that way.

The new Synset Wordnet information might be nice in the public view. There are also social media links in the public view.

rem0g:
> I will discuss this together with Floris and Marloes - on the one hand reconsidering what should be publicly available, on the other hand also keeping in mind that colleagues might be developing a more public friendly interface anyway.

Not to start a discussion here; but since Signbank is a scientifically lexicographically dictionary subsided by the dutch government and they are all available on the Signbank commercially-free; it would not make sense not to publish all relevant details if you get my gist?  

I also want to know opinion from @ocrasborn

uklomp:
> > I will discuss this together with Floris and Marloes - on the one hand reconsidering what should be publicly available, on the other hand also keeping in mind that colleagues might be developing a more public friendly interface anyway.
> 
> Not to start a discussion here; but since Signbank is a scientifically lexicographically dictionary subsided by the dutch government and they are all available on the Signbank commercially-free; it would not make sense not to publish all relevant details if you get my gist?
> 
> I also want to know opinion from @ocrasborn

I completely agree about the subsidy etc.. but at the same time I'm thinking: having an account is not really difficult right? You only need an email address. People get access automatically so I dare say it's quite publicly available. And some might say that the simple overview is actually more user-friendly. 

[...]

PR #1419, opened 2024-12-08
Expand public view

susanodd:
This implementation does the following:

- improvements to dynamic language fields in search forms for List View searches (query_parameters.py, and other ListView classes, makes use of `self.search_form.is_bound` in `get_queryset` to initialise dynamic form language fields as soon as possible, since they need the request which is not available in `__init__`

--  done in this branch but applicable to all code, for the Lemma Groups, the search form was not being processed correctly, the code was falling through to the end of get_queryset in GlossListView, I first noticed this in the branch and repaired it here, the lemma groups is a "sideways" call of `/signs/search/` and not via a search form

- anonymous and logged in share the GlossDetailView now

- a new template for the public view of the gloss, `gloss.html`

- new settings.PUBLIC_PHONOLOGY_FIELDS, etc

- modified object `get` function for template to account for obtaining the gloss id from the request and more permission checks to make sure it can be viewed if the user is anonymous and direct to appropriate template

- modified Gloss Search form for anonymous users (uses settings)

- shared context variables make use of a new Python file `context_data_gloss.py`

- For original non-shared context and Phonology (not yet moved to new Python file) See #1445


susanodd:
The sentence video I can't figure out where the width is being set (@Jetske ?)

See the screenshots in #1403 

I could not figure out where the width of the sentence is being set either. It wraps.

There is some overlapping usage of class names or css settings on elements. The social media links needed to be in a div outside the rest of the page elements nesting structure.

uklomp:
Please wait until we actually made a specific request - I don't want you to implement anything we don't want (yet)

susanodd:
No reviewers are assigned yet. If you want to see how it is taking form, it's on `signbank-susan`

vanlummelhuizen:
@susanodd Can this be a draft PR?

susanodd:
Ready for review. See #1445 also.

Woseseltops (commented):
I've checked the functionality, which looks fine, and I've checked the functionality, which looks fine as well, so I think this can be merged... but I don't understand on a high level what is going on here. To improve/document the code here, I've created #1415

Issue #1424, opened 2024-12-10
Selected Datasets being overwritten to be NGT


susanodd:
[SEE PULL REQUEST CODE]

[I FIXED THE PROBLEM BY UPDATING THE CONTEXT VARIABLE CODE]

@vanlummelhuizen please check the pull request code. Short and Sweet.

The code still ends up getting to the DEFAULT dataset if all the other options are exhausted. I don't know if that's related to the package/info

[ORIGINAL REPORT OF WEIRD ERROR]

<details>

This happens when you try to browse and search other datasets.

I put a bunch of print statements in the code and I can't find out where it's happening. It depends on control flow.

The "selected datasets" context variable ends up being NGT, while the Signbank banner remains correct.

It's doing this in the context processor file!

The order of evaluation is unclear to me. The following does not set the session variable. But it does set the dataset to NGT.

The views also have context variables and also set the selected datasets.

https://github.com/Signbank/Global-signbank/blob/1a0d0e5f7b2a5e7921746c1c58197938023dffda/signbank/context_processors.py#L9-L30

susanodd:
This should be working now. Maybe check #1130

PR #1425, opened 2024-12-11
#1424: Rework of get selected datasets context

susanodd:
See the bug in action! Get your tickets here!

If you try out the anonymous view on `signbank-susan` and select `Kata Kolok` (which has public signs), you will see that this does not work. Internally it keeps setting the dataset to `NGT`.

This branch fixes that.

In order to not keep duplicating code, this solution is on a different branch. (Because @vanlummelhuizen complains about repeating code.)

This pull request is needed by the public views branch. Kindly approve asap.

I realise that there are more instances of the function in question that need to be replaced (function call instead of explicit code). But the function needs to be repaired first to avoid massive amounts of code changing everywhere.

This has to do with the anonymous user not having any profile with stored "selected datasets". Those are stored in the session variable for anonymous users.


vanlummelhuizen:
@susanodd When only looking at what you changes do, it gets an OK from me. 

I noticed that the implementation of `get_selected_datasets` in `context_data.py` has quite a few nesting if/elif/else statement. I thought this could be improved. See the diff below. Do you think this is a more maintainable implementation? Could you see if it still works as expected/it should work?

<details>
<summary>Diff of my refactor</summary>

```diff
diff --git a/signbank/dictionary/context_data.py b/signbank/dictionary/context_data.py
index c7afdcf6..1c9ef999 100644
--- a/signbank/dictionary/context_data.py
+++ b/signbank/dictionary/context_data.py
@@ -24,26 +24,14 @@ def get_selected_datasets(request):
     """
     Get the selected datasets
     """
-    if request.user.is_authenticated:
-        selected_datasets = get_selected_datasets_for_user(request.user)
-        if 'selected_datasets' in request.session.keys():
-            session_datasets = request.session['selected_datasets']
-            selected_datasets_acronyms = [ds.acronym for ds in selected_datasets]
-            if session_datasets != selected_datasets_acronyms:
-                # if the session variable does not match, fix it
-                request.session['selected_datasets'] = selected_datasets_acronyms
-                request.session.modified = True
-        else:
-            # add the selected datasets to the session variables
-            selected_datasets_acronyms = [ds.acronym for ds in selected_datasets]
-            request.session['selected_datasets'] = selected_datasets_acronyms
-            request.session.modified = True
-    elif 'selected_datasets' in request.session.keys():
-        selected_datasets = Dataset.objects.filter(acronym__in=request.session['selected_datasets'])
-    else:
-        # this will end up setting it to the DEFAULT DATASET for anonymous users
-        selected_datasets = get_selected_datasets_for_user(request.user)
-        request.session['selected_datasets'] = [ds.acronym for ds in selected_datasets]
+    if not request.user.is_authenticated and 'selected_datasets' in request.session.keys():
+        return Dataset.objects.filter(acronym__in=request.session['selected_datasets'])
+
+    selected_datasets = get_selected_datasets_for_user(request.user)
+    selected_dataset_acronyms = [ds.acronym for ds in selected_datasets]
+    dataset_acronyms_in_session = request.session.get('selected_datasets', [])
+    if not dataset_acronyms_in_session or selected_dataset_acronyms != dataset_acronyms_in_session:
+        request.session['selected_datasets'] = selected_dataset_acronyms
         request.session.modified = True
     return selected_datasets
 ```

</details>

Woseseltops (commented):
Agree with suggestion of @vanlummelhuizen (another example of the guard clauses I often bring up ;) ). If that is realized, feel free to merge.

susanodd:
There probably needs to be a similar method in context_data for the search type. It is also used in the session variables as well as the forms. It is passed around between the various views. This is linked to the search results which are also passed around. There is a lot of code that inspects the session variables as well as these two variables when they belong to the context of self variables of the view. This has sometimes caused duplication because both the query set and the context and the forms need this.

susanodd:
I did not merge or deploy this yet. The amount of changes increased, although it's all just function calls.

Issue #1427, opened 2024-12-11
search query doesnt work with split up senses


uklomp:
my colleague tries to find glosses starting with a B to start working on the 'split up senses tool' and first it worked, but they didn't get all the results so they tried again and now it's not working anymore. In multiple browsers: firefox and chrome.

![image](https://github.com/user-attachments/assets/a81b1fa1-beb3-4777-bc74-8ba52da49fda)

they got this result: 

![image](https://github.com/user-attachments/assets/a8bd33f9-d3e6-4f34-976a-446e6a31443b)

susanodd:
I think I found the bug. There is a setting MAX_SCROLL_BAR that is interfering with the search results. The setting is very old. It is being used in the code of GlossListView.

susanodd:
@uklomp this is live now

If you still experience anything strange, please keep track of the sequence of commands.

The fix assumes you:
do the "gloss search" then immediately go to the Split Up Senses page (i.e., using the results you just obtained)

The search results are stored in a session variable.

<details>

Otherwise,
For example,

If you do the search but do something else before going to the Split Up Senses page, we need to know the sequence.

Say, if you do a search, look at some glosses, etc., then go to Split Up Senses.

Keep track of the "stuff in between". Also if you have multiple windows open. The results are stored in a session variable.

(That would be some "other" part of the code is changing the search results or resetting the search type before you get to Split Up Senses.....

uklomp:
thanks! I will ask my colleague if it works now.

PR #1428, opened 2024-12-11
#1427: Removed Max Scroll Bar limit on Gloss Search

susanodd:


signbank/dictionary/adminviews.py
```diff
@@ -6713,8 +6703,9 @@ def get_queryset(self):
             # the query set is a list of tuples (gloss, keyword_translations, senses_groups)
             return []
 
-        if ('search_type' in self.request.session.keys() and
-                self.request.session['search_type'] != self.search_type):
+        session_search_type = self.request.session['search_type'] if 'search_type' in self.request.session.keys() else ''
```
vanlummelhuizen, line 6706: You could do something like `... = self.request.session.get('search_type', '')` here. See also https://www.w3schools.com/python/ref_dictionary_get.asp

vanlummelhuizen (commented):
I believe this is all good (one minor remark), although I did not test.

Woseseltops (approved):

Issue #1188, opened 2024-03-14
Add the possibility to order recently added signs from new to old


uklomp:
The list of recently added signs is ordered from old to new. Some colleagues are fine with this, others prefer the order to be from new to old. Could you add this feature, to click on 'creation date' and change the order?

susanodd:
Recent glosses are now shown with most recent first.

uklomp:
I asked for the option to create the possibility to change the order, not to have the order changed forever. Some colleagues prefer the one option, others the other. Is that possible?

susanodd:
My hesitancy to make this have full functionality has to do with implementation.

At the moment, the code that implements this is a short simple 33 line function in `dictionary.views`

To add the facility to sort, it will need to become a class view in `dictionary.adminviews` in order to receive url requests from the template (page you see)

I can implement this, of course. But it needs much more code. I implemented the shortest solution.

If I go ahead and add a "sort" button, it would make more sense to just add a more full "time" feature, to be able to change the number of days since creation as well as sort order.

I can do it cheaply by making two urls for the two different sort orders and a simple button that goes to the "opposite" one.

susanodd:
@uklomp I am working on this.
But one flaw is that when you actually start using the recently added glosses on the live server, this interferes with the API.
Since it is uploading videos etc for these glosses....
Signbank blocks the browser.

Woseseltops:
This is what @susanodd built atm, which I think can merged:

![Image](https://github.com/user-attachments/assets/d9b2753f-d972-4bd8-a8da-723519a446d6)

I do propose the UI could be more minimal and closer to the original proposal:

For timestamp column, just use both
Toggle the sort by clicking the column label
Make the number of days a small dropdown in the corner

susanodd:
Just a short comment. I had lots of trouble formatting the stupid radio buttons. I got it to work using jQuery and self variables in the view to store the settings.

I will check out the suggested code improvement.

However, showing both columns would maybe be weird since the "retrieval" is only on either the creation date otherwise the last updated date.

A potential problem on the live server is that the API is rapidly updating glosses, which causes the recently added signs to encounter a locked database...

PR #1431, opened 2024-12-12
#1188: Class for Recent Gloss List View

susanodd:
This branch creates a class and list view for recent glosses.
It includes radio buttons to set the time requirements, which are kept as view self parameters so they remain set in the user interface:
- Chronological order, Reverse chronological order
- 7 days, 14 days, 30 days, 60 days, 90 days
- Creation Date, Last Updated

(the development servers are filled with other branches, so here's a local screenshot; my database is older)

recent-signs-toggles-timeframe-setting


susanodd:
This is working now.

Only thing left is to show alternate showing column "creation date" or "last updated" in the results list, depending on choices.

Woseseltops (commented):
Functionality
--------------
I think the UI could be more minimal and closer to the original proposal:
* For timestamp column, just use both
* Toggle the sort by clicking the column label
* Make the number of days a small dropdown in the corner

Code
------
Small comment (so up to you whether you agree): I think this

```python
RECENTLY_ADDED_TIMEDELTA_60_DAYS = datetime.timedelta(days=60)
RECENTLY_ADDED_TIMEDELTA_30_DAYS = datetime.timedelta(days=30)
RECENTLY_ADDED_TIMEDELTA_14_DAYS = datetime.timedelta(days=14)
RECENTLY_ADDED_TIMEDELTA_7_DAYS = datetime.timedelta(days=7)
```

could be

```python
RECENTLY_ADDED_TIME_DELTAS = {days: timedelta(days=days) for days in [60, 30, 14, 7]}
```

... but perhaps just using `datetime.timedelta(days=7)` instead of a constant is just as easy to read and understand.

susanodd:
Yes, there is a lot of redundancy in those time delta "dynamic constants".

<details>

For the time delta's I had a bit of trouble with them. In the (dynamic) form field, they are: model values as the value, and strings as the "radio label". That's why the code looks so weird. It also (Python Editor) complained about the scoping of the variables in the form. That may have something to do with dynamic versus static.

I'll try using a dictionary for the values instead of normal variables. Or just code. (Locally, since I don't know exactly why it was complaining about the original code with fewer variables I just had to repeat them because they were out of scope for some reason.) Perhaps the values in the actual choice list need to be constants (variables that have already been evaluated) in the choice list, not expressions to evaluate (because of lazy evaluation). Since the expressions are evaluated at compile time, but the form is created at runtime.

susanodd:
The improvements have been made and work correctly.

I did not move the radio buttons, they are still above. I do not want to make a pull-down list because it's kind of confusing what it is actually doing. The way it's like a header displaying all the settings in a visible way at the top.

(You will see what I mean if you start using it. There are two columns now. But you are actually filtering on "creation" or on "last updated" then seeing both columns. To me this is confusing. That's why I want the settings shown above, not as column header sort buttons.)

Issue #1430, opened 2024-12-12
Value Error on viewing NME videos with direct URL when Gloss WebView is set on True


rem0g:
https://signbank.cls.ru.nl/dictionary/protected_media/glossvideo/NGT/30/300-B-48928_nme_0.mp4

I am getting value error when viewing the video on Incognitio mode or obtaining URL via signCollect intranet.

```
Environment:


Request Method: GET
Request URL: http://signbank.cls.ru.nl/dictionary/protected_media/glossvideo/NGT/30/300-B-48928_nme_0.mp4

Django Version: 4.2.16
Python Version: 3.12.2
Installed Applications:
('colorfield',
 'modeltranslation',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.sites',
 'django.contrib.messages',
 'django.contrib.admin',
 'django.contrib.admindocs',
 'django.contrib.staticfiles',
 'corsheaders',
 'reversion',
 'tagging',
 'guardian',
 'bootstrap3',
 'django_summernote',
 'signbank.dictionary',
 'signbank.feedback',
 'signbank.pages',
 'signbank.attachments',
 'signbank.video')
Installed Middleware:
('corsheaders.middleware.CorsMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.locale.LocaleMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'signbank.pages.middleware.PageFallbackMiddleware',
 'reversion.middleware.RevisionMiddleware',
 'django.middleware.common.CommonMiddleware')



Traceback (most recent call last):
  File "/var/www/env/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/var/www/env/lib/python3.12/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1435, opened 2024-12-14
Fixes to public view protected media for new video types

susanodd:
#1403, #1419, #1430: Copied fixes to public view protected media to its own branch.

Since links to the new video types are already being used publically.

<details>
When they're ready, the new perspective NME video paths will also need to be added.

I'm certain the code of the pattern matching function on filenames can be made fancier, but it works as it is.
Dit is recht toe recht aan.

### DIFF
--- a/signbank/dictionary/views.py
+++ b/signbank/dictionary/views.py
@@ -2261,26 +2261,61 @@ def info(request):
     return HttpResponse(json.dumps(user_datasets_names), content_type='application/json')
 
 
+def extract_glossid_from_filename(filename):
+    file = os.path.basename(filename)
+    filename_without_extension = file.split('.')[0]
+
+    if filename_without_extension.endswith('_small'):
+        m = re.search(r".+-(\d+)\_small$", filename_without_extension)
+        gloss_pk = m.group(1)
+    elif filename_without_extension.endswith('_left'):
+        m = re.search(r".+-(\d+)\_left$", filename_without_extension)
+        gloss_pk = m.group(1)
+    elif filename_without_extension.endswith('_right'):
+        m = re.search(r".+-(\d+)\_right$", filename_without_extension)
+        gloss_pk = m.group(1)
+    elif re.search(r"\_nme_\d+$", filename_without_extension):
+        m = re.search(r".+-(\d+)\_nme_\d+$", filename_without_extension)
+        gloss_pk = m.group(1)
+    else:
+        gloss_pk = int(filename.split('.')[-2].split('-')[-1])
+
+    return gloss_pk
+
+
 def protected_media(request, filename, document_root=WRITABLE_FOLDER, show_indexes=False):
 
     if not request.user.is_authenticated:
 
         # If we are not logged in, try to find if this maybe belongs to a gloss that is free to see for everbody?
         (name, ext) = os.path.splitext(os.path.basename(filename))
-        if 'handshape' in name:
+        if 'annotatedvideo' in filename:
+            # check that the sentence exists
+            try:
+                file = os.path.basename(filename)
+                sentence_pk = int(file.split('.')[0])
+            except IndexError:
+                return HttpResponse(status=401)
+
+            lookup_sentence = AnnotatedSentence.objects.filter(pk=sentence_pk).first()
+            if not lookup_sentence:
+                HttpResponse(status=401)
+            pass
+        elif 'handshape' in name:
             # handshape images are allowed to be seen in Show All Handshapes
             pass
         else:
             try:
-                gloss_pk = int(filename.split('.')[-2].split('-')[-1])
-            except IndexError:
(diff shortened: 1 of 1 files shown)


vanlummelhuizen:
@susanodd I think your code is working as should be. Again I took some time to make your code more condensed, but still very maintainable. See below. Could you test whether this still works as should be?

<details>
<summary>Git diff of the refactor</summary>

```diff
diff --git a/signbank/dictionary/views.py b/signbank/dictionary/views.py
index 4cdc7efd..978ec4c2 100644
--- a/signbank/dictionary/views.py
+++ b/signbank/dictionary/views.py
@@ -2261,26 +2261,14 @@ def info(request):
     return HttpResponse(json.dumps(user_datasets_names), content_type='application/json')
 
 
-def extract_glossid_from_filename(filename):
-    file = os.path.basename(filename)
-    filename_without_extension = file.split('.')[0]
-
-    if filename_without_extension.endswith('_small'):
-        m = re.search(r".+-(\d+)\_small$", filename_without_extension)
-        gloss_pk = m.group(1)
-    elif filename_without_extension.endswith('_left'):
-        m = re.search(r".+-(\d+)\_left$", filename_without_extension)
-        gloss_pk = m.group(1)
-    elif filename_without_extension.endswith('_right'):
-        m = re.search(r".+-(\d+)\_right$", filename_without_extension)
-        gloss_pk = m.group(1)
-    elif re.search(r"\_nme_\d+$", filename_without_extension):
-        m = re.search(r".+-(\d+)\_nme_\d+$", filename_without_extension)
-        gloss_pk = m.group(1)
-    else:
-        gloss_pk = int(filename.split('.')[-2].split('-')[-1])
-
-    return gloss_pk
+def extract_glossid_from_filename(filename: str) -> int | None:
+    filename_without_extension, ext = os.path.splitext(os.path.basename(filename))
+    try:
+        if m := re.search(r".+-(\d+)_(small|left|right|nme_\d+)$", filename_without_extension):
+            return int(m.group(1))
+        return int(filename.split('.')[-2].split('-')[-1])
+    except (IndexError, ValueError) as e:
+        return None
 
 
 def protected_media(request, filename, document_root=WRITABLE_FOLDER, show_indexes=False):
@@ -2305,14 +2293,8 @@ def protected_media(request, filename, document_root=WRITABLE_FOLDER, show_index
             # handshape images are allowed to be seen in Show All Handshapes
             pass
         else:
-            try:
-                gloss_pk = extract_glossid_from_filename(filename)
-                glosspk = int(gloss_pk)
-            except (IndexError, ValueError):
-                return HttpResponse(status=401)
-
-            lookup_gloss = Gloss.objects.filter(pk=glosspk, archived=False, inWeb=True)
-            if not lookup_gloss.count() == 1:
+            glosspk = extract_glossid_from_filename(filename)
+            if not Gloss.objects.filter(pk=glosspk, archived=False, inWeb=True).count() == 1:
                 HttpResponse(status=401)
 
         # If we got here, the gloss was found and in the web dictionary, so we can continue
```

</details>

vanlummelhuizen:
> I don't know whether you can pass None as an object pk filter? Does that type check at runtime?
> Doesn't it need to catch the exception if it's None?
> Maybe if not glosspk or ...

`Gloss.objects.filter(pk=None)`  should find nothing because pk should never be None (or NULL) in the database. It is the same as `Gloss.objects.filter(pk__isnull=True)`. If we really want to be sure, we could do `glosspk is not None and ...` in the if statement.

susanodd:
> > I don't know whether you can pass None as an object pk filter? Does that type check at runtime?
> > Doesn't it need to catch the exception if it's None?
> > Maybe if not glosspk or ...
> 
> `Gloss.objects.filter(pk=None)` should find nothing because pk should never be None (or NULL) in the database. It is the same as `Gloss.objects.filter(pk__isnull=True)`. If we really want to be sure, we could do `glosspk is not None and ...` in the if statement.

It would depend on the implementation of Django, when it does type-checking, if None is an acceptable value for int.
Sometimes it seems that Django is really lazy and does not evaluate things at all until needed.
-1 is probably better to use than None for an int.

I changed it as you suggest! That's a good catch, since there could be a gloss with an id 0, then that would be Null. (Apparently. Depending on the implementation.)
(I have no idea what the implementation is doing sometimes. When the tests are run, it does find None errors on things. Then I fix the code.)

susanodd:
I tested it out. (Ignore if not interested. The code works fine, but)

<details>

In the browser, it correctly does not show a video if the path is wrong:

`[Error] Failed to load resource: the server responded with a status of 401 (Unauthorized) (buffelmozzarella-44048__nme_0.mp4, line 0)`

But this is in the javascript console when the path is correct (and the video is shown):

`[Error] Failed to load resource: Door plug-in afgehandelde hoeveelheid (buffelmozzarella-44048_nme_0.mp4, line 0)`

Is that the browser doing that?

PR #1444, opened 2024-12-18
Upload different format .eaf files #1399

### LINKED ISSUE #1399: Annotated sentences - not able to upload from other source
@Jetske 
I tried to upload a video and an .eaf from another source of annotated sentences and Signbank doesn't seem ready for this.
It didn't manage to read the glosses, I think maybe because the gloss tier has a different name compared to the .eafs from  the Corpus NGT files, and then it didn't upload. 

image

Jetske:
Makes it possible to upload .eaf files with different tier names

closes #1399


susanodd:
@Jetske This is live now!

Issue #1446, opened 2024-12-19
possibility to search for signs with non-manual video available


uklomp:
I'd like to have the option to search for glosses with a video in the 'videos with non-manual elements' panel available. Suggestion: with the extended search, under "search by publication status and notes", there's the 'other media' row. Now there is just the 'yes' and 'no', and I think it is for video's under 'other media'? Which we don't use at the moment but maybe is used by other people or in the future. Maybe we can implement the NME option here, e.g. a 'yes: NME' and 'yes: other media' and 'no'? 

[I'll make a new issue for the following thought: Now that I'm thinking about it, it would also be good if we could search for glosses with an annotated sentence here. I know this is a separate search function now, but maybe it could be put together.]

PR #1450, opened 2024-12-20
#1446: Search for glosses with NME videos

susanodd:
I added a simple boolean choice the same is as done for normal videos.

UPDATE AFTER MAKING CHANGES PER REVIEW

@Woseseltops I modified the code with comments, as requested. See here some questions. (Motivated by your questions.)

<details>
I also modified the "normal" Has Video code, since False had a bug. (It seems the sub-queries were cancelling each other? I rewrote it so it's only one expression in the Q, not an "or". Also for NME. I removed indentation, etc to make it more clear what is being done.)

For the "False" case, probably this is a good place to compare the file name to the "correct" filename format?
Then False can also match videos with the wrong filename, in addition to ones with missing files.
But then what should True find? True finds NME videos with files that exist. But if the file has the wrong name, is it a match to True?


signbank/query_parameters.py
```diff
@@ -169,6 +169,33 @@ def apply_video_filters_to_results(model, qs, query_parameters):
         return qs.filter(Q(**{filter_id: filter_no_video}))
 
 
+def apply_nmevideo_filters_to_results(model, qs, query_parameters):
+    gloss_prefix = 'gloss__' if model in ['GlossSense', 'AnnotatedGloss'] else ''
+    filter_id = gloss_prefix + 'id__in'
+    gloss_ids = [gl.id if model == 'Gloss' else gl.gloss.id for gl in qs]
+    if 'hasnmevideo' not in query_parameters.keys():
+        return qs
+    elif query_parameters['hasnmevideo'] == '2':
```
Woseseltops, line 178: Took me quite a while to work out `2` is the machine value of the option `True`. I know we also don't explain this in other places in the code, but perhaps a quick comment `#2 = True` would be helpful

signbank/query_parameters.py
```diff
@@ -169,6 +169,33 @@ def apply_video_filters_to_results(model, qs, query_parameters):
...
+        null_videos = qs.filter(Q(**{gloss_prefix + 'glossvideo__isnull': True}))
+        gloss_ids_without_video = [gl.id if model == 'Gloss' else gl.gloss.id for gl in null_videos]
+        glossvideo_queryset = GlossVideoNME.objects.filter(gloss__id__in=gloss_ids, version=0)
+        queryset_res = glossvideo_queryset.values('id', 'videofile')
+        results = [qv['id'] for qv in queryset_res
+                   if matching_file__does_not_exist(qv['videofile'])]
+        glosses_with_missing_videofile = GlossVideoNME.objects.filter(id__in=results)
+        gloss_ids_missing_videos = [gv.gloss.id for gv in glosses_with_missing_videofile]
+        filter_no_video = Q(**{filter_id: gloss_ids_without_video}) | Q(**{filter_id: gloss_ids_missing_videos})
+        return qs.filter(Q(**{filter_id: filter_no_video}))
```
Woseseltops, line 196: Apparently, if there are no query parameters, the code filters glosses that either have no associated GlossvideoNME entry or have one where the video file is missing. I don't understand why exactly this happens when there are no parameters... wouldn't  it be better if nothing happens when there are no parameters, and execute this code if `has_no_nme` is true?

Woseseltops (commented):
I like the work overall, but I was confused about two things

signbank/query_parameters.py
susanodd: Yes, I know. The use of '2' is very old. It's used in all the Boolean choices because of the pull-down that is used in all the forms. Add a comment everywhere? This is not documented at all, that's correct.

signbank/query_parameters.py
susanodd: The above code is for the "parameters" being "3" (which is "False"). (The "-" don't care choice has already been if-elsed away....)

<details>

This is the case for False.

For me, it is ambiguous what "False" should actually do? There are Boolean choices for True (2), False (3), and - (0) don't care). (These have always been this way. "1" is being used for "Neutral" for 3-valued Boolean fields. None.)

But what does the user expect to happen if they choose False? For normal videos this ends up being that the video is missing. But for NME videos, what should this mean?

Maybe some comments will help? Yes, it is really confusing to read. The check if it's 'Gloss' is because the code is also used by the query in the Annotated Gloss and Senses search forms. (They share the code, so some need to do extra stuff in the query formulation.)

susanodd:
I added comments to the code. But as described above, it's not clear what "False" should be doing.

I tried out the normal search for normal videos where it's False, but that actually doesn't find anything. (I suspect the different cases are cancelling each other out in the code.)

I'm going to make it so that False means "has an NME video but there is no video file for it (the file is missing)".

Issue #1445, opened 2024-12-19
Improve code in get_context_data in GlossDetailView


susanodd:
As part of Public View in #1419 #1403, a template has been created `dictionary/gloss.html `and well as a new python file  for the context variables that are shared and used in both views, `context_data_gloss.py`

The phonology code that is originally used and some of the code only used by the original template `dictionary/gloss_detail.html` has not yet been cleaned up.

Improve the Phonology "loop" and separate the choice lists (used in Edit functionality) out of the loop so the shared functionality can be moved to the Python file `context_data_gloss.py` and shared.

Do the remaining code cleanup here.

PR #1454, opened 2025-01-11
#1445: Code improvement Gloss Detail context variables

susanodd:
Shared context for Public and Logged In views consolidated


Woseseltops (commented):
Seems okay to me, except for:

```python
        for (key, value) in display_choice_list.items():
            this_value = value
            static_choice_lists[key] = this_value
```

You do this a few this, why not just

```python
        for (key, value) in display_choice_list.items():
            static_choice_lists[key] = value
```

or perhaps something with a list comprehension.

Furthermore, I'm not sure about using strings as variable/field types, but that's not something you introduce in this pull request.

susanodd:
> Furthermore, I'm not sure about using strings as variable/field types, but that's not something you introduce in this pull request.

@Woseseltops can you explain what you mean here? The field names are in the settings as strings. Those are actually the names of fields in model Gloss.

Issue #1360, opened 2024-10-29
Problems sending senses to api_update_gloss


rem0g:
I am noticing a lot of errors regarding this gloss: 
 
While submitting request to update phonology I am getting error Database locked at HUREN-C. 

WHen submitting empty phonology feature such as: 

"Handshape Change": ""

The handshape change is still not modified to empty value.

rem0g:
Ok, using "-" works! Thank you very much. 

About gloss HUREN-C; 

TransactionManagementError at /dictionary/api_update_gloss/5/49249/
An error occurred in the current transaction. You can't execute queries until the end of the 'atomic' block.

This is what i have sent: 

{"Handedness": "N/A", "Strong Hand": "-", "Weak Hand": "B", "Handshape Change": "-", "Relation Between Articulators": "-", "Location": "-", "Contact Type": "-", "Movement Shape": "-", "Movement Direction": "-", "Repeated Movement": "", "Alternating Movement": "", "Relative Orientation: Movement": "-", "Relative Orientation: Location": "-", "Orientation Change": "-", "Virtual Object": "-", "Phonology Other": "-", "Mouth Gesture": "-", "Mouthing": "-", "Phonetic Variation": "-", "Senses": "{\"en\": [[\"hire\"], [\"rent, lease\"]], \"nl\": [[\"huren\"], [\"huren\"]]}"}

rem0g:
Semantic Field and Phonology Other fields dont seem to be updated too.

susanodd:
> Phonology Other fields dont seem to be updated too.
> 
> Modifing the values to the ones you mentioned didnt help, it is causing only by this gloss but it could happen at other glosses too.

Is this a newly created gloss, or an older gloss you are updating?

(If it's the latter there could be an old field inside the gloss from when the field choices used to be "character strings with machine value" instead of "object reference to FieldChoice.)

I will investigate the gloss itself.

Does your error message mention a specific line of code?

It's true that if it finds an error in the input fields, it will not do the update. (It's implemented as all or nothing, otherwise the transaction would be inconsistent.)

Do you want Handedness to be N/A ?
(Some of the query code excludes these from queries.)

susanodd:
These fields:

"Virtual Object": "-", "Phonology Other": "-", "Mouth Gesture": "-", "Mouthing": "-", "Phonetic Variation": "-", 

are strings, so you can just put empty strings (the ones that are choices need to be "-")

[...]

PR #1460, opened 2025-01-16
#1360: Moved senses field code for update gloss API earlier.

susanodd:


Woseseltops (approved):
Surprised the solution is so simple... you sure this solves it?

EDIT: upon re-reading, I notice you already mentioned it isn't solved

susanodd:
> Surprised the solution is so simple... you sure this solves it?
> 
> EDIT: upon re-reading, I notice you already mentioned it isn't solved

Yes, it only solves that the code "didn't get to the update". This solves the "wrong branch of the loop" bug.
Now it gets to actually update the senses. But that does not work.

It is difficult for me to fix this. I wrote all the code except the senses code. I can't see anything obvious that would be a bug.
I also don't know why it's getting a consistency error from the example given.

I'll try reordering the code first to make sure it's not something with the variables being reused after deleted.

Issue #1436, opened 2024-12-15
Settings DEBUG is True somehow


susanodd:
The database is somehow getting locked on the production server and it's not possible to login.

```
OperationalError at /accounts/login/
database is locked
```

On attempts to login, it is showing all the settings because DEBUG is somehow set to True on the production server.

In the base.py settings, it's set to True

It looks like the server specific are not overriding this to False.

vanlummelhuizen:
> The database is somehow getting locked on the production server and it's not possible to login.
> 
> ```
> OperationalError at /accounts/login/
> database is locked
> ```

I see lots of 'database is locked' in yesterday's logs. For a possible solution see #673 .

> On attempts to login, it is showing all the settings because DEBUG is somehow set to True on the production server.
> 
> In the base.py settings, it's set to True
> 
> It looks like the server specific are not overriding this to False.

That is because there is no actual code doing any overriding. `DEBUG = False`  is simply not in the code base right now. If there should be such a line, I don't know. @Woseseltops ?

susanodd:
It seems to be in base.py

(Is that not imported if the database is locked? Should that be False?)

It's actually done "after" it imports from server_specific (in base.py)

susanodd:
@vanlummelhuizen can you fix this bug?

The settings on the live server need to be "git stash"ed every time before a pull because there are conflicts with the master branch.
This used to not be necessary.

Moreover, DEBUG was not set to True.

susanodd:
For some reason, PyCharm needs DEBUG to be True in base.py

susanodd:
> It seems to be in base.py
> 
> (Is that not imported if the database is locked? Should that be False?)
> 
> It's actually done "after" it imports from server_specific (in base.py)

@vanlummelhuizen  (in response to his comments in closed pull request February 13.)

The quote above was on December 16

Apparently the entire line of code needs to be removed.

But why was this in the code in the first place?

Probably because we used to use the DEBUG TOOLBAR on the live server in order to observe whether the Django query had been evaluated.

vanlummelhuizen:
See https://github.com/Signbank/Global-signbank/pull/1473#issuecomment-2655832707

PR #1473, opened 2025-01-30
#1436: Set DEBUG to False in base.py settings

susanodd:
This prevents screen dumps of data structures if a command fails

### DIFF
--- a/signbank/settings/base.py
+++ b/signbank/settings/base.py
@@ -5,7 +5,7 @@ from django.utils.encoding import smart_str
 from signbank.settings.server_specific import *
 from datetime import datetime
 
-DEBUG = True
+DEBUG = False
 
 PROJECT_DIR = os.path.dirname(BASE_DIR)
 


vanlummelhuizen:
@susanodd DEBUG is False by default, so I suggest removing this line from base.py. DEBUG should then be set in server_specific.py. This helps to avoid DEBUG =True is set on the live server.

susanodd:
> @susanodd DEBUG is False by default, so I suggest removing this line from base.py. DEBUG should then be set in server_specific.py. This helps to avoid DEBUG =True is set on the live server.

It's only False now because I made it be False.

See the commit histories. I already changed it to be False, that's why it's False now by default.
Somebody had it set to True before.
But as explained in the issue, it has to be True for PyCharm to work. So locally a user should set this to True. (Needs to be documented.)

vanlummelhuizen:
> It's only False now because I made it be False.

No, see https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-DEBUG, second line below the heading.

susanodd:
> > It's only False now because I made it be False.
> 
> No, see https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-DEBUG, second line below the heading.

Huh?

I made it be False because nobody had replied on this since December.

Yes, I believe you. But couldn't you have responded last year?

PR #1474, opened 2025-01-31
Added citation image API endpoint

Woseseltops:
A new endpoint for poster images as requested by @rem0g . 

The endpoint can be called like this (the test environment will be online for 2 weeks, you can use this temperorary token to test if you want):

```python
import requests

# The URL where the video file will be sent
url = 'http://52.57.57.71/dictionary/api_update_gloss/4483/image'

# The path to your video file
image_file_path = 'VLINDERS-IN-DE-BUIK-B-48960.png'

# Auth
headers = {'Authorization': f'Bearer JZnWi1BoQvtEmzcL'}

# Open the video file in binary mode and send it in a POST request
payload = {'File': open(image_file_path, mode='rb')}
response = requests.post(url, files=payload, headers=headers)

# Print the server's response
print(response.status_code)
print(response.text)
```

When working on this, I created #1475 and #1476, tasks to be picked up once this is live.

### DIFF
--- a/signbank/api_interface.py
+++ b/signbank/api_interface.py
@@ -776,6 +776,9 @@ def api_add_video(request, gloss_id):
 @csrf_exempt
 @put_api_user_in_request
 def api_add_image(request, gloss_id):
+    """
+    Handles the upload of an image file for a specific gloss via the API, and saves it to the correct location.
+    """
 
     error_response = verify_conditions_for_file_upload(request, gloss_id)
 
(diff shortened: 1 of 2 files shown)


signbank/api_interface.py
```diff
@@ -759,4 +770,56 @@ def api_add_video(request, gloss_id):
...
+def api_add_image(request, gloss_id):
+
+    error_response = verify_conditions_for_file_upload(request, gloss_id)
+
+    if error_response:
+        return error_response
+
+    gloss = Gloss.objects.filter(id=gloss_id).first()
+
+    label = list(request.FILES.keys())[0]
```
vanlummelhuizen, line 787: Even shorter: `image_file = list(request.FILES.values())[0]`

signbank/api_interface.py
```diff
@@ -759,4 +770,56 @@ def api_add_video(request, gloss_id):
...
+    if not os.path.exists(goal_path):
+        try:
+            os.makedirs(goal_path)
+        except OSError as ose:
+            return JsonResponse({'error': f'Failed to create directory {goal_path}: {ose}'}, status=500)
+
+    # Save the file
+    try:
+        destination = File(open(goal_location.encode(sys.getfilesystemencoding()), 'wb+'))
+
```
vanlummelhuizen, line 815: Remove empty line.

vanlummelhuizen (commented):
All in all I think this quite readable code that does not have of surprises or esoteric constructions.  I would remove some newlines and comments because the code is pretty readable. 
Perhaps there is a bit of time to write a  docstring for the new function.

signbank/api_interface.py
```diff
@@ -759,4 +770,56 @@ def api_add_video(request, gloss_id):
...
+    label = list(request.FILES.keys())[0]
+    image_file = request.FILES[label]
+
+    #First check the extension
+    file_extension = os.path.splitext(image_file.name)[1].lower()
+    if file_extension not in settings.SUPPORTED_CITATION_IMAGE_EXTENSIONS:
+        return JsonResponse({'error': 'File extension not supported! Please convert to png or jpg'}, status=400)
+
+    # Construct a filename for the image, use sn if present, otherwise use idgloss+gloss id
+    if gloss.sn is not None:
```
vanlummelhuizen, line 796: This could be done in one line with a ternary operator. If the line gets too long, you could determine the file name first and then add the file extension.

susanodd:
Did you try this on anonymous user?

There's a test whether request.user is defined, but shouldn't that test whether the user is not anonymous?

`    if not request.user:`

Woseseltops:
Thanks @vanlummelhuizen , agreed with everything!
Good question @susanodd . I just checked, but this should be handled by the decorator function (the one all API functions have) already!

Issue #1484, opened 2025-02-04
Translations


susanodd:
Some template code is evading the translations.

Make sure they have been generated (on the master branch), fuzziness disambiguated ('po' file), and compiled ('mo' on the servers) in order to include up-to-date template filenames and line numbers in the translation 'po' file.

I did this on all three development server and the production server yesterday.

- What remains is to generate them again and make sure there are translations for all the recent text in the templates. And then repeat the remaining steps.

PR #1485, opened 2025-02-05
BUG FIXES gloss update API -- stop putting empty things in revision history

susanodd:
Many small bug fixes. Bugs found during trial and error of all the different fields

- Code for multilingual gloss update translation code #1484

- Renamed functions to make it clear what is being done. CODE IMPROVEMENT

- Added status code 400 to error branches to make sure failure of the operation is observed by caller. CODE IMPROVEMENT

- Fail for transaction error on save BUG FIX

- Repaired code that translated API input json "left hand side". BUG FIX

- _("Senses") translation to Dutch is hard-coded, since it does not work to retrieve the verbose name from the Gloss field. Unknown why, maybe because it's a verbose name on a relation. This used to work. Maybe Django updates.

- Normalise Boolean values (this was not working at all, bug was not reported) BUG FIX

- Only update those fields where the new value differs from the original #1486

- Improved code for comparison of stored and new value depending on type. BUG FIX

### DIFF
--- a/signbank/csv_interface.py
+++ b/signbank/csv_interface.py
@@ -1141,6 +1141,21 @@ def csv_focusgloss_to_minimalpairs(focusgloss, dataset, language_code, csv_rows)
     return csv_rows
 
 
+def normalize_boolean(gloss_field, new_boolean, original_boolean):
+    # return 'True', 'False', 'None' depending on field type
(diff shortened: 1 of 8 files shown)


Woseseltops (approved):
Couldn't find anything to complain about

Issue #1492, opened 2025-02-07
The video overwriting/relinking problem still exists if you rename a gloss


Woseseltops:
The problem we analyzed in #1453 is apparently still present if you rename glosses. @rem0g reports:

> Glos AAP was aangepast naar AAP-A, video links en midden en NMM video zijn dan weg. Er blijft dan alleen video rechts over en deze zit ook in NMM index 0

susanodd:
The original videos of a deleted gloss are not deleted.

This should only apply if you rename the Lemma, not the Annotation. The video filename includes the name of the Lemma. (@vanlummelhuizen to solve this, I thought he did solve this?)

I told you in the office on Thursday, there are still many NME videos that point to the normal video.

susanodd:
I copied the production database to signbank-susan where the video admin commands live.

There are still 37 glosses with a WRONG NAME for an NME video that points to the normal video.

vanlummelhuizen:
Fixed it. See #1495.

Issue #1453, opened 2024-12-23
Videos randomly disappearing


rem0g:
Some glosses has disappeared videos for example: 

https://signbank.cls.ru.nl/dictionary/gloss/47361

https://signbank.cls.ru.nl/dictionary/gloss/47475

Some glosses has wrong still image from the video. 

This is happening everywhere, can you take a look into that?

rem0g:
At Glos AFBROKKELEN (https://signbank.cls.ru.nl//dictionary/gloss/46883), I am seeing this: 


      {
                    "ID": "211108",
                    "Index": "0",
                    "Description: Dutch": "afbrokkelen",
                    "Description: English": "crumble",
                    "Link": "https://signbank.cls.ru.nl//dictionary/protected_media/glossvideo/NGT/AF/AFBROKKELEN-46883.mp4"
                },
                {
                    "ID": "208048",
                    "Index": "1",
                    "Description: Dutch": "afbrokkelen",
                    "Description: English": "crumble",
                    "Link": "https://signbank.cls.ru.nl//dictionary/protected_media/glossvideo/NGT/AF/AFBROKKELEN-46883.mp4"
                }
                
                
                Both has same link and that should not be the case.

susanodd:
@rem0g the university is closed this week and next. We won't be able to work on this until then.
We are all on vacation now.

susanodd:
Can you increase the amount if time between updating/uploading and retrieving?

Like maybe increase it to 10 minutes? There was previously a problem that the time between operations was too frequent.
With videos in could be that a previous operation was not completed before a new one is performed. (It makes a difference if the file system is continuously writing new files. The transactions to create objects are faster than the file system.

susanodd:
I had this problem locally on my own computer using iCloud for storage.
But I could not repeat it on the real server.
It was that the files had not been completely copied to iCloud. So they "disappeared" like that. They were actually on disc, but had names with "." before the filename, so not visible. This was on Apple's Unix.

rem0g:
Yes I could do that but i would rather for Signbank side do something about the transactions, for example for incoming transactions I always enqueue them in a list instead of executing them immediately. When the server has finished processing a certain transaction then it could handle the next one. 

Would that be something you can work on?

susanodd:
@rem0g that sounds like an interesting approach. I will discuss that with @vanlummelhuizen how to implement that. He is the Django expert. A queuing mechanism.

susanodd:
[...]

PR #1495, opened 2025-02-10
fix: Applied fix for #1453; #1492

vanlummelhuizen:


susanodd:
Signbank-dev is free if you want to experiment there.

Issue #1477, opened 2025-02-01
Commands to remove redundant entries from Gloss Revision History


susanodd:
Add Admin and Gloss Edit commands to cleanup repetitive and empty entries from the Gloss Revision History.

This is live on `signbank-dev`. It's totally awesome!

PR #1499, opened 2025-02-10
#1477: Template command to cleanup revision history

susanodd:
- Fixed display to show revision history in the current interface language being used, including the field values (objects in the revision history have strings for everything, these are in the interface language being used at the time that they were stored)
- Template command to clean up the revision history (removes empty entries, removes repeated duplicates leaving one)

This is on signbank-dev.

Try searching on "recently added glosses", then look at their Revision History (the glosses with lots of empty stuff in the history), then press "Clean Up". Presto! This is totally awesome!

Note, pull request #1485 is also needed, since it stops storing these empty things in the revision history.

### DIFF
--- a/signbank/dictionary/gloss_revision.py
+++ b/signbank/dictionary/gloss_revision.py
@@ -1,12 +1,190 @@
 
-from signbank.dictionary.models import Gloss, GlossRevision
+from signbank.dictionary.models import Gloss, GlossRevision, Language, FieldChoice, Handshape
 import datetime as DT
 from django.shortcuts import render, get_object_or_404, redirect
 from django.contrib.auth.decorators import permission_required
-from django.http import HttpResponseRedirect
+from django.http import HttpResponseRedirect, JsonResponse
 from signbank.dictionary.update import okay_to_update_gloss
 from django.urls import reverse
 from django.conf import settings
+from signbank.dictionary.translate_choice_list import check_value_to_translated_human_value
+from django.utils.translation import gettext_lazy as _, activate, gettext
+from signbank.dictionary.context_data import get_selected_datasets
+from django.core.exceptions import PermissionDenied, ObjectDoesNotExist, MultipleObjectsReturned
+from signbank.csv_interface import normalize_field_choice
+from django.db.models import Q, Count, CharField, TextField, Value as V
+
+
+def get_field_choice_from_name(fieldname, value, language_codes):
+    # try to get a matching field choice in one of the languages
+    if fieldname not in Gloss.get_field_names():
+        return value
+    field = Gloss.get_field(fieldname)
+    if not hasattr(field, 'field_choice_category'):
+        return value
+    field_choice_category = field.field_choice_category
+    for language_name_field in language_codes:
+        # next_language_name = 'name_'+language_code
+        fieldchoice = FieldChoice.objects.filter(Q(**{'field': field_choice_category,
+                                                      language_name_field: value})).first()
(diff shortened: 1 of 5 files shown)


signbank/dictionary/gloss_revision.py
```diff
@@ -0,0 +1,233 @@
...
+            language_2char = revision.field_name[-2:]
+            language = Language.objects.get(language_code_2char=language_2char)
+            revision_verbose_fieldname = gettext('Annotation ID Gloss') + " (%s)" % language.name
+        elif revision.field_name == 'archived':
+            revision_verbose_fieldname = gettext("Deleted")
+        elif revision.field_name == 'restored':
+            revision_verbose_fieldname = gettext("Restored")
+        else:
+            revision_verbose_fieldname = gettext(revision.field_name)
+            # print('fall through: ', revision.field_name, revision_verbose_fieldname)
```
Woseseltops, line 101: I think the easy cases here could benefit from a translation dict. Something like this

```python
        translation_dict = {
            'sequential_morphology': gettext("Sequential Morphology"),
            'simultaneous_morphology': gettext("Simultaneous Morphology"),
            'blend_morphology': gettext("Blend Morphology"),
            'archived': gettext("Deleted"),
            'restored': gettext("Restored"),
        }

        if revision.field_name in translation_dict:
            revision_verbose_fieldname = translation_dict[revision.field_name]
```

signbank/dictionary/gloss_revision.py
```diff
@@ -0,0 +1,233 @@
...
+                delete_command = gettext('Delete')
+                field_name_qualification = ' (' + delete_command + ')'
+            elif revision.new_value and not revision.old_value:
+                # this translation exists in the interface of Gloss Edit View
+                add_command = gettext('Create')
+                field_name_qualification = ' (' + add_command + ')'
+            else:
+                # this translation exists in the interface of Gloss Edit View
+                add_command = gettext('Update')
+                field_name_qualification = ' (' + add_command + ')'
```
Woseseltops, line 155: Here too I see a lot of duplication. Perhaps, at the very least, add the `()` once at the end instead of separately for each condition.

signbank/dictionary/gloss_revision.py
```diff
@@ -0,0 +1,233 @@
...
+                duplicate_revisions[revision.field_name].append(revision)
+        else:
+            if not tuple_updates:
+                tuple_updates.append((revision.field_name, revision.old_value, revision.new_value))
+            else:
+                (last_field, last_old_value, last_new_value) = tuple_updates[-1]
+                if last_field == revision.field_name and last_old_value == revision.old_value and last_new_value == revision.new_value:
+                    tuple_update_already_seen.append(revision)
+                else:
+                    tuple_updates.append((revision.field_name, revision.old_value, revision.new_value))
```
Woseseltops, line 222: Three times a nested `else` is difficult to read, I wonder if this code could be less indented. Every time you do an `else`, can't you just do `continue` after the conditions above so it becomes the main path?

signbank/dictionary/gloss_revision.py
susanodd: Yes! This code kind of evolved and we just kept adding stuff to it. Believe it or not, there was even a repeated if-else clause!

I'll work on this....

It's computing backwards based on the "strings" that have been put in the revision history.
Probably some comments are needed as well...

susanodd:
Changes as requested.

Issue #1497, opened 2025-02-10
Document that the API urls are multilingual


susanodd:
- The API urls are multilingual. They have already been multilingual since last October, because the UvA users use the Dutch phonology.

- Because the Wiki examples page was removed, it is not evident on Swagger that the urls are multilingual, because Swagger does not accommodate the Accept-Language header.

- It is not feasible to add the Accept-Language code to Swagger, without hacking Swagger. (There are examples of this on the web, but those users are not using Django.)

The easiest, least amount of English and Json typing is to simply do it this way:

Issue: Document that the API also works with gloss fields and field choices in Dutch.

Propose to do it this way:

- add an _optional_ language code to the url itself as shown below:

This can be done simply with an illustration of one url:

`get_gloss_data/{datasetid}/{glossid}/`
`get_gloss_data/{datasetid}/{glossid}/(en|nl|zh-hans)/`

In the first case, the exact same url is used. This defaults to English, unless the user is using the Accept-Language in their request.
This is what happens with Swagger, which uses English.

For the second case, the Swagger example is added:

`get_gloss_data/{datasetid}/{glossid}/nl/`

For this, the Dutch fields and example output with Dutch phonology can be shown. This can also be done for Chinese.

PR #1503, opened 2025-02-13
#1497: Multilingual api urls with optional language code argument

susanodd:


Woseseltops (commented):
They only thing that strikes my eye is the name `abstract_machine.py`, which doesn't make it clear (for me) what the functions inside have in common. 

I realize you didn't create the file for this issue, but perhaps you can use this issue to clear this up a bit. Could be just renaming the file, but perhaps it also requires moving function to other (new?) files?

PR #1525, opened 2025-03-21
Introducing perspectives for NME videos

Woseseltops:
When using the UI, you now have this dropdown:

image

And from the API, you do it like this:

```python

url = 'https://signbank.cls.ru.nl/dictionary/api_create_gloss_nmevideo/5/4483/'
payload = {'Index': 0, 'Description (Dutch)': 'hi', 'Description (English)': 'hi'}
files = {'left': open(video_a_file_path, mode='rb')} #Notice the key

response = requests.post(url, data=payload, files=files, headers=headers)
```

Changing the perspective

```python
url = 'https://signbank.cls.ru.nl/dictionary/api_update_gloss_nmevideo/5/4483/268392/'
payload = {'Index': 0, 'Description (Dutch)': 'hi', 'Description (English)': 'hi','Perspective': 'right'}

response = requests.post(url, data=payload, headers=headers)
```

When there are multiple NME videos from different perspective, the UI does not show the perspective video viewer yet. Will create a separate issue for that.

### DIFF
--- a/conf/locale/nl/LC_MESSAGES/django.po
+++ b/conf/locale/nl/LC_MESSAGES/django.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2025-03-14 11:15+0100\n"
+"POT-Creation-Date: 2025-03-26 15:14+0100\n"
 "PO-Revision-Date: 2018-03-08 18:54+0100\n"
 "Last-Translator: susanodd\n"
 "Language-Team: Dutch <LL@li.org>\n"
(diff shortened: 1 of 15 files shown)


susanodd:
- for the display in the Gloss Detail View, you may want to create a `get_perspective_display` method instead of retrieving the field. This would allow dynamic translations to be shown. To get them into the choice list, they need to be added dynamically to the choice list. (From memory of doing other issues. There are quite a few of this kind of method. Although I do see some like the relation role that don't have any translations, so perhaps this could be a new issue.)

vanlummelhuizen (changes requested):
I did not test, but it looks good to me. I made a minor comment about the code. See below.

I also noticed you added quite a few blank lines, especially after `if`, `else` and `for`. In my view they are unnecessary.

Woseseltops:
Thanks, both of you @susanodd @vanlummelhuizen . I feel your reviews really lead to better code :)

Issue #1527, opened 2025-03-25
Get rid of *'s in import blocks


vanlummelhuizen:
In the import block, replace wildcard asterisks (*) with the actual things that should be imported. This should solve some of the grayed out imports while they seem to be used after all. See also a quick explanation here: https://stackoverflow.com/questions/2386714/why-is-import-bad

Issue #1515, opened 2025-02-27
Clean up Signbank imports


Woseseltops:
For several files, the imports are quite messy. Random example:

```python
import json

from signbank.dictionary.models import *
from django.db.models import FileField
from django.core.files.base import ContentFile, File
from tagging.models import Tag, TaggedItem
from signbank.dictionary.forms import *
from django.utils.translation import override, gettext_lazy as _, activate
from signbank.settings.server_specific import LANGUAGES, LEFT_DOUBLE_QUOTE_PATTERNS, RIGHT_DOUBLE_QUOTE_PATTERNS

from django.contrib import messages

from django.core.exceptions import ObjectDoesNotExist

from django.shortcuts import get_object_or_404

from django.contrib.auth.decorators import permission_required
from django.db import DatabaseError, IntegrityError
from django.db.transaction import TransactionManagementError

from django.http import HttpResponse, HttpResponseRedirect, HttpResponseForbidden, HttpResponseBadRequest, JsonResponse
from signbank.settings.server_specific import VIDEOS_TO_IMPORT_FOLDER
import zipfile
from django.urls import reverse, reverse_lazy
import urllib.request
import tempfile
import shutil
import os
from signbank.video.models import GlossVideo, GlossVideoHistory
from django.http import StreamingHttpResponse
from django.contrib.auth.models import Group, User
from signbank.zip_interface import *
```

Let's use this issue to make all files more orderly.

On top of this, here and there are imports that cannot go at the top because of circular import problems. I feel there's a more elegant way to solve that problem... why are the circular imports there in the first place?

vanlummelhuizen:
> Let's use this issue to make all files more orderly.

Good idea. But first determine what is orderly and what is not. My idea:

1. All general packages (os, json, etc.)
2. All Django packages (from django import ...)
3. All third party Django packages (e.g from guardian.shortcuts import ...)
4. Project packages (e.g. from signbank.tools import ...)

vanlummelhuizen:
> On top of this, here and there are imports that cannot go at the top because of circular import problems. I feel there's a more elegant way to solve that problem... why are the circular imports there in the first place?

I think they are to that big of a problem right now. I also think trying to solve circular imports in this code base will cost us too much time right now. I'd rather try to clean up clearly delimited code blocks or files, or do a clean up for a specific topic (e.g. handling media). We should try to solve circular import within that clean up process, not as a separate clean up process.

Woseseltops:
[...]

PR #1530, opened 2025-03-27
#1527, #1515: Specific imports instead of star, remove unused

susanodd:
THIS IS ON SIGNBANK-TEST NOW. ALL OF THE TESTS PASS.

The rest of the imports have been moved upwards, with the exception the circular ones shown in #1539.

Can we get this branch reviewed asap. It's a pain to merge it with other branches. (The imports kind of cascade and it grabs the things it needs from wherever it can before deciding it needs to be declared in the Python file in question.)

This involved lots more coding than merely replacing the "stars"!

It also revealed some potential bugs (that nobody has reported, see the documentation below)

<details>

In making the imports explicit instead of using star, it kind of became necessary to remove unused imports and reorder them, because imports it used to be obtaining from other files were no longer imported!

What do do when imports aren't needed?
<details>
1. Question: If a Python file contains functions that take objects as parameters, for example "gloss", is it necessary to import the model Gloss? The function may retrieve fields from the argument, such as "gloss.lemma.dataset"? What needs to be imported? Since the parameter is used, it type checks without importing anything.
- Solution: I added type assertions so that the models can be explicitly imported (without being marked as unused by PyCharm)

2. Type checking in related_objects.py:

Function okay_to_move_gloss calls gloss_exists_in_dataset, it gves the following on the dataset argument (can this only be confirmed at runtime?)

- `Type 'Dataset' doesn't have expected attribute 'lemmaidgloss_set'`
</details>
Feel free to browse what has been changed as confirm whether it also follows a desired order according to #1515

I'm only doing the files that have "`import *`" in them...

Settings
<details>
But what do we think of:

- from signbank.settings.base import *
- from django.conf.urls import *

There are also some imports of settings in the settings files themselves....

I modified the use of (non-Django) settings that are just Signbank abbreviations for things, so that they are imported.
There are some Signbank settings that are in base for some reason, like DISABLE_MOVING_THUMBNAILS_ABOVE_NR_OF_GLOSSES. I moved some to default, since it has nothing to do with Django.
</details>
PyCharm warnings video models:
<details>
Some specific PyCharm type checking errors for `video/models.py` (these look like errors, but are not this issue):

- Unresolved attribute reference 'path' for class 'FileField'
- Unresolved attribute reference 'url' for class 'FileField'
- Package containing module 'ffmpeg' is not listed in the project requirements
- Unresolved attribute reference 'poster_path' for class 'ExampleVideo'
- Unresolved attribute reference 'get_dataset' for class 'OneToOneField'
- Expected type 'int', got 'IntegerField' instead

</details>
Missing requirements:

<details>
From `dictionary/admin.py`:

- Package containing module 'reversion' is not listed in the project requirements
- Redeclared 'UserAdmin' defined above without usage (This is imported, but also defined)

</details>
I found some bugs in dictionary/update.py

<details>
Because the imports are explicit now, PyCharm is complaining about many more things than it used to. The following looks like an undocumented bug.

The "else" clause inside of the try is messed up. (It gives a warning: Parameter 'message' unfilled")
But what is this code doing?

https://github.com/Signbank/Global-signbank/blob/02869891bcf830cf2fcc0dc0c6bd683dca5b1277/signbank/dictionary/update.py#L2609-L2622

This is also messed up:

`def update_annotation_idgloss(request, gloss, field, value):`

The function uses "original_value" which is undefined. It looks like the code used to be in a different order.
I changed it to BadRequest. (The code detects that the language parameter is wrong, but then wants to do nothing.)

New code (these used to return "original_value", which is undefined in the code). In the first one, there is not original value because the language is wrong:

https://github.com/Signbank/Global-signbank/blob/02869891bcf830cf2fcc0dc0c6bd683dca5b1277/signbank/dictionary/update.py#L1181-L1190

https://github.com/Signbank/Global-signbank/blob/02869891bcf830cf2fcc0dc0c6bd683dca5b1277/signbank/dictionary/update.py#L1197-L1200

</details>


susanodd:
@vanlummelhuizen I can see that you commented to move an import to the top, but it does not show up anywhere in the files view, so I don't know where it applies?

vanlummelhuizen:
@susanodd I think this is ready to be merged. There is conflict with 'master' though. Can you solve that?

susanodd:
Ouch, conflicts again.

PR #1536, opened 2025-04-02
#1406: Gloss Videos List View, Search Form, template

susanodd:
- This pull request sets up a Template, List View, and Search Form for Gloss Videos of a Dataset

- This is completed. It's on signbank-dev.

- To navigate to the page, it's linked to under Manage Datasets -> (Dataset) Manage Video Storage

https://signbank-dev.cls.ru.nl/datasets/glossvideos/5

- The new page replaces the existing page.

<details>
(The old template file, class, and url have not been removed yet, they are just unnavigable now. If you type in the former url you can still compare the old page to the new, which is now linked to as described above.)
</details>

- New: Search functionality
- New: Delete buttons for cleaning up excessive backup files. (Moves them to trash folder as is done in ADMIN.) (Note, when you try this on the development server, it merely deletes the objects if there are no video files.)

- I added a limit of 1000 objects to the search and a message to refine your query to retrieve fewer results.
This is necessary because there are more than 100,000 backup videos for NGT.  I added a count column for the backups.

- For filtering on Wrong Filename this is kind of backfiring because most of the backup filenames are wrong, so overload the number of results. The user will need to add an annotation search to decrease results.

<details>
(The display chokes if there are too many results, hence the limit. The ADMIN only retrieves 100 results at a time. But that is too complicated to implement something like that here.)
</details>

The code revises the setup on the existing page with a search functionality.

The "kind" of videos can be queried as well as the "annotations".

For the display of the results, the same setup is used as the existing page, a separate scroll list for each type of video, ordered by Gloss Lemma.
<details>
(That may not be true anymore. There were lots of problems because sorting by lemma was causing duplicate objects in the query.)
<details>

Because the kind of videos are shown in separate sections, to keep the query retrieval simple, the videos are "partitioned" with the gloss as the gloss videos are examined per kind of video.

(See the code.)

<details>
I haven't come up with an alternative "data structure" combined with "query" for efficiency.
(You'll see what I mean if you look at the code. All of the objects are of type GlossVideo, but they are further categorised by subclasses. I made it as it is in order to put the videos in separate display areas, but only iterate (fetch all the video paths and make a string from them for display in the tables.)
This is to prevent all of the different kinds of videos being in the same "gloss row".
</details>

### DIFF
--- a/signbank/dataset_operations.py
+++ b/signbank/dataset_operations.py
@@ -367,36 +367,42 @@ def update_gloss_video_backups(request, glossid):
     return JsonResponse({'videos': list_videos})
 
 
-def wrong_filename(videofile, nmevideo, perspective, version):
+def has_correct_filename(videofile, nmevideo, perspective, version):
     if not videofile:
         return False
     video_file_full_path = Path(settings.WRITABLE_FOLDER, videofile)
     if nmevideo is not None:
         filename_is_correct = filename_matches_nme(video_file_full_path) is not None
-        return not filename_is_correct
+        return filename_is_correct
     elif perspective is not None:
         filename_is_correct = filename_matches_perspective(video_file_full_path) is not None
-        return not filename_is_correct
+        return filename_is_correct
     elif version > 0:
         filename_is_correct = filename_matches_backup_video(video_file_full_path) is not None
-        return not filename_is_correct
+        return filename_is_correct
     else:
         filename_is_correct = filename_matches_video(video_file_full_path) is not None
-        return not filename_is_correct
+        return filename_is_correct
 
 
 def wrong_filename_filter(glossvideos):
     filenames = []
     queryset_tuples = glossvideos.values('id', 'videofile', 'glossvideonme', 'glossvideoperspective', 'version')
     for qv in queryset_tuples:
-       if wrong_filename(qv['videofile'],
-                            qv['glossvideonme'],
-                            qv['glossvideoperspective'], qv['version']):
+       if not has_correct_filename(qv['videofile'],
+                                   qv['glossvideonme'],
+                                   qv['glossvideoperspective'], qv['version']):
            filenames.append(qv['id'])
     return filenames
 
 
-def gv_preface(glossvideo):
+def file_display_preface(glossvideo):
+    """
+    This function yields extra information to be displayed in front of the video filename
+    For primary videos: version; for perspective videos: left or right; for NME videos order_(left|center|right)
+    :param glossvideo: GlossVideo
+    :return: string pattern
+    """
     if hasattr(glossvideo, 'glossvideonme'):
         nme_video = glossvideo.glossvideonme
         nme_perspective = nme_video.perspective if nme_video.perspective else 'center'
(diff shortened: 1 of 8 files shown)


signbank/dataset_operations.py
```diff
@@ -354,7 +360,144 @@ def update_gloss_video_backups(request, glossid):
...
+
+
+def wrong_filename_filter(glossvideos):
+    filenames = []
+    queryset_tuples = glossvideos.values('id', 'videofile', 'glossvideonme', 'glossvideoperspective', 'version')
+    for qv in queryset_tuples:
+       if wrong_filename(qv['videofile'],
+                            qv['glossvideonme'],
+                            qv['glossvideoperspective'], qv['version']):
+           filenames.append(qv['id'])
```
Woseseltops, line 394: To me associating wrong with `True` and correct with `False` feel counterintuitive... I'd prefer something like

```python
if not filename_is_correct(args):
```

signbank/dataset_operations.py
```diff
@@ -354,7 +360,144 @@ def update_gloss_video_backups(request, glossid):
...
+
+def gv_preface(glossvideo):
+    if hasattr(glossvideo, 'glossvideonme'):
+        nme_video = glossvideo.glossvideonme
+        nme_perspective = nme_video.perspective if nme_video.perspective else 'center'
+        return str(nme_video.offset)+'_'+nme_perspective
+    if hasattr(glossvideo, 'glossvideoperspective'):
+        perspective_video = glossvideo.glossvideoperspective
+        return str(perspective_video.perspective)
+    return str(glossvideo.version)
```
Woseseltops, line 406: This needs a more descriptive name or a docstring.

signbank/dataset_operations.py
```diff
@@ -354,7 +360,144 @@ def update_gloss_video_backups(request, glossid):
...
+
+
+def get_primary_videos_for_gloss(gloss):
+    glossvideos = GlossVideo.objects.filter(gloss=gloss,
+                                            version=0,
+                                            glossvideonme=None,
+                                            glossvideoperspective=None).distinct().order_by('version')
+    display_glossvideos = ', '.join([str(gv.version) + ': ' + str(gv.videofile) for gv in glossvideos])
```
Woseseltops, line 414: I think this works too:

```python
display_glossvideos = ', '.join(f'{gv.version}: {gv.videofile}' for gv in glossvideos)
```

also, I see you do this a few times... perhaps a sperate function for it?

Woseseltops (commented):
I agree that it's well written like you said :). I only looked at the code, not the functionality yet... do you want functional feedback as well?

signbank/dataset_operations.py
```diff
@@ -354,7 +360,144 @@ def update_gloss_video_backups(request, glossid):
...
+        return not filename_is_correct
+    elif perspective is not None:
+        filename_is_correct = filename_matches_perspective(video_file_full_path) is not None
+        return not filename_is_correct
+    elif version > 0:
+        filename_is_correct = filename_matches_backup_video(video_file_full_path) is not None
+        return not filename_is_correct
+    else:
+        filename_is_correct = filename_matches_video(video_file_full_path) is not None
+        return not filename_is_correct
```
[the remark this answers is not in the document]
susanodd: I want the negative case. (Incorrect perhaps sounds better, or maybe "does not match file naming pattern")

signbank/dataset_operations.py
susanodd: That's a bit of a hack because of the new NME perspective which wasn't accounted for yet. I'll rename it.

signbank/dataset_operations.py
susanodd: I prefer to keep the explicit strings. The f-strings were causing problems with the filenames !!! (I know this because of the DEBUG setting that prints everything that is happening. The names ended up wrong with the f-strings. It was randomly removing the ".")

Issue #1409, opened 2024-12-03
Different perspectives also visible for video's with NME


uklomp:
Would it be possible to apply our new functionalities of the neutral gloss videos also to the videos uplaoded in the non-manual elements panel? That is: different perspectives, play, full screen etc.

susanodd:
Are you asking to be able to upload three videos for each of the NME ?
(They have numbers on them, but that is just the order of display. You want 3 for each? (As opposed to displaying the already uploaded ones as alternatives?)
(Or do you want them displayed on top of each other with the numbers on top?)

uklomp:
For every sign that will be uploaded for NME, for which we agreed it will be minimum 1 and maximum 3, we want to be able to show different perspectives - just like with the neutral gloss video. So camera left and right etc.

susanodd:
I'll let @Woseseltops do this. He did the API for the perspective video upload.

I'm not sure yet of how to properly name the new files. Also a new model class is needed. Maybe a subclass of both Perspective and NME videos.

Woseseltops:
### Functional design (mockup)

I think we need an extra dropdown during entry

> ![image](https://github.com/user-attachments/assets/f9e36022-e91f-492d-9b2a-67dd816902d1)

And if there are multiple videos with the same index, but different perspectives, it won't show up like this:

> ![image](https://github.com/user-attachments/assets/96ba0e29-ac4c-4440-8a85-290e2dd14696)

but like this

> ![image](https://github.com/user-attachments/assets/50fdf5ad-7432-4fcf-ab3a-1ac277bec26b)

Woseseltops:
### Technical design

1. Add a perspective field to ``GlossVideoNME``, reusing ``PERSPECTIVE_CHOICES``
2. Dropdown on the UI, processed and saved in the ``GlossVideoNME``
3. Change logic for the index counter: only increase if there already is a video **with the same perspective**
4. On display, detect which videos belong together, show the perspective interface.

Woseseltops:
No preference, as long as we're consistent. Your question makes me think you might want to implement this one. If so, make yourself assignee! If not, I'll make a start this afternoon

Woseseltops:
To do:
* Merge master, fix conflicts
* Use the perspective video player if applicable
* Investigate video file names

uklomp:
If the big bugs are fixed, it would be good if this issue is on the radar again.

uklomp:
Like we discussed just now, please note: 
[...]

PR #1537, opened 2025-04-03
fix: Correct the order of NME video file name parts #1409

vanlummelhuizen:
@susanodd signaled me about a bug in @Woseseltops 's implementation of the construction of NME video file names. 

I also searched for places where the new NME video file names are used and corrected it if I thought it was incorrect.

### DIFF
--- a/signbank/dataset_operations.py
+++ b/signbank/dataset_operations.py
@@ -316,6 +316,7 @@ def find_unlinked_video_files(dataset, linked_file_names):
             if (filename_without_extension.endswith('_small')
                     or filename_without_extension.endswith('_left')
                     or filename_without_extension.endswith('_right')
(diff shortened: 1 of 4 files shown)


susanodd:
So we will allow both "" as well as "_center" for the non-perspective NME files? Or should the files be changed when they become part of a perspective collection? (In other words, is "_center" optional?)

Woseseltops:
Apologies for messing this up; I missed/underestimated this because the NME functionality will still work with incorrect filenames. 

> So we will allow both "" as well as "_center" for the non-perspective NME files? 

Good question. Newly added videos will always have a perspective, but let's allow the option without too for backwards compability.

Anyway, this has my seal of approval, feel free to merge this if you think this solves the problem @susanodd !

Woseseltops (approved):

Issue #1526, opened 2025-03-21
For NME perspective videos, use the perspective video viewer


Woseseltops:
And if there are multiple NME videos with the same index, but different perspectives, it shouldn't show up like this:

![Image](https://github.com/user-attachments/assets/903843fe-5168-4775-b6b7-b5131e04f70c)

but like this

![Image](https://github.com/user-attachments/assets/a8dbfa73-2cfe-467d-9a57-26fa44e7464c)

susanodd:
This is a bit problematic to implement. All (left, right) perspective NME video objects also have the descriptions.

susanodd:
These things have been implemented:


-  Added perspective to NME video __str__ method
-  Added dictionary gloss method to retrieve NME lookup dictionary per NME offset to group the perspectives with the normal one. Nested dictionary
-  Boolean "has" methods for checking whether perspective exists for particular offset. These will be used in the template to conditionally generate code for the left/right icons and video display, etc.
-  Do not add description to perspective video objects of same NME video (also need to make sure form does not do this) NOTE: do we need these descriptions on the perspective "video"? I think not. (The perspective video object is an implementation, just a different camera) The perspective video viewer just puts all three videos on top of each other and switches which is visible. If we have text descriptions on each of these, they cannot be displayed.
-  Modifications to update NME method as described
-  Added the parsing and mechanics to the Detail View.
-  Implemented the show/hide/play/pause javascript. Makes use of element "data-offset" in the elements.

susanodd:
![Image](https://github.com/user-attachments/assets/00fe575c-13c8-4916-9377-aa4dfe2e52e2)

PR #1542, opened 2025-04-15
#1526: Display viewer for NME videos with perspective

susanodd:
I have implemented new model methods to accomplish what is needed. The javascript in the detail view has been parameterized with the offset of the NME video (collection) parameters.

Everything is implemented except a "full screen" button. (Jetske wrote the existing full screen code for normal videos.)

This issue is about the display of NME perspective videos.  As a side effect, I had to change the "NME update" because the descriptions and offset fields are no longer visible for the perspective objects in the template.
Another side effect is the "package". It needs to code explicitly include the perspective videos for the NME videos.

1. Added perspective to NME video `__str__` method
2. Added dictionary gloss method to retrieve NME lookup dictionary per NME offset to group the perspectives with the normal one. Nested dictionary
3. Boolean "has" methods for checking whether perspective exists for particular offset. These will be used in the template to conditionally generate code for the left/right icons and video display, etc.
4. Do not add description to perspective video objects of same NME video (also need to make sure form does not do this) NOTE: do we need these descriptions on the perspective "video"? I think not. (The perspective video object is an implementation, just a different camera) The perspective video viewer just puts all three videos on top of each other and switches which is visible. If we have text descriptions on each of these, they cannot be displayed.
5. Modifications to update NME method as described
6. Added the parsing and mechanics to the Detail View. 
7. Implemented the show/hide/play/pause javascript. Makes use of element "data-offset" in the elements.

PROBLEM: RETRIEVAL

SOLUTION: Implemented numerous helper functions and a dictionary data structure to assist the template (see new code in the files changed)

<details>
MASTER CODE:
The Gloss Edit View iterates over all the NME videos and provides editing of them if the user has permission.
This is why the perspective videos are showing up by themselves in the row.
The `get_nme_videos` method retrieves all of them for the gloss, regardless of whether or not they have a perspective.
The template currently iterates over all these gloss nme video objects.

The entire editing is a problem now because it is intertwined with the iteration over the collection of all the NME objects for the gloss.

The user is also able to edit the descriptions of all of the perspective videos as well as the normal one independently.
(It's not a collection per offset in the template.)
</details>

ANOTHER PROBLEM: UPDATE

SOLUTION: Did the following to not add descriptions to perspective objects, and made the delete also delete the perspectives, and the update of the offset, also move the perspective videos.

<details>
MASTER

When updating an NME video, if the offset is updated all of the objects with that offset also need to be updated. (The code does not do anything with this on master.)

Also in the dictionary/update.py => `update_nmevideo` method, there is a delete. Probably all nme's with the same offset need to be deleted, not just the one clicked on. (It's not sure what the user has clicked on on master.)
This is a problem in the display because the perspectives will be hidden unless played. The "offset" and the "perspectives" actually (ought to) function as data structure constraints. If the user would delete the primary NME video with a specific offset, then also the perspective objects should be deleted (since it's a different camera angle of the same video.)

### DIFF
--- a/signbank/dictionary/models.py
+++ b/signbank/dictionary/models.py
@@ -2470,6 +2470,17 @@ class Gloss(models.Model):
         nmevideos = GlossVideoNME.objects.filter(gloss=self)
         return nmevideos
 
+    def get_nme_videos_lookup(self):
+        # Group the NME video objects together by their offset
+        from signbank.video.models import GlossVideoNME
+        nmevideos = GlossVideoNME.objects.filter(gloss=self)
+        nme_videos_dict = dict()
+        for nmevideo in nmevideos:
+            if str(nmevideo.offset) not in nme_videos_dict.keys():
+                nme_videos_dict[str(nmevideo.offset)] = []
+            nme_videos_dict[str(nmevideo.offset)].append(nmevideo)
+        return nme_videos_dict
+
     def add_nme_video(self, user, videofile, new_offset, recorded, perspective='center'):
         # Preventing circular import
         from signbank.video.models import GlossVideoNME, GlossVideoHistory, get_video_file_path
--- a/signbank/dictionary/models.py
+++ b/signbank/dictionary/models.py
@@ -2466,26 +2466,40 @@ class Gloss(models.Model):
         return nmevideos.count()
 
     def get_nme_videos(self):
+        # Only retrieve the primary NME video
         from signbank.video.models import GlossVideoNME
-        nmevideos = GlossVideoNME.objects.filter(gloss=self)
+        nmevideos = GlossVideoNME.objects.filter(gloss=self, perspective__in=['', 'center'])
         return nmevideos
 
     def get_nme_videos_lookup(self):
-        # Group the NME video objects together by their offset
+        # Group the NME video objects together by their offset; this is used for retrieval in the template
         from signbank.video.models import GlossVideoNME
         nmevideos = GlossVideoNME.objects.filter(gloss=self)
         nme_videos_dict = dict()
         for nmevideo in nmevideos:
-            if str(nmevideo.offset) not in nme_videos_dict.keys():
-                nme_videos_dict[str(nmevideo.offset)] = []
-            nme_videos_dict[str(nmevideo.offset)].append(nmevideo)
+            if nmevideo.offset not in nme_videos_dict.keys():
+                nme_videos_dict[nmevideo.offset] = dict()
+            if nmevideo.perspective not in nme_videos_dict[nmevideo.offset].keys():
+                nme_videos_dict[nmevideo.offset][nmevideo.perspective] = nmevideo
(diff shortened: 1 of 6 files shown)


signbank/video/models.py
```diff
@@ -939,15 +940,32 @@ def __str__(self):
...
+        perspectivevideos = GlossVideoNME.objects.filter(gloss=self.gloss, offset=self.offset, perspective__in=['left', 'right'])
+        return perspectivevideos.count() > 0
+
+    def has_left_perspective(self):
+        nmevideos = GlossVideoNME.objects.filter(gloss=self.gloss, offset=self.offset, perspective='left')
+        return nmevideos.count() > 0
+
+    def has_right_perspective(self):
+        nmevideos = GlossVideoNME.objects.filter(gloss=self.gloss, offset=self.offset, perspective='right')
+        return nmevideos.count() > 0
```
Woseseltops, line 963: Can't you implement this is as one function with an argument?

signbank/dictionary/models.py
```diff
@@ -2466,15 +2466,40 @@ def has_nme_videos(self):
...
+        # this is used in the template to conditionally generate code
+        from signbank.video.models import GlossVideoNME
+        primary_nmevideo = GlossVideoNME.objects.filter(gloss=self, offset=offset, perspective__in=['', 'center']).first()
+        return primary_nmevideo.has_left_perspective if primary_nmevideo else False
+
+    def nme_video_has_right_perspective(self, offset):
+        # this is used in the template to conditionally generate code
+        from signbank.video.models import GlossVideoNME
+        primary_nmevideo = GlossVideoNME.objects.filter(gloss=self, offset=offset, perspective__in=['', 'center']).first()
+        return primary_nmevideo.has_right_perspective if primary_nmevideo else False
```
Woseseltops, line 2496: Can't this be one method with an argument?

Woseseltops (commented):
Without having seen the the actual functionality in action, these things caught my eye when looking at the code :)

signbank/video/models.py
susanodd: This is used in the template. If there were a parameter, then an additional "template filter function" is needed in order to apply it. 

I implemented it this way because the buttons above the video are hard-coded in a particular order, so this avoids extra code in the template.

(During implementation, I added such functions as needed to implemented the code in the template. The original primary video perspectives both have a loop each to find the correct perspective object.)

signbank/dictionary/models.py
susanodd: Again, same as above.

It's implemented this way because of the nested looping in the template. The offset is used everywhere in the element ids to make them unique. The perspective is also used, but it's inside nested code. This is used to direct the code generation.

susanodd:
nme-perspective-with-buttons

Woseseltops (approved):
Also reviewed functionally, looks great

PR #1545, opened 2025-04-22
Update gloss last updated timestamp on video upload

susanodd:


signbank/dictionary/gloss_revision.py
```diff
@@ -118,6 +120,18 @@ def pretty_print_revisions(gloss):
...
+                prefix, operation, offset, perspective = revision.field_name.split('_')
+            except ValueError:
+                prefix, operation, offset, perspective = 'gloss', 'nmevideo', '0', ''
+            if perspective == 'left':
+                video_perspective = " (" + gettext("Left") + ")"
+            elif perspective == 'right':
+                video_perspective = " (" + gettext("Right") + ")"
+            else:
+                video_perspective = ""
+            revision_verbose_fieldname = gettext("Upload NME Video") + ' ' + offset + video_perspective
```
Woseseltops, line 134:
```python
if perspective == 'left':
    video_perspective = f" ({gettext('Left')})"
elif perspective == 'right':
    video_perspective = f" ({gettext('Right')})"
else:
    video_perspective = ""

revision_verbose_fieldname = f"{gettext('Upload NME Video')} {offset}{video_perspective}"
```

signbank/dictionary/models.py
```diff
@@ -2449,13 +2448,18 @@ def add_video(self, user, videofile, recorded):
             except OSError:
                 msg = "The video could not be saved in the GlossVideo object for gloss " % self.pk
                 raise ValidationError(msg)
+
+            import datetime as DT
+            from django.utils.timezone import get_current_timezone
```
Woseseltops, line 2453: Perhaps do this right one go, and have imports at the top? I understand you probably did this because of the datetime import problem I described elsewhere, but let's not create more mess to be cleaned up later ;).

signbank/dictionary/models.py
```diff
@@ -2518,7 +2522,11 @@ def add_nme_video(self, user, videofile, new_offset, recorded, perspective='cent
             # Save the new videofile in the video object
             video.videofile.save(relative_path, videofile)
         else:
-            return GlossVideoNME(gloss=self, offset=offset, perspective=perspective) 
+            return GlossVideoNME(gloss=self, offset=offset, perspective=perspective)
+        import datetime as DT
+        from django.utils.timezone import get_current_timezone
```
Woseseltops, line 2527: Same thing about imports

signbank/dictionary/models.py
```diff
@@ -2559,6 +2567,10 @@ def add_perspective_video(self, user, videofile, new_perspective, recorded):
             msg = "No video file supplied for perspective video upload of gloss %s" \
                   % (self.pk)
             raise ValidationError(msg)
+        import datetime as DT
+        from django.utils.timezone import get_current_timezone
```
Woseseltops, line 2571: Same thing about imports

signbank/video/views.py
```diff
@@ -29,92 +32,111 @@ def addvideo(request):
...
-                    annotated_sentence.delete()
-                else:
-                    annotated_sentence.save()
+        if not form.is_valid():
+            # redirect back to the referring page or root
+            if 'HTTP_REFERER' in request.META:
+                url = request.META['HTTP_REFERER']
+            else:
+                url = '/'
+            return redirect(url)
```
Woseseltops, line 41: Thanks for this :)

signbank/dictionary/models.py
susanodd: I had the imports at the top to start, but was getting some errors so I moved them to be local. (I will try again. I'm not sure if it imports the same in the imported model's python file. That would be circular.)

signbank/dictionary/gloss_revision.py
susanodd: I have had trouble using those f-strings. I know they are not allowed when you include translations. (I thought about making a new issue. It won't translate them if it's inside an f-string. It needs to be a "format" string for it to translate / generate translations and then use them) (Probably to do with when it's evaluated)

PR #1548, opened 2025-05-06
Add creator and date to notes

susanodd:
This is completed.

Please see the comment in the issue for a question about the Notes "Note Type" pull-down and creation date.

I tried searching on notes to see what kind of data there already is for NGT.
See #1549 
It looks like when the user specifies a type for the Note that this probably is not something the user would change.
The note can just be deleted and a new note created. (I assume that is why we do not have the edit Type functionality anymore).

This can be tried out on `signbank-test`

Already existing notes have no creation date per review.

I added a lastUpdated field for the notes, which gets updated when the notes are updated.
This is not shown at the moment, but we'll have data in the database available if a search feature on note dates ends up on the wishlist. The notes show up in the revision history now.


signbank/dictionary/models.py
```diff
@@ -189,6 +189,9 @@ def __str__(self):
                                     verbose_name=_("Note Type"), related_name="definition")
     count = models.IntegerField(default=3)
     published = models.BooleanField(default=True)
+    creationDate = models.DateField(_('Creation date'), default=datetime(2015, 1, 1))
```
vanlummelhuizen, line 192: These fields can be set to NULL if `null=True`. Then you don't have choose a default.

signbank/dictionary/update.py
```diff
@@ -1564,34 +1564,46 @@ def update_definition(request, gloss, field, value):
...
-    
+
     if what == 'definitiondelete':
+        original_value = defn.note_text()
         defn.delete()
+        add_gloss_update_to_revision_history(request.user, gloss_or_morpheme, 'definitiondelete', original_value, '')
         return HttpResponseRedirect(reverse(reverse_url, kwargs={'pk': gloss_or_morpheme.id})+'?editdef')
-    
+
+    defn.lastUpdated = DT.datetime.now()
```
vanlummelhuizen, line 1574: The `lastUpdated` field already has `auto_now=True`. This would make this line not necessary.

signbank/dictionary/update.py
```diff
@@ -1755,7 +1767,12 @@ def add_definition(request, glossid):
     text = form.cleaned_data['text']
 
     defn = Definition(gloss=gloss_or_morpheme, count=count, role=role, text=text, published=published)
+    defn.creationDate = DT.datetime.now()
```
vanlummelhuizen, line 1770: The lastUpdated field already has auto_now=True. This would make this line not necessary.

signbank/dictionary/update.py
```diff
@@ -1755,7 +1767,12 @@ def add_definition(request, glossid):
...
+    defn.lastUpdated = DT.datetime.now()
```
vanlummelhuizen, line 1771: Ditto

signbank/dictionary/models.py
susanodd: Okay, I'll try that. It didn't work with blank=True

signbank/dictionary/update.py
susanodd: Okay, I'll do that. For other issues I always did that explicitly because I had to update it if related objects were updated, otherwise the gloss was not shown as having been updated (in the package). (You may have some quirks about other places in the code where it is done like this.)

signbank/dictionary/update.py
susanodd: Does the creation date ever get set if it's not set explicitly?
The users can create notes (definitions) and then update the notes.

Woseseltops (approved):
Good work. I was confused about this for a while until I realized you were also updating the gloss revision history, which is a nice touch.

signbank/dictionary/templates/dictionary/gloss_detail.html
```diff
@@ -2841,7 +2841,7 @@ <h2 id='modalTitleNote'>{% trans "Delete This Note" %}</h2>
                     {% endif %}
                     {% endif %}
                 </td>
-                <td>{{def.creationDate}}</td>
+                <td>{% if def.creationDate %}{{def.creationDate}}{% endif %}</td>
```
vanlummelhuizen, line 2844: This could also be done with the `default` template filter: https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#default

PR #1564, opened 2025-05-26
BUG FIX: Fixed weakdrop weakprop, made editable mimic list code

susanodd:
coerce Boolean internal values to strings using method to display, explicit use of internal Booleans in choice list for update

I'm not sure when the bug arose. It looks like it was using different choice lists. In the template, numbers were being used, and the (old) values were the actual NullBoolean internal values.
I added a coercion to make the internal values become strings (but using 'None' instead of it ending up the same as False for None values)
I modified the template to test on these strings rather than the internal "value" (NullBoolean).
Likewise, the update processing converts the internal (stored) values to numbers (corresponding to the choice list in the form)
The implementation of the editable code uses choice lists with numerical domains. Probably this went wrong since it does not work if the choice lists have actual Boolean values in the domain. The coercion was messed up.

### DIFF
--- a/media/js/gloss_edit_color.js
+++ b/media/js/gloss_edit_color.js
@@ -436,23 +436,55 @@ function configure_edit() {
          checkbox: { trueValue: yes_str, falseValue: no_str },
 		 callback : update_view_and_remember_original_value
      });
-     $('.edit_WD').editable(edit_post_url, {
-         params : { a: handedness_weakdrop,
-                    field: 'weakdrop',
-                    choices: handedness_weak_choices,
-                    colors: handedness_weak_choices_colors },
-         type      : 'select',
-         data: handedness_weak_choices,
-		 callback : update_view_and_remember_original_value
+     $('.edit_WD').click(function()
+	 {
+	     var this_data = $(this).attr('value');
+	     var edit_list_choices = handedness_weak_choices;
+	     var edit_list_choice_colors = handedness_weak_choices_colors;
+	     var index_of_modified_field = '1';
+         for (var key in edit_list_choices) {
+            var value = edit_list_choices[key];
+            if (value == this_data) {
+                index_of_modified_field = key;
+                break;
+            }
+         };
+         $(this).attr('data-value', index_of_modified_field);
+		 $(this).editable(edit_post_url, {
+		     params : { a: index_of_modified_field,
+		                field: 'weakdrop',
+		                display: $(this).attr('value'),
+		                colors: edit_list_choice_colors,
+		                choices: edit_list_choices },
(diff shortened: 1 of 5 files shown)


media/js/gloss_edit_color.js
```diff
@@ -436,23 +436,55 @@ function configure_edit() {
...
-                    choices: handedness_weak_choices,
-                    colors: handedness_weak_choices_colors },
-         type      : 'select',
-         data: handedness_weak_choices,
-		 callback : update_view_and_remember_original_value
+     $('.edit_WD').click(function()
+	 {
+	     var this_data = $(this).attr('value');
+	     var edit_list_choices = handedness_weak_choices;
+	     var edit_list_choice_colors = handedness_weak_choices_colors;
```
Woseseltops, line 443: I feel like this renaming of vars doesn't add much

media/js/gloss_edit_color.js
```diff
@@ -436,23 +436,55 @@ function configure_edit() {
...
-                    choices: handedness_weak_choices,
-                    colors: handedness_weak_choices_colors },
-         type      : 'select',
-         data: handedness_weak_choices,
-		 callback : update_view_and_remember_original_value
+     $('.edit_WP').click(function()
+	 {
+	     var this_data = $(this).attr('value');
+	     var edit_list_choices = handedness_weak_choices;
+	     var edit_list_choice_colors = handedness_weak_choices_colors;
```
Woseseltops, line 468: Same here, probably doesn't add much.

signbank/dictionary/models.py
```diff
@@ -1308,6 +1308,24 @@ def get_domhndsh_display(self):
...
+        return coerced_value
+
+    def get_weakprop_display(self):
+        if self.weakprop is None:
+            coerced_value = 'None'
+        elif self.weakprop is True:
+            coerced_value = 'True'
+        else:
+            coerced_value = 'False'
+        return coerced_value
```
Woseseltops, line 1327: This can be DRYer:

```py

def _bool_display(value):
    return str(value) if value in [True,None] else 'False'

def get_weakdrop_display(self):
    return self._bool_display(self.weakdrop)

def get_weakprop_display(self):
    return self._bool_display(self.weakprop) 
```

Woseseltops (commented):
Looks good. Now we're touching this code, a few minor requests.

signbank/dictionary/models.py
susanodd: The PyCharm parser doesn't like that `_bool_display` method at all. It wants a self argument. It also doesn't like using `self.weakdrop `as an argument in the call of its own method

Issue #1532, opened 2025-03-28
Too broad exception clauses


susanodd:

susanodd:
CODE ANALYSIS

This exception occurs when a translation language for a Lemma is empty/missing

This only applies to legacy lemmas that already exist (since ?). The code now requires all the languages to be filled in upon creation of a lemma.

The code of the `LemmaUpdateView` is dodgy.

I'm re-writing the code for the form to make the language fields be initialized using a set-up command rather than in the `_init__` method. The code as it is on master actually calls the `__init__` method twice on the form. I've made it be part of the view self.

susanodd:
Note, there are bugs in the original code. Some of the questionable code shown here (#1570) are actually causing the update to fail.

PR #1575, opened 2025-06-25
Rewrote update lemma view and related code

susanodd:
#1532 

Rewrote the update lemma view (`LemmaUpdateView`) and its related code because of bugs having to do with control flow.

The constraints have been normalised to all test the same things in the various places (form, view, model).

Some helper methods have been written to test the "obsolete" constraints and show these in the LemmaListView.

There's a new button in the Lemma List View to allow the user to copy annotations from the (unique for the lemma) gloss for missing lemma languages. (See screenshots https://github.com/Signbank/Global-signbank/issues/1579#issuecomment-3052320469) This accomplishes 2 (second paragraph) of https://github.com/Signbank/Global-signbank/issues/1579#issuecomment-3051672734

The wrench is conditionally visible button in Lemma List View and Lemma Update View.

A constraints column has been added to List View to show any (legacy) problems. These are shown in red and can be fixed (if available) using the wrench, or by going to Lemma Update View.
The constraint problems are also shown in Lemma Update View.
For any "duplicates" of the lemma, those are shown along with the glosses belonging to that other lemma.

PREVIOUS COMMENTS DURING CODE DEVELOPMENT

<details>

Another bug found. The constraint checks on the form do not catch the potential constraint failures in the save method. (#1579)

<details>
The Update Lemma page is referred to by Gloss Edit as well as by Lemma List View. It had both a get and a post method.

- The get method is called upon loading the page, or by getting there via the Lemma List.
- The post method was being called by Gloss Edit supposedly after "updating" the form, but there is no form on that page so it was called upon loading the page.

- A separate update method has been defined in dictionary/update.py (`update_lemma_idgloss`)
The previous "post" code of the view has been moved here and the urls reworked.

- The "referring page" is kept in variables as well as stored in a class variable and in the form of Lemma Update View.
Then when the update has been performed, control is returned there. (With caveats.)

- Returning to the two callers works when there is no "failure control flow"
Some additional context data has been added to return control flow to the correct page if possible. In implementing this, various options were experimented on. The failure control flow now goes to a "warning page with a link" (which already existed for other views). 
- The control on failure allows the page to return back to the form the user was working on. Some resetting of the stored url (the button in the Lemma Update template where it offers to navigate back to the caller) has been done to determine what this should be if it ended up being to a totally different lemma as result of the user navigating in the scrollbar.

</details>

 If the user is busy editing the lemma for a while and there is only one gloss for this lemma, it will return to gloss detail view. (This pertains to the hyperlink at the top of the Edit Lemma view. The depends on what the `HTTP_REFERER` has been doing.)

The `HTTP_REFERER` gradually ends up set to the Edit Lemma page instead of the original caller, if the user tries out some unsuccessful update attempts, reloads the page a bunch of times, or navigates to different lemmas in the scroll bar.

### DIFF
--- a/signbank/dictionary/views.py
+++ b/signbank/dictionary/views.py
@@ -20,7 +20,6 @@ from django.contrib import messages
 from django.core.exceptions import ValidationError, ObjectDoesNotExist, MultipleObjectsReturned, PermissionDenied
 from django.utils.translation import gettext_lazy as _, activate, override, gettext
 from django.views.decorators.csrf import csrf_exempt
-from django.utils.timezone import get_current_timezone
 from django.views.static import serve
 from django.db.transaction import atomic
 from django.db.models import Q
@@ -28,13 +27,12 @@ from django.core.files import File
 
 import guardian
 from guardian.shortcuts import get_user_perms, get_objects_for_user
-from tagging.models import TaggedItem
 from wsgiref.util import FileWrapper
 
 from signbank.settings.server_specific import (PREFIX_URL, WRITABLE_FOLDER, GLOSS_VIDEO_DIRECTORY, GLOSS_IMAGE_DIRECTORY,
                                                SHOW_DATASET_INTERFACE_OPTIONS, SIGNBANK_PACKAGES_FOLDER,
                                                LANGUAGE_CODE, USE_HANDSHAPE, HANDSHAPE_IMAGE_DIRECTORY,
-                                               RECENTLY_ADDED_SIGNS_PERIOD, USE_REGULAR_EXPRESSIONS,
+                                               USE_REGULAR_EXPRESSIONS,
                                                FIELDS, DEFINITION_FIELDS, SHOW_QUERY_PARAMETERS_AS_BUTTON,
                                                DEFAULT_LANGUAGE_HEADER_COLUMN, DEFAULT_DATASET_ACRONYM, OBLIGATORY_FIELDS)
 from signbank.settings.base import (MAXIMUM_UPLOAD_SIZE, SUPPORTED_CITATION_IMAGE_EXTENSIONS, USE_X_SENDFILE,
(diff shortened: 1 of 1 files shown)


susanodd:
The automatic merge messed up a lot of the formating in the update.py and actually erased all changes to both files.

I had to reconstruct it from the file on master plus the file on this branch.

Still, it reformatted a lot of the code (PyCharm pretty printed some stuff when it was copy pasted)

The code is all there except now it looks like some trivial reformatting has been done in this branch to other code that's not this issue.....

susanodd:
TO DO: Update the lemma lookahead to show the Lemma ID for the user to be able to resolve choice among duplicates.

signbank/tools.py
```diff
@@ -2347,3 +2349,73 @@ def get_checksum_for_path(file_path):
...
+        new_lemma_translation = LemmaIdglossTranslation(lemma=lemma, language=language,
+                                                        text=gloss_translation_for_language.text)
+        new_lemma_translation.save()
+    return
+
+
+def duplicate_lemmas(lemma):
+    # this finds other lemmas in the same dataset with duplicates of the translations of this lemma
+    # it returns a list of unique lemma ids.
+    if lemma.dataset is None:
```
Woseseltops, line 2409: More descriptive name: `find_duplicate_lemmas`

signbank/tools.py
```diff
@@ -2347,3 +2349,73 @@ def get_checksum_for_path(file_path):
             return file_hash.hexdigest()
     except FileNotFoundError:
         return None
+
+
+def dataset_no_multiple_lemma_translation_objects(dataset):
+    # for use in a command to check all the lemma's of a dataset for constraint violations
+    # the "none" case is legacy data, so not necessarily a constraint violation, but they cannot be empty if updated
+    results = dict()
+    results[dataset] = dict()
```
Woseseltops, line 2358:
```python
results = {dataset: {}}
```

signbank/tools.py
```diff
@@ -2347,3 +2349,73 @@ def get_checksum_for_path(file_path):
...
+    dataset_lemmas = dataset.lemmaidgloss_set.all()
+    for language in dataset.translation_languages.all():
+        results[dataset][language] = dict()
+        results[dataset][language]['none'] = []
+        results[dataset][language]['multiple'] = []
+        results[dataset][language]['empty'] = []
```
Woseseltops, line 2364:
```python
        violations[dataset][language] = {
            'none': [],
            'multiple': [],
            'empty': []
        }
```

signbank/dictionary/update.py
```diff
@@ -1261,6 +1282,102 @@ def update_annotation_idgloss(request, gloss, field, value):
...
+
+    copy_missing_lemmaidglosstranslation_from_annotationidglosstranslation(lemma)
+
+    lemma_translations = []
+    for translation in lemma.lemmaidglosstranslation_set.all():
+        lemma_translations.append({'lang2char': translation.language.language_code_2char,
+                                   'text': translation.text})
+    results = {}
+    results['lemmaid'] = lemmaid
+    results['lemma_translations'] = lemma_translations
```
Woseseltops, line 1377:
```python
results = {'lemmaid': lemmaid, 'lemma_translations': lemma_translations}
```

Woseseltops (commented):
Seems okay to me. TheP R was smaller than I initially thought, because there we a lot of changes related to newlines or to speces around +-signs ;)

signbank/tools.py
susanodd: Yeah, that structure wasn't planned as such and kept getting modified by what the data analysis was finding. Much better like that!

susanodd:
All the little formatting changes were done by PyCharm when I was merging it. The merge ended up deleting all the differences between this branch and master. I had to reconstruct the update file. Then PyCharm pretty printed it.

susanodd:
@Woseseltops I made all the requested improvements. I also improved the lemma lookahead and added the texts for "what to do".
It seems best to actually try using the code on a development server to see what it does. Also to make sure the helper text is actually helpful. (See the screenshots)

susanodd:
@Woseseltops this is on signbank-test now!!

It took me a while. There was a bug with the tests. (It seems related to that the "form" is set up as a dict in the code and not a real form.) All the tests work now.

(I modified the settings import as they did not match what I had done in the other files. They weren't being imported correctly in the venv in my setup. Now they work.)

PR #1578, opened 2025-06-30
Use mixin class for class meta methods

susanodd:
when desired

### DIFF
--- a/signbank/dictionary/models.py
+++ b/signbank/dictionary/models.py
@@ -188,7 +188,7 @@ class MetaModelMixin:
         return field
 
 
-class Definition(models.Model, MetaModelMixin):
+class Definition(MetaModelMixin, models.Model):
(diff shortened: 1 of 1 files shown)


signbank/dictionary/models.py
```diff
@@ -297,7 +300,7 @@ class Admin:
         search_fields = ['gloss__idgloss']
 
 
-class Handshape(models.Model):
+class Handshape(models.Model, MetaModelMixin):
```
vanlummelhuizen, line 303: I believe it is better to put the mixin class in front of the main class so that all methods of the main class are available (the last class overrides the methods of the classes in front).

vanlummelhuizen (changes requested):

susanodd:
Completed!

vanlummelhuizen:
Great @susanodd !

PR #1584, opened 2025-07-03
Use decorator require_http_methods instead of testing request.method

susanodd:


susanodd:
This overlaps with another issue, as noted by the conflict with update.py. This branch makes the same code improvements but is applied to more files.

susanodd:
This needs to be reviewed and merged into master because it is getting tedious to keep merging master by hand into this branch. Git has a problem auto-merging the `update.py` file.

Woseseltops (approved):

PR #1593, opened 2025-07-24
Extra function to check update dataset permissions

susanodd:
Similar code indentation of guards simplified

Some of #1582 is done here.

### DIFF
--- a/signbank/dictionary/update.py
+++ b/signbank/dictionary/update.py
@@ -1690,11 +1690,9 @@ def update_other_media(gloss,field,value):
     return HttpResponse(str(value), {'content-type': 'text/plain'})
 
 
+@require_http_methods(["POST"])
 def add_relation(request):
     """Add a new relation instance"""
-    
-    if not request.method == "POST":
-        return HttpResponseForbidden("Add relation method must be POST")
 
     form = RelationForm(request.POST)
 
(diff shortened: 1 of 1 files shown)


signbank/dictionary/update.py
```diff
@@ -3212,78 +3185,76 @@ def update_excluded_choices(request):
...
+    if new_color[0] == '#':
+        new_color = new_color[1:]
 
-            return HttpResponse(category + '\t' + fieldchoiceid + '\t' + str(original_value) + '\t' + str(new_color) + '\t' + machine_value,
-                                {'content-type': 'text/plain'})
+    original_value = thisfieldchoice.field_color
+    machine_value = str(thisfieldchoice.machine_value)
+    thisfieldchoice.field_color = new_color
+    thisfieldchoice.save()
+    # category = thisfieldchoice.field
```
Woseseltops, line 3217: If something is commented out but needs to stay in the code, explain why :)

signbank/dictionary/update.py
```diff
@@ -3212,78 +3185,76 @@ def update_excluded_choices(request):
...
 
-    # If we get here the request method has apparently been changed to get instead of post, can this happen?
-    raise Http404('Incorrect request')
+    return HttpResponse(category + '\t' + fieldchoiceid + '\t' + str(original_value) + '\t' + str(new_color) + '\t' + machine_value,
+                        {'content-type': 'text/plain'})
```
Woseseltops, line 3220: As you know, I like f-strings :)

Woseseltops (commented):
Only minor stuff :)

signbank/dictionary/update.py
susanodd: That will become part of the other issue that's to get rid of all that weirdness with the tabs.

signbank/dictionary/update.py
susanodd: Good point. I have no idea why that code is there. Oh, I just remembered. It's probably meant to be added to that string of "tabbed" results nobody likes. (But was not implemented. Those tabbed things are used by the javascript code to process the "results" and if it's say Phonology, show the user a message to reload the page because it was updated. (I'll add this to that other issue....)

Woseseltops (approved):

PR #1595, opened 2025-07-31
Condense update response tabbed results

susanodd:
Condensed dataset update method to combine fields of same type into single brances

in order to tab the correct types together in the response

### DIFF
--- a/signbank/dictionary/update.py
+++ b/signbank/dictionary/update.py
@@ -1104,9 +1104,6 @@ def update_gloss(request, glossid):
         try:
             fieldchoice = FieldChoice.objects.get(field=gloss_field.field_choice_category, machine_value=value)
         except (ObjectDoesNotExist, MultipleObjectsReturned):
-            print('Update field choice no unique machine value found: ', gloss_field.name,
-                  gloss_field.field_choice_category, value)
-            print('Setting to machine value 0')
             fieldchoice = FieldChoice.objects.get(field=gloss_field.field_choice_category, machine_value=0)
         gloss.__setattr__(field, fieldchoice)
         gloss.save()
@@ -1114,7 +1111,6 @@ def update_gloss(request, glossid):
 
     # Regular field updating
     else:
-        print('fall through update gloss: ', field, value, type(value))
         # Alert: Note that if field is idgloss, the following code updates it
         gloss.__setattr__(field, value)
         gloss.save()
@@ -1422,7 +1418,6 @@ def update_semanticfield(request, gloss, field, values):
     # expecting possibly multiple values
     # values is a list of strings
     new_semanticfields_to_save = []
-    print('update_semanticfield: ', gloss, values, field)
     # fetch all the valid semantic field choices
     # create a lookup dictionary mapping names to objects
     # the name is a unique field in the model
@@ -1453,7 +1448,6 @@ def update_semanticfield(request, gloss, field, values):
                              field_name='semField',
                              gloss=gloss, user=request.user, time=DT.datetime.now(tz=get_current_timezone()))
     revision.save()
-    print(str(new_semanticfield_value))
     return HttpResponse(str(new_semanticfield_value), {'content-type': 'text/plain'})
 
 
--- a/signbank/dictionary/views.py
+++ b/signbank/dictionary/views.py
@@ -1806,7 +1806,7 @@ def add_image(request):
         return redirect(url)
 
     # construct a filename for the image, use idgloss+gloss id
-    imagefile.name = gloss.idgloss + "-" + str(gloss.pk) + extension
+    imagefile.name = f'{gloss.idgloss}-{gloss.pk}{extension}'
 
     redirect_url = form.cleaned_data['redirect']
 
--- a/media/js/gloss_edit_color.js
+++ b/media/js/gloss_edit_color.js
@@ -692,7 +692,6 @@ function update_articulation(change_summary) {
 function update_view_and_remember_original_value(change_summary)
 {
 	split_values_count = change_summary.split('\t').length - 1;
-	console.log('update_view_and_remember_original_value: '+change_summary)
 	if (split_values_count > 0)
 	{
         split_values = change_summary.split('\t');
@@ -983,7 +982,6 @@ function update_affiliation_delete(data)
         return;
     };
     var affiliation_id = data.affiliation;
-    console.log('update affiliation after delete: '+affiliation_id)
     var gloss_affilication_tag = '#gloss_affiliation_' + affiliation_id;
     $(gloss_affilication_tag).remove();
 }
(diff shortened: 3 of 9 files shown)


susanodd:
If necessary for code improvement, a new spin-off follow-up issue can be made of the intentions of this issue.

For example, a json response instead of an http response and revised javascript code in the templates.

The tabbed response strings are taken apart in the javascript code by looking at how many tabs there are.
They are used to update the display with the specific gloss fields that have been updated.
Since this original code, we have done the same with more directed urls to update individual fields.

Woseseltops (commented):
Looks good @susanodd . I notice the part with `'\t'.join()` is still repeated often. Suggestion to make it even more DRY, up to you whether you like it:

```python
def generate_response(values):
    str_values = map(str, values)
    content = '\t'.join(str_values)
    return HttpResponse(content, content_type='text/plain')
```

So instead of

```python
    results = [str(original_value), str(newvalue), str(value), category_value, str(lemma_gloss_group), input_value]
    return HttpResponse('\t'.join(results), {'content-type': 'text/plain'})
```

you simply do

```python
return generate_response([original_value, newvalue, value, category_value, lemma_gloss_group, input_value])
```

susanodd:
Nice improvement!!

vanlummelhuizen:
> ```python
> return generate_response([original_value, newvalue, value, category_value, lemma_gloss_group, input_value])
> ```

Sorry for the nitpicking, but I think the name `generate_response` is rather generic and therefore not really informative.  Readability was better in the old code, IMHO.

I think if you replace the `results = [ ... ]` and `'\t'.join(results)` parts with an f-string, and leave the `HttpResponse` bit, the code shows exactly what is going on: a text string as a text/plain http response.

(If you want to get rid of `{'content-type': 'text/plain'}` you could extend HttpResponse to something like TextResponse where the content-type is always text/plain.)

@susanodd I am not saying this should be done. I accept the current solution except perhaps the name of the function. `generate_tabbed_text_response` would probably be more informative.

vanlummelhuizen:
> The idea behind this was because there are an arbitrary number of things tabbed together. The f-strings would still be using an arbitrary number of tab characters in them.

But it is *not* an arbitrary number of things. Those things are hardcoded and a rather limited number of variables that are used in these view-functions.

susanodd:
@Woseseltops @vanlummelhuizen what would be a more elegant solution would be to return a dictionary and rewrite the javascript function that takes apart the tabbed string.

https://github.com/Signbank/Global-signbank/blob/15282a30eb5ce3a77bea2beb6b1f6706bcaca09f/media/js/gloss_edit_color.js#L598-L693

Woseseltops:
Ha love these kind of discussions :) . I think we all agree the name `generate_response` I suggested is too vague, so let's make that more descriptive.

So we now have the f-string option:

```python
    return HttpResponse(f'{original_value}\t{newvalue}\t{category_value}\t{lemma_gloss_group}\t{input_value}', {'content-type': 'text/plain'})
```

Or the function option:

```python
return generate_tabbed_text_response([original_value, newvalue, value, category_value, lemma_gloss_group, input_value])
```

The f-string option also has most of the advantages the function option introduced; advantages like being able to switch the delimiter in one go will probably never materialize. In other words, I have no clear preference. You pick, @susanodd ;)

susanodd:
I'll think about it. 

[TO SELF]
Now I'm starting to want to make it be a dictionary but that would take much more coding. The things being returned probably ought to be more "typed named things". It just looks so arbitrary. The code is preparing things that are needed for it to display the update in the table in the template.

susanodd:
> If we only take the Python code into account, I would prefer the HttpResponse + f-string option. The reason is that the function abstracts this functionality (normally a good thing), but also obscures it. When using the HttpResponse + f-string option it is immediately clear what is going on, while the function only abstracts very little.
> 
> You could make the `\t` some kind of constant or setting because we use it in some views _and_ the Javascript. But...
> 
> If we also take the JS code into account, I would suggest returning a JSON with JsonResponse, which can easily be constructed from a list or dict with `json.dumps(...)`. It fits the unpacking of values as done in the JS much better than a string with `\t` as separator.

For convenience, I just renamed the function. It's used 9 places and it's way too much editing to make the f-strings again.

I agree with the JSON reponse. That's how most of the other templates work now.

I will have a go at that and see how it would impact the functions in Gloss Detail. (They're all using editable.) We only need to be able to direct what the response is supposed to change in the view. (All the table cells have identifiers, for example.)

susanodd:
I commented out some functions related to updating the sign language (many to many) field of a gloss. These seem obsolete, since it's not offered in the gloss edit interface anymore. (Only the dialect can be changed.)

I revised the "callback" function for dialect to not do the "tabs" stuff anymore. This also required changes in the Python code. It turns out the "sign language" code was actually just empty, since these are missing from gloss objects, since it's in the dataset.)

I'll continue to revise segments of the code for `update_gloss (dictionary/update.py)`. The control flow of this function is really sloppy and confusing since some branches go to other functions and others fall through. It's tedious to trace what the state of the variables is by the time control flow gets to the end.

susanodd:
The working version of this code is on signbank-susan

It's not entirely rewritten yet. But is being subsequently verified that all the small changes keep working correctly.

Additional bugs related to control flow and typing have been fixed as part of this issue as found.

susanodd:
The update_gloss method has been rewritten and simplified.

- the field types are now handled in separate control flow branches and have separate http response return statements
- some now have separate callbacks in the Gloss Edit view
- this makes the code ready to be rewritten as JsonResponse updates
- that will be done in a next phase

signbank/dictionary/views.py
```diff
@@ -1828,12 +1805,8 @@ def add_image(request):
...
         return redirect(url)
 
-    # construct a filename for the image, use sn
-    # if present, otherwise use idgloss+gloss id
-    if gloss.sn is not None:
-        imagefile.name = str(gloss.sn) + extension
-    else:
-        imagefile.name = gloss.idgloss + "-" + str(gloss.pk) + extension
+    # construct a filename for the image, use idgloss+gloss id
+    imagefile.name = gloss.idgloss + "-" + str(gloss.pk) + extension
```
Woseseltops, line 1809: Screams for f-strings :)

signbank/dictionary/update.py
```diff
@@ -1064,186 +983,156 @@ def update_gloss(request, glossid):
...
+    elif isinstance(Gloss.get_field(field), BooleanField):
+        # value is the html 'value' received during editing
+        if field in ['weakdrop', 'weakprop']:
+            # the weakdrop and weakprop fields make use of three-valued logic and None is a legitimate value aka Neutral
+            NEUTRALBOOLEANCHOICES = {None: '1', True: '2', False: '3'}
+            category_value = 'phonology'
+            if value not in ['1', '2', '3']:
+                # this code is for the case the user has not selected a value in the list
+                if value in [None, True, False]:
+                    value = NEUTRALBOOLEANCHOICES[value]
```
Woseseltops, line 1054: I see a lot of nested `for`s... would guard clauses work here?

Woseseltops (commented):
The changes to `update.py` are too large for me to fully grasp if the logic is still okay, but what I do see is a lot duplicate code being centered in a single helper function, which is good.

PR #1598, opened 2025-08-08
Moved imports to top

susanodd:
Fixed some type errors identified by PyCharm in dictionary:models.py. Used f-strings

### DIFF
--- a/signbank/csv_interface.py
+++ b/signbank/csv_interface.py
@@ -19,6 +19,7 @@ from signbank.dictionary.models import (Dataset, Gloss,
                                         Sense, GlossSense, SenseTranslation, ExampleSentenceTranslation)
 from signbank.dictionary.consistency_senses import check_consistency_senses
 from signbank.dictionary.update_senses_mapping import add_sense_to_revision_history
+from signbank.dictionary.adminviews import Echo
 
 from tagging.models import Tag, TaggedItem
 
@@ -1248,7 +1249,6 @@ def export_csv_template(request):
     # two rows are output to ensure a line break character is in the csv file
     csv_rows = [header_row, empty_row]
     # this is based on an example in the Django 4.2 documentation
-    from signbank.dictionary.adminviews import Echo
 
     pseudo_buffer = Echo()
     new_writer = csv.writer(pseudo_buffer)
--- a/signbank/dictionary/admin.py
+++ b/signbank/dictionary/admin.py
@@ -34,7 +34,10 @@ from signbank.dictionary.models import (Dataset, Gloss, Translation, LemmaIdglos
 from signbank.dictionary.forms import (FieldChoiceForm, SemanticFieldForm, HandshapeForm,
                                        QueryParameterFieldChoiceForm, SearchHistoryForm, QueryParameterBooleanForm,
                                        QueryParameterMultilingualForm, QueryParameterHandshapeForm)
-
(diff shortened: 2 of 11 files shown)


vanlummelhuizen (changes requested):
@susanodd This all looks good to me. You only should change the f-string path concatenation back to the `os.path.join` version.

susanodd:
We probably don't even need this command.....

susanodd:
> Perhaps it is a good idea to replace `magic` with [`mimetypes`](https://docs.python.org/3/library/mimetypes.html) because the former needs a OS specific library, see https://github.com/ahupp/python-magic?tab=readme-ov-file#debianubuntu while the latter is in the standard Python library.

Okay. I thought it was weird it needed a different version of magic in that file.
I removed that file and the other conversion command.

I checked the "create small videos" command and that works fine with the version of magic.
I checked the places where magic is used. I removed some other code that used magic during uploading of files.
The templates used to not restrict the type of the files to upload. (In the past we had even uploaded C++ files.)
Now it restricts this to images and videos (i.e., media)

In the admin, it's possible to search for non-mp4 files. But there's no command to convert. This was meant to be able to scout out errors if something is wrong. Same as the showing the file permissions.

The create small videos command works. Although probably this ought to be something the user can invoke, not us.
(No small videos are being made for a very long time.)

vanlummelhuizen:
Why did you remove those commands exactly? Perhaps we should discuss it first.

susanodd:
Using mimetypes is not working.

`ImportError: cannot import name 'guess_file_type' from 'mimetypes' (/usr/local/lib/python3.12/mimetypes.py)`

I followed the examples in the Python docs, but perhaps this does not work with Python 3.12 ?

I'm leaving it alone using magic.

There is a function that uses magic on eaf files, which Jetske wrote. I am not able to check whether replacing magic with mimtypes would work since I have no files for testing this code. (The annotated sentences files.)

susanodd:
> Why did you remove those commands exactly? Perhaps we should discuss it first.

They don't work at all. I tied the one for other media and it couldn't open the files.

I'll put them back. I'm pretty sure I wrote those commands. But that was when I had apache running on my computer.
I had to downgrade magic to get it to work.

vanlummelhuizen:
> Using mimetypes is not working.
> 
> `ImportError: cannot import name 'guess_file_type' from 'mimetypes' (/usr/local/lib/python3.12/mimetypes.py)`
> 
> I followed the examples in the Python docs, but perhaps this does not work with Python 3.12 ?

`guess_file_type` is not a function in `mimetypes`. `guess_type` is (https://docs.python.org/3.12/library/mimetypes.html#mimetypes.guess_type)

susanodd:
> > Using mimetypes is not working.
> > `ImportError: cannot import name 'guess_file_type' from 'mimetypes' (/usr/local/lib/python3.12/mimetypes.py)`
> > I followed the examples in the Python docs, but perhaps this does not work with Python 3.12 ?
> 
> `guess_file_type` is not a function in `mimetypes`. `guess_type` is (https://docs.python.org/3.12/library/mimetypes.html#mimetypes.guess_type)

Hmm.

I'm just worried about the eaf files, since I can't test that code once it's changed.

susanodd:
Completed! No more magic.

vanlummelhuizen:
@susanodd You can merge and close this pull request now.

PR #1602, opened 2025-08-15
Reverse synonym relations automatically added

susanodd:
The reverse relations are already added by the code. But the transitive relations are not.
I implemented the transitive synonym relations and revised the delete (synonym) relation to also delete the "target" of the relation from the other synonym relations of other glosses in the synonym group.

To assist the user:

Show all of the annotation translations and the gloss id in the pull-down for selecting the other gloss in the relation.

This is intended to make it easier for the user to confirm they are linking the gloss to the intended other gloss.

This is on signbank-test now.

SIDE ISSUE

- Implicit API endpoint ajax gloss lookahead and `str`(gloss)

<details>

The tests of Tobias for his API broke because I modified the gloss lookahead to include more information, via the `__str__` method.
Perhaps more than one version of the lookahead is needed?

I see what the problem is. I modified the `__str__` method of gloss to be similar to what the `__str__` method does for lemmas. That caused the tests to fail.
The method returns ``self.idgloss` `(!!!!!!) I switched it back and added a new method `to_string` to retrieve the desired information for display in the lookahead.

I don't know whether Tobias makes use of the lookahead function for glosses. It is in a url. So it's possible other Signbank modules may use this.

### DIFF
--- a/signbank/dictionary/adminviews.py
+++ b/signbank/dictionary/adminviews.py
@@ -6167,8 +6167,12 @@ class LemmaListView(ListView):
             if get_key.startswith(LemmaSearchForm.lemma_search_field_prefix) and get_value:
                 language_code_2char = get_key[len(LemmaSearchForm.lemma_search_field_prefix):]
                 language = Language.objects.get(language_code_2char=language_code_2char)
-                qs = qs.filter(lemmaidglosstranslation__text__icontains=get_value,
-                               lemmaidglosstranslation__language=language)
+                if USE_REGULAR_EXPRESSIONS:
+                    qs = qs.filter(lemmaidglosstranslation__text__iregex=get_value,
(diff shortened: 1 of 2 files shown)


signbank/tools.py
```diff
@@ -2429,3 +2429,61 @@ def find_duplicate_lemmas(lemma):
             duplicate_to_me.append(duplicate_translation.lemma.id)
     duplicate_lemma = list(set(duplicate_to_me))
     return duplicate_lemma
+
+
+def ensure_synonym_transitivity(gloss):
```
Woseseltops, line 2434: Isn't there a better place for this than the generic tools? Hmm, as this is related to data, it should be in `views.py`, but it's also not a model definition, so doesn't belong in `models.py`... perhaps there are more functions that work with relations between glosses, and they could fit together in something like `relation_tools`?

PR #1607, opened 2025-08-29
Added validation to Gloss Search template javascript

susanodd:
prior to doing search it replaces any bad input strings

### DIFF
--- a/signbank/dictionary/templates/dictionary/admin_gloss_list.html
+++ b/signbank/dictionary/templates/dictionary/admin_gloss_list.html
@@ -71,6 +71,40 @@ var morpheme_idgloss = '{{morpheme_idgloss|safe}}';
 var gloss_fields_to_populate = {{gloss_fields_to_populate|safe}};
 var gloss_fields_to_populate_keys = {{gloss_fields_to_populate_keys|safe}};
 
+function sanitizeInput(input) {
+  return input
+    .replace(/&/g, "&amp;")
+    .replace(/</g, "&lt;")
+    .replace(/>/g, "&gt;")
+    .replace(/"/g, "&quot;")
+    .replace(/'/g, "&#x27;");
+}
+
+function validateForm(formId) {
+{% for dataset_lang in dataset_languages %}
+    var targetAnnotationObjectId ='id_annotation_idgloss_{{dataset_lang.language_code_2char}}';
+    var input_annotation_value = document.forms[formId][targetAnnotationObjectId].value;
+    var sanitized = sanitizeInput(input_annotation_value);
+    var target_annotation_object = document.getElementById(targetAnnotationObjectId);
+    target_annotation_object.setAttribute("value", sanitized);
+    target_annotation_object.value = sanitized;
+
+    var targetLemmaObjectId = 'id_lemma_{{dataset_lang.language_code_2char}}';
+    var input_lemma_value = document.forms[formId][targetLemmaObjectId].value;
+    var sanitized = sanitizeInput(input_lemma_value);
+    var target_lemma_object = document.getElementById(targetLemmaObjectId);
+    target_lemma_object.setAttribute("value", sanitized);
+    target_lemma_object.value = sanitized;
+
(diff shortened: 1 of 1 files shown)


susanodd:
The most recent changes are not on the development server yet. I changed this to draft.

susanodd:
The branch is on signbank-test up to here.

susanodd:
The revised code is on signbank-test.

There's a noscript element that makes display:none for the main blocks of the template.

(This is not 100% at making things invisible. But as far as I can tell using the curl example, the resulting template "code" is made harmless.)

I tried to make it hide the javascript elements, but there seems to be nested javascript script elements. So perhaps that other code is imported makes this not work.)

susanodd:
The branch is on signbank-test up to here.

The tests all pass.

susanodd:
The branch is on signbank-test up to here.

susanodd:
The branch is on `signbank-test `up to here.

vanlummelhuizen:
@susanodd I see you are trying to catch potential dangerous URL parameters individually, e.g. in https://github.com/Signbank/Global-signbank/pull/1607/commits/b1c1c4f9a2039465cb056f394e1346753785f890 with the introduction of VIEW_TYPES and SEARCH_TYPES. Why do you do this instead of the broader solution I wrote about in the second part of [this comment](https://github.com/Signbank/Global-signbank/pull/1607#issuecomment-3278999734)?

susanodd:
> @susanodd I see you are trying to catch potential dangerous URL parameters individually, e.g. in [b1c1c4f](https://github.com/Signbank/Global-signbank/commit/b1c1c4f9a2039465cb056f394e1346753785f890) with the introduction of VIEW_TYPES and SEARCH_TYPES. Why do you do this instead of the broader solution I wrote about in the second part of [this comment](https://github.com/Signbank/Global-signbank/pull/1607#issuecomment-3278999734)?

Because if these have bad parameters the control flow will break.

susanodd:
Another code question about the pages mechanism. Can we ensure that the href links to pages are “relative” links in the template? (This probably doesn’t matter but I’m not a friendly hacker.) As it is so far with replacing the dodgy parameters the queries end up not finding any matches so there are no page links (the fixed templates). But I don’t know what happens if people start obtaining links and swapping stuff in them.

susanodd:
> > Because if these have bad parameters the control flow will break. They’re basically constants.
> 
> Alright, can you tell me where the flow would break? All I can find are `if`-statements that compare (`self.`)`search_type` to specific strings (sign, sense, etc) and that have `else` block. I think it would be totally OK if `search_type` is some garbage/dangerous strings in this situation.
> 
> By the way, catching garbage/dangerous parameters to prevent breaking the control flow is outside the XSS problem solution.

The scrollbar makes use of this as well as the session variables and the Query View. The type of stuff in the scrollbar is shown differently if possible.

I’m concerned that a mischievous user could cause Signbank to crash. (The things that are supposed to be numbers cause this.)

vanlummelhuizen:
> The scrollbar makes use of this as well as the session variables and the Query View. The type of stuff in the scrollbar is shown differently if possible.

Please, be more specific. Where in the code? I would like to see a filename and line numbers, or better yet, a direct link (see https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-a-permanent-link-to-a-code-snippet).

susanodd:
The code is on signbank-test up to here.

susanodd:
signbank-test is synced up to here.

All of the tests pass.

susanodd:
https://github.com/Signbank/Global-signbank/blob/667d4c1ddc439e5433cf05c840741ea7f2ff20a6/signbank/dictionary/templates/dictionary/gloss_detail.html#L97

https://github.com/Signbank/Global-signbank/blob/667d4c1ddc439e5433cf05c840741ea7f2ff20a6/signbank/dictionary/templates/dictionary/search_result_bar.html#L14

I don't really know if anything could happen. But the variable appears in the code.

susanodd:
Aaargh!!!! I didn't do the Handshapes List View template

PR #1608, opened 2025-09-01
Ajax gloss complete added datasetid to url

susanodd:
Revised code to accomodate datasets with one language and the interface language not being in the dataset languages


Woseseltops (approved):
Only thing that came to my mind: if I'm not mistaken it's now obligatory to add the datasetid to the url, but I also kind of liked that it tried to figure out itself if not. If you feel it makes sense, you could consider making the datasetid optional, but totally up to you. 

Whatever you choose, no need for me to review the pull request again, merge if you're happy with your result :)

PR #1610, opened 2025-09-02
BUG FIX: Revision History identity entries

susanodd:
This bug happens unintendedly when the user is using auto-advance to edit phonology in Gloss Edit view.
A choice list field auto-advances to the repeat checkbox field. If the user clicks ok (auto-advance proceeds) or cancel (auto-advance stops). Pressing OK has the effect of sending an update even if the checkbox hasn't been changed.

The code now takes this into account and adds extra identity checks before saving update data to the revision history.

Omit identity boolean field updates from revision history. 
Legacy code checks in Clean Up in case of combination of user friendly and booleans in revision history.
Check for identity of combined format Boolean (string) and user friendly.

Don't add identity "updates" in general update gloss code. (The fall through case.)

This is a bug fix. The update gloss code is being revised in a different issue.where the focus is on simplifying the control flow for all the different types of fields. This bug was due to sloppy, obscured control flow.

This issue is for the revision history storage.

### DIFF
--- a/signbank/dictionary/gloss_revision.py
+++ b/signbank/dictionary/gloss_revision.py
@@ -1,4 +1,5 @@
 from django.db.models import Q, BooleanField
+from django.db.models.fields import TextField, CharField
 from django.utils.translation import gettext, gettext_lazy as _
 
 from django.contrib.auth.decorators import permission_required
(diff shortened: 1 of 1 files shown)


susanodd:
Baah! Did CoPilot just do Chinese translations?

Woseseltops (commented):
Like Copilot, I also feel that these lists of strings could be clenaer, but other than that it looks good. Feel free to merge directly after processing (or not) the comments.

PR #1616, opened 2025-09-25
987 xss attempt 2

vanlummelhuizen:
The [`json_script` template filter](https://docs.djangoproject.com/en/5.2/ref/templates/builtins/#std-templatefilter-json_script) solves the XSS vulnerability in a concise and elegant way. It puts potential dangerous strings on the page in a safe way. With this filter it is not necessary to encode/escape the string(s) in another way, whether on the front-end with Javascript or on the back-end somewhere in Python. The result is that the string entered by the user is kept in the input box unchanged, while still preventing XSS.


susanodd:
Did you check that the translation blocks still work correctly? I had trouble with nesting when I wanted to move template code to auxiliary files to decrease the file size of gloss_detail.html. It broke the translations.

Woseseltops (approved):

PR #1630, opened 2025-10-01
Pagination parameters

susanodd:


signbank/tools.py
```diff
@@ -2363,6 +2364,34 @@ def get_checksum_for_path(file_path):
         return None
 
 
+def get_page_parameters_for_listview(search_form, request_get_parameters, query_parameters):
+    # fill page parameters
+    page_params_list = []
+    page_get_parameters = ""
```
vanlummelhuizen: Can be removed and handled when `return`ing.

PR #1634, opened 2025-10-02
Guard against url parameters not in model

susanodd:
Check that extra parameters not in the model have been added to the url, if so, ignore them

- form fields signLanguage and dialects differ from the Gloss model fields for them, these have been explicitly added to the check whether the url field appears in the form
- some helper functions for determining whether a given url parameter is okay or not okay; this covers totally bogus field names and multiple select field names without the brackets. The brackets are used to ascertain whether a "getlist" is needed on the request.GET parameters

Original formulation below (ignore)
<details>

This form code is interesting:

https://github.com/Signbank/Global-signbank/blob/ec51da5d45cb02ecf16bbfde0e43d58cfea75137/signbank/dictionary/forms.py#L361-L368

The domains of the choice lists are different types.

That these are different types in the (dynamic) form fields, ends up that in the url, they are still all strings.

I only just noticed this when making a more dynamic check on the parameters:

`dict(searchform.fields[get_key].widget.choices).keys()`

I suspect the code evolved this way because the weird choice lists with "2" being True were in the code originally.
(We could choose to normalise these in a different issue.)

I'm checking where the code falls through while processing the parameters.
Then checking what the key and value is. It needs to test whether the value has the correct type for the key.
If you put repeat=foo in the query, it falls through. This ends up ignored, but it is visible as an "empty" value in Query View.
Instead it should not be put in the parameters at all. (Since we decided to throw away bad values)

### DIFF
--- a/signbank/dictionary/adminviews.py
+++ b/signbank/dictionary/adminviews.py
@@ -505,6 +505,14 @@ class AnnotatedSentenceListView(ListView):
         return qs
 
     def get_context_data(self, **kwargs):
+        if 'page' in self.request.GET:
+            safe_pages = filter_page_values(self.request.GET.getlist('page', []))
+            self.request.GET._mutable = True
+            if safe_pages:
+                self.request.GET['page'] = safe_pages[0]
+            else:
+                del self.request.GET['page']
+            self.request.GET._mutable = False
         context = super().get_context_data(**kwargs)
 
         selected_datasets = get_selected_datasets(self.request)
@@ -958,7 +966,14 @@ class SenseListView(ListView):
         set_up_fieldchoice_translations(self.sentence_search_form, sentence_fields_with_choices)
 
     def get_context_data(self, **kwargs):
-        # Call the base implementation first to get a context
+        if 'page' in self.request.GET:
+            safe_pages = filter_page_values(self.request.GET.getlist('page', []))
+            self.request.GET._mutable = True
+            if safe_pages:
+                self.request.GET['page'] = safe_pages[0]
+            else:
+                del self.request.GET['page']
+            self.request.GET._mutable = False
         context = super(SenseListView, self).get_context_data(**kwargs)
 
         context = get_context_data_for_list_view(self.request, self, self.kwargs, context)
@@ -2040,7 +2055,14 @@ class MorphemeListView(ListView):
         set_up_fieldchoice_translations(self.search_form, fields_with_choices)
 
     def get_context_data(self, **kwargs):
-        # Call the base implementation first to get a context
+        if 'page' in self.request.GET:
+            safe_pages = filter_page_values(self.request.GET.getlist('page', []))
+            self.request.GET._mutable = True
+            if safe_pages:
+                self.request.GET['page'] = safe_pages[0]
+            else:
+                del self.request.GET['page']
+            self.request.GET._mutable = False
         context = super(MorphemeListView, self).get_context_data(**kwargs)
 
         selected_datasets = get_selected_datasets(self.request)
@@ -2626,6 +2648,14 @@ class MinimalPairsListView(ListView):
         set_up_fieldchoice_translations(self.search_form, fields_with_choices)
 
     def get_context_data(self, **kwargs):
+        if 'page' in self.request.GET:
+            safe_pages = filter_page_values(self.request.GET.getlist('page', []))
+            self.request.GET._mutable = True
+            if safe_pages:
+                self.request.GET['page'] = safe_pages[0]
+            else:
+                del self.request.GET['page']
+            self.request.GET._mutable = False
         context = super(MinimalPairsListView, self).get_context_data(**kwargs)
 
         selected_datasets = get_selected_datasets(self.request)
@@ -6257,6 +6287,14 @@ class LemmaListView(ListView):
         return results
 
     def get_context_data(self, **kwargs):
+        if 'page' in self.request.GET:
+            safe_pages = filter_page_values(self.request.GET.getlist('page', []))
+            self.request.GET._mutable = True
+            if safe_pages:
+                self.request.GET['page'] = safe_pages[0]
+            else:
+                del self.request.GET['page']
+            self.request.GET._mutable = False
         context = super(LemmaListView, self).get_context_data(**kwargs)
 
         context['SHOW_DATASET_INTERFACE_OPTIONS'] = SHOW_DATASET_INTERFACE_OPTIONS
@@ -6700,6 +6738,14 @@ class KeywordListView(ListView):
         return super(KeywordListView, self).get(request, *args, **kwargs)
 
     def get_context_data(self, **kwargs):
+        if 'page' in self.request.GET:
+            safe_pages = filter_page_values(self.request.GET.getlist('page', []))
+            self.request.GET._mutable = True
+            if safe_pages:
+                self.request.GET['page'] = safe_pages[0]
+            else:
+                del self.request.GET['page']
+            self.request.GET._mutable = False
         context = super(KeywordListView, self).get_context_data(**kwargs)
 
         selected_datasets = get_selected_datasets(self.request)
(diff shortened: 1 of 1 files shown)


signbank/query_parameters.py
```diff
@@ -895,6 +920,7 @@ def queryset_from_get(formclass, searchform, GET, qs):
                                   'translation__language': language})
             else:
                 # normal text field
+                print('queryset from get, text field fall-through: ', get_key, get_value)
```
vanlummelhuizen: Remove print statement

signbank/tools.py
```diff
@@ -2364,9 +2365,52 @@ def get_checksum_for_path(file_path):
...
+    search_form_fields = list(search_form.fields.keys()) + ['signlanguage', 'dialect']
+    multiselect_fields = get_multiselect_fieldnames() + ['signlanguage', 'dialect']
+
+    multiselect_fields_of_form = [f'{field}[]' for field in multiselect_fields if field in search_form_fields]
+    multiselect_fields_missing_brackets = [field for field in multiselect_fields if field in search_form_fields]
+    okay_parameters = []
+    not_okay_parameters = []
+    workflow_parameters = []
+    for key, value in request_get_parameters.items():
+        if key == 'page' or key in query_parameters.keys():
```
vanlummelhuizen: `key == 'page'` is also handled in line 2400. Remove one.

signbank/tools.py
```diff
@@ -2364,9 +2365,52 @@ def get_checksum_for_path(file_path):
...
+def check_get_parameters_are_available(search_form, request_get_parameters, query_parameters):
+    search_form_fields = list(search_form.fields.keys()) + ['signlanguage', 'dialect']
+    multiselect_fields = get_multiselect_fieldnames() + ['signlanguage', 'dialect']
+
+    multiselect_fields_of_form = [f'{field}[]' for field in multiselect_fields if field in search_form_fields]
+    multiselect_fields_missing_brackets = [field for field in multiselect_fields if field in search_form_fields]
+    okay_parameters = []
+    not_okay_parameters = []
+    workflow_parameters = []
+    for key, value in request_get_parameters.items():
```
vanlummelhuizen: `value` is never used. Do `.keys()` instead of `.items()`.

signbank/tools.py
```diff
@@ -2364,9 +2365,52 @@ def get_checksum_for_path(file_path):
...
+            not_okay_parameters.append(key)
+            continue
+        if key in ['csrfmiddlewaretoken', 'search_type', 'view_type', 'format', 'show_all', 'page']:
+            # key is not in  form and not an operational parameter
+            workflow_parameters.append(key)
+            continue
+        if key not in search_form_fields:
+            not_okay_parameters.append(key)
+            continue
+        okay_parameters.append(key)
```
vanlummelhuizen: I would reverse the logic here. So start with `if key in ...`  and end with `not_okay_parameters.append(key)`. Then, the last line gathers all that remains, which are things we don't know and are therefore not okay.

signbank/tools.py
```diff
@@ -2364,9 +2365,52 @@ def get_checksum_for_path(file_path):
...
+    return okay_parameters, not_okay_parameters, workflow_parameters
+
+
+def get_page_parameters_for_listview(search_form, request_get_parameters, query_parameters):
     page_params_list = []
+    okay_parameters, not_okay_parameters, workflow_parameters = check_get_parameters_are_available(search_form, request_get_parameters, query_parameters)
```
vanlummelhuizen: Only `not_okay_parameters` are used. So we do not need to check whether a key is either okay of workflow. Leaving those checks out would make check_get_parameters_are_available shorter (and it would need another name).

signbank/tools.py
```diff
@@ -2380,6 +2424,8 @@ def get_page_parameters_for_listview(request_get_parameters, query_parameters):
         # add other url parameters and ignore if processed above, this allows processing of multi-select parameters
         if key == 'page' or key in query_parameters.keys():
             continue
+        if key in not_okay_parameters:
```
vanlummelhuizen: See my comment above.
Another way to go is to list all okay parameters and when key is not in it, continue.

vanlummelhuizen (changes requested):
@susanodd Apart from other changes, I think we need to rethink the way `check_get_parameters_are_available` works and how it is used. We should be able to simplify it dramatically by putting only expected URL parameters in a single list and then testing against that.

signbank/tools.py
```diff
@@ -2364,9 +2365,52 @@ def get_checksum_for_path(file_path):
...
+    multiselect_fields_of_form = [f'{field}[]' for field in multiselect_fields if field in search_form_fields]
+    multiselect_fields_missing_brackets = [field for field in multiselect_fields if field in search_form_fields]
+    okay_parameters = []
+    not_okay_parameters = []
+    workflow_parameters = []
+    for key, value in request_get_parameters.items():
+        if key == 'page' or key in query_parameters.keys():
+            okay_parameters.append(key)
+            continue
+        if key.endswith('[]') and key[:-2] not in multiselect_fields_missing_brackets:
```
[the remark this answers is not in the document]
susanodd: There was originally a problem with things that fell through, to see how to catch them. I modified the order of evaluation in many places. (In the past the "update gloss" code was such a mess with all the different branches.)
I will attempt to make these more purposeful.

signbank/tools.py
susanodd: I forgot to document this. The query parameters are only the non-empty ones, that is why this function is written so. It's checking all the "extra" stuff that may be in the url. Here that 'page' is there is unique to the page references in the template. The 'page' parameter is written in the template, so it can't be put into the result. (I wasn't sure of the evaluation order nor when the function would be called.)

signbank/dictionary/adminviews.py
```diff
@@ -560,6 +560,12 @@ def __init__(self, *args, **kwargs):
 
     def get_context_data(self, **kwargs):
         # Call the base implementation first to get a context
+        if 'page' in self.request.GET:
+            tempdict = self.request.GET.copy()
+            self.request.GET._mutable = True
+            safe_pages = filter_page_values(self.request.GET.getlist('page', []))
+            tempdict['page'] = safe_pages[0] if safe_pages else ['1']
+            self.request.GET = tempdict
```
vanlummelhuizen: When setting `request.GET` to mutable, you could change the `['page']` element directly, and get rid of the `tempdict` variable. Perhaps it is wise to make `request.GET` immutable again at the end. Moreover, we could consider doing this in a function called something like `force_change_querydict`. 

`filter_page_values` is such a small function, and it is used only once. I would just do the list comprehension right here.

susanodd: > When setting `request.GET` to mutable, you could change the `['page']` element directly, and get rid of the `tempdict` variable. Perhaps it is wise to make `request.GET` immutable again at the end. Moreover, we could consider doing this in a function called something like `force_change_querydict`.
> 
> `filter_page_values` is such a small function, and it is used only once. I would just do the list comprehension right here.

I recoded this piece of code as suggested, except I left the helper function.
I tested it out on examples. If a user mocks up the page parameters, it now removes the page parameter if no good ones are left. (This is still visible in the initial url, but it's not in the page references, those work correctly.)

This block of code needs to be added to 9 other templates.
I'm not exactly sure if it's okay to be passing around the request.GET and modifying it thusly in a helper function.
(That seems like the kind of stuff they tell you not to do when you learn programming. Thus side effects inside a different function.)
Any preference for this? Or is it okay as it is now, and just copy it to the other templates?

susanodd:
> @susanodd When doing tests, I get
> 
> ```
> AttributeError: type object 'LemmaIdgloss' has no attribute 'get_field_names'
> ```
> 
> LemmaIdgloss probably needs MetaModelMixin. Can you check?

I fixed some of these on master a few weeks ago after the tests found them. Maybe merging other branches ended up undoing it if I forgot to merge master to it. It was after the names were changed to actual model names instead of strings or else it’s just the new code here. Except this seems like Déjà vu.

susanodd:
This one is ready now.

PR #1643, opened 2025-10-16
Rework sort order buttons

susanodd:
Because the font elements are no longer available and because PyCharm did not approve of the way the search form sort order was used in the template, moreover the x button was not shown because of this (at least Gloss List View)

Fixes:
- The code has been rewritten to use buttons instead of links with hrefs
- The "x" button is shown conditionally, only when that column has been sorted either up or down
- The visibility colour of the active sort order is obtained from the query parameters and set on page load
- MorphemeListView, HandshapeListView, GlossListView, AnnotatedSentenceListView
- Obsolete sort order unused code removed from templates without sort buttons (Exception to this is QueryListView which shares ajax calls with GlossListView and needs the function header for type-checking purposes)

DEPENDENT ON #1630
Please approve that first since the sorting spans multiple pages and that page references are broken on master.

### DIFF
--- a/signbank/dictionary/templates/dictionary/admin_morpheme_list.html
+++ b/signbank/dictionary/templates/dictionary/admin_morpheme_list.html
@@ -716,46 +716,50 @@ function do_set_lemma_language(el) {
   </tbody>
 </table>
 
+{% if page_obj.has_other_pages %}
 <div class="pagination">
-    <span class="step-links">
-
+    <div class="step-links">
         <ul class='pagination pagination-sm'>
         {% if page_obj.has_previous %}
-            <li><a href="?page={{page_obj.previous_page_number}}{% for key,value in request.GET.items %}{% if key != 'page' %}&{{ key }}={{ value }}{% endif %}{% endfor %}">&laquo;</a></li>
+            {% if page_obj.number > 1 %}
+            <li><a href="?page=1{{page_get_parameters}}">&laquo;&laquo;</a></li>
+            {% endif %}
+            <li><a href="?page={{page_obj.previous_page_number}}{{page_get_parameters}}">&laquo;</a></li>
         {% endif %}
-          
+
           {% if  page_obj.number > 10 %}
             <li><a>...</a></li>
           {% endif %}
 
           {% for p in page_obj.paginator.page_range %}
-          
+
              {% if p < page_obj.number|add:"10" and  p > page_obj.number|add:"-10" %}
-             <li {% if p == page_obj.number %}class='active'{% endif %}>
-             <a href='?page={{p}}{% for key,value in request.GET.items %}{% if key != 'page' %}&{{ key }}={{ value }}{% endif %}{% endfor %}'>{% if p == 0 %}Start{% else %}{{p}}{% endif %}</a>
-             </li>
+                {% if p == page_obj.number %}
+                     <li class='active'>
+                 {% else %}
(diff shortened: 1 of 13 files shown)


signbank/dictionary/templates/dictionary/admin_morpheme_list.html
```diff
@@ -183,6 +195,29 @@
...
+            stop_button_id = stop_button_id + query_initial + "_del";
+         } else {
+            button_id = button_id + query_initial + "_asc";
+            stop_button_id = stop_button_id + query_initial + "_del";
+         }
+         $(button_id).addClass("active");
+         $(button_id).trigger('change');
+         $(stop_button_id).show();
+         $(stop_button_id).trigger('change');
+     };
```
Woseseltops: I notice very similar code here 3 different times. If possible centralize it in 1 one place?

Also, I believe it can be shorter; perhaps you like this, not 100% sure if it will work:

```javascript
if (populate_fields_keys.includes('sortOrder')) {
    let query = populate_fields.sortOrder;
    const base = query.startsWith('-') ? query.slice(1) : query;
    const dir = query.startsWith('-') ? 'desc' : 'asc';
    const button = `#sort_button_${base}_${dir}`;
    const stopButton = `#sort_button_${base}_del`;

    $(button).addClass("active").trigger('change');
    $(stopButton).show().trigger('change');
}
```

susanodd:
About the "reset" comments above for Annotated Sentence List View,
there are a lot of templates that use "reset". Except it's unclear if these are being tested appropriately.
They are using "strings" in the field, but some of the code is testing like this:

https://github.com/Signbank/Global-signbank/blob/e65cdd08c4c38daaee9256a948b341e29d1df233/signbank/dictionary/adminviews.py#L6166

So if it's '0' that ends up being true since it's a string?

susanodd:
Hmmm. I've been checking some of the other code for the "reset" code in the form. A couple of the others also were using "reset" as the value, so I changed them also to "1" as suggested by CoPilot.

Code Guru @vanlummelhuizen 

I also checked the various buttons where the name "reset" appears.
Normally this is hidden in the form. But the `admin_handshape_list.html` is older code. It has a translation as the value.
(There is some inconsistency between older code and newer code. Normally we just leave stuff, if it's not broken, don't fix it.)
But sometimes I'm wondering if it is broken but we just haven't noticed. It could also be that the template language has gradually changed.

https://github.com/Signbank/Global-signbank/blob/e65cdd08c4c38daaee9256a948b341e29d1df233/signbank/dictionary/templates/dictionary/admin_handshape_list.html#L333-L334

versus

https://github.com/Signbank/Global-signbank/blob/e65cdd08c4c38daaee9256a948b341e29d1df233/signbank/dictionary/templates/dictionary/admin_lemma_list.html#L195-L197

Here the "name" is associated with a hidden input field.

Which is better?

susanodd:
```
grep "includes.*sortOrder" -R signbank/
signbank/dictionary/templates/dictionary/admin_gloss_list.html:                     if (populate_other_parameters.includes('sortOrder')) {
signbank/dictionary/templates/dictionary/admin_annotatedsentence_list.html:     if (query_parameters_keys.includes('sortOrder')) {
signbank/dictionary/templates/dictionary/admin_morpheme_list.html:     if (populate_fields_keys.includes('sortOrder')) {
```

PR #1650, opened 2025-10-27
Translations and template generating translations

susanodd:
- Added translation text for release information.
- Disambiguated fuzzy translations
- Repaired block translations that were too long to translate, and block translations that did not parse according to PyCharm


susanodd:
Getting rid of the "list" html nodes applies to multiple templates.
These are used in lists that are generated dynamically where the length of the list and the first row play a part in the display.
This needs to be in a separate issue because recoding it causes lots of other complaints by PyCharm about missing closing elements and forms not being allowed where they are.
That is too much to change in this "translations" issue.

PR #1655, opened 2025-11-05
Add convert video to MP4 button to admin

susanodd:
- For convenience a filter on video format / file type Webm has been added.

- The convert button (choice in the pull-down in GlossVideo admin) is implemented so that if the file extension / filename does not match the type of video file, it is renamed first so it has a correct extension. Then the conversion is applied to the (renamed) video file.
This is necessary for the case that the non-mp4 video has mp4 as an extension.
- The original (renamed with correct extension) file is moved to the backups folder after conversion. This is necessary if the user converts "bakNNN" files that the original ones do not pollute the folder. The original files that have been converted are not referenced anymore by the objects.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 setuptools==78.1.1
 attrs==17.4.0
 beautifulsoup4==4.12.3
-Django==4.2.25
+Django==4.2.26
 django-appconf==1.0.5
 django-bootstrap3==22.2
 django-colorfield==0.8.0


signbank/video/admin.py
```diff
@@ -429,15 +463,126 @@ def unlink_files(modeladmin, request, queryset):
...
+            desired_relative_path = os.path.join(GLOSS_VIDEO_DIRECTORY,
+                                                 dataset_dir, two_letter_dir, desired_filename)
+            if DEBUG_VIDEOS:
+                print('video:admin:convert_non_mp4_videos:convert_video: ', source, destination)
+            if not os.path.exists(video_file_full_path):
+                return
+            okay = convert_video(source, destination)
+            if DEBUG_VIDEOS:
+                print('video:admin:convert_non_mp4_videos:videofile.name := ', desired_relative_path)
+            if okay:
```
Woseseltops: This could be a guard clause
```python
 if not okay:
   continue
```

susanodd: Hahaha. It's obvious after you point it out. Baah!

Woseseltops:
> @Woseseltops the only thing I’m wondering about is doing all this file renaming inside of a possible loop. Will Ubuntu/Django finish these commands before the next iteration? Recall all the problems with the API running requests to upload videos every 5 seconds. They ended up not completed before the next request (for a package) was already coming in, and hence yet another request to upload the same video again.
> 
> Does Ubuntu perform these sequentially or will Django interleave commands with subsequent iterations? Or should I just ignore thinking about this?

I think each os.rename(), convert_video(), and shutil.move() call will fully complete before the next iteration begins :)

susanodd:
I rewrote the function as requested and it now has multiple return values.

I cam across another glitch. The admin function for converting moves the source (possibly renamed) video to the trash location (does not delete it).
But the "ensure_mp4" function that has been revived everywhere just converts the video and leaves the (uploaded) video (with the wrong type) in place. It ought to also move these to the trash location, otherwise there will exist extra files that are not accessible. This could end up capturing one of these in the case that a video is deleted and a previous version (reversion) is made the video.

The convert_video function takes two arguments, source and destination. So it should not be the one deleting the source video, obviously. So all of the "ensure_mp4" functions in all of the various video models ought to delete the video if it was converted.

PR #1658, opened 2025-11-13
Communication class and admin for email texts

susanodd:
Of the existing kinds of email subject and body text files, the various places that they are sent by the code, it looks like the case where a user has request change permission for a dataset does not send any email if this has been granted.

- Mail to dataset owner

- - new user granted view access (sent to inform the owner if the dataset is public, includes users motivation)
- - user has requested access (sent if the dataset is not public, this includes a link for the owner to go to the manage page to grant access for this user, includes motivation from the user

- Mail to user

- - welcome email to new users
- - access granted to public dataset (default dataset, plus other public datasets requested view access for)
- - access granted to dataset (mail on behalf of dataset owner, if the dataset is not public and the manager has approved)

Context variables:

signbank_name
url
user.firstname
user.lastname
user.email
user.username
new_user_firstname
new_user_lastname
new_user_email
dataset
motivation
site.name
https://{{site}}/datasets/manager/

### DIFF
--- a/signbank/registration/templates/registration/dataset_to_owner_new_user_given_access.txt
+++ b/signbank/registration/templates/registration/dataset_to_owner_new_user_given_access.txt
@@ -1,6 +1,6 @@
 A new user has signed up at {{site.name}}.
 
-{{user.firstname}} {{user.lastname}} (user {{user.username}}, email address {{user.email}})
+{{user.first_name}} {{user.last_name}} (user {{user.username}}, email address {{user.email}})
 has been given access to dataset {{dataset}}.
 
 {% if motivation %}


susanodd:
Here is what the "instructions help text" in the admin looks like:

<details>
add_communication_json_context_instructions

</details>

This includes all of the possible context variables that are used in all the template files for the various places where an email is sent.

Each of the places in the code creates a context for the email.

Since we don't know whether the fixtures initial content will be imported, there is no "starting block" for the various emails if the user wants to construct these from scratch.

Since the content of the context depends on the file and the place in the code, it would be best to make the data structure shown above have an additional set of keys above that is the label, then show the context variables dynamically based on what the user has chosen.

Issue #1666, opened 2025-11-27
Small videos


susanodd:
Small videos were originally created in order to display a low-res video when the user hovers over a gloss in the search results.

Small videos are not generated anymore because of a bug when generating them for some formats. (The code for generating them is in the CNGT scripts and was commented out in the videos code.)

Many small videos (`_small`) exist on disc, however. The small videos are not referenced by any objects.
It could happen that an "old" small video ends up becoming visible for a "newer" video that does not have a small video. (Because the gloss video files keep using the same naming convention.)

What do do about this situation?

There is a quite old command for generating small videos for a dataset.

PR #1668, opened 2025-11-27
Video files specific to gloss

susanodd:
Video files specific to gloss, preliminary functions to set up context data for view.
Display the video files relative paths in the Gloss Videos view, within collapsible panels.

The intention of this functionality is so that the dataset editors can clean up the stored files on a per gloss basis.
Details such as relative pathnames and checksums are shown. The various kinds of video files are in separate panels. This allows comparison of the checksum for the backup files to the primary file to double check the files can be removed. If more specific operations are required, the Gloss Video Admin can be used. There a user can select specific objects for deletion.

(FULLY FUNCTIONAL MOCKUP)

See the screenshots in the issue.

- the Video Storage is under a separate panel that only dataset editors can see
- the other kinds of media (Example Sentence videos for Senses, and Other Media) have previews of the media, these are also now in panels to keep the display consistent

Buttons have been added to:
- renumber version numbers on backup video objects
- remove obsolete files, those with wrong naming conventions

- button to remove backup files (some videos have up to 70 backup files)
- it is not possible to allow preview of these, since the extension must be a video file extension in order for it to play. - - a checksum column has been added so a dataset editor can confirm it's the same files before deleting

QUIRKS
<details>
- dangling video files without any object pointing to them. These exist but this is mainly a problem with the test dataset and the development servers since we delete stuff and try out code during development. However, this can arise during the time the "webm" videos were not being converted, and the time that the code has recently been revised to convert them. Originally it converted them and referenced the converted file but left the original file next to it in storage. Such files can be found and displayed to the user. This can also have happened with the original API code for the NME videos that were mysteriously being renamed. (Thus if these are found, then this was due to a code development quirk.) Small videos, however are still on disc (#1666) and since no new small videos are being created, a small video of an old gloss video could still exist for the gloss.
</details>

I added some checksums and functionality to show perspective videos with incorrect filenames.
(These appear in the panel "incorrect filenames" and since the filename is that of the primary video, when these are deleted, the filename is erased from the object to prevent the primary video file from being erased.)
(I simulated the file, since the file was not present locally, but in the database it has the wrong filename. I wanted to be sure the code can find such things. The checksums are included so the user can confirm that files are the same file, just attached to different objects. Basically in this example, the perspective videos are missing and the objects refer to the primary video.)

incorrect-perspective-filenames-with-checksum

### DIFF
--- a/signbank/dataset_operations.py
+++ b/signbank/dataset_operations.py
@@ -387,7 +387,7 @@ def file_display_preface(glossvideo):
     return str(glossvideo.version)
 
 
-def check_sum_relative_path(videofile):
+def get_check_sum_relative_path(videofile):
     if not videofile or not videofile.path:
         return ""
     fullpath = os.path.join(WRITABLE_FOLDER, videofile.path)
@@ -406,7 +406,7 @@ def get_primary_videos_for_gloss(gloss, string_result=True, include_subclasses=F
                                                 glossvideonme=None,
                                                 glossvideoperspective=None).distinct().order_by('version')
     if not string_result:
-        return [(gv.pk, gv.version, str(gv.videofile), check_sum_relative_path(gv.videofile)) for gv in glossvideos]
+        return [(gv.pk, gv.version, str(gv.videofile), get_check_sum_relative_path(gv.videofile)) for gv in glossvideos]
     display_glossvideos = ', '.join([str(gv.version) + ': ' + str(gv.videofile) for gv in glossvideos])
     return display_glossvideos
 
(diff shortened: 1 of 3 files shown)


susanodd:
`signbank-dev` synced up to here.

signbank/dataset_operations.py
```diff
@@ -418,38 +387,63 @@ def file_display_preface(glossvideo):
     return str(glossvideo.version)
 
 
-def get_primary_videos_for_gloss(gloss):
-    glossvideos = GlossVideo.objects.filter(gloss=gloss,
-                                            version=0,
-                                            glossvideonme=None,
-                                            glossvideoperspective=None).distinct().order_by('version')
+def check_sum_relative_path(videofile):
```
Woseseltops: A verb makes a function name often more descriptive

PR #1693, opened 2026-02-09
FieldChoie relation role

### LINKED ISSUE #1688: Add new items to RELATION_ROLE_CHOICES *OR* make RELATION_ROLE_CHOICES dynamic
Got a request from Rose:

> 
> My team and I are wondering whether it is possible to do the following for "Relation to Other Signs" in SignBank:
> 
>   - Ideally, give an option to allow us to add in options?
>   - If above is not possible, can you please add the following as subcategories in the dropdown:
>       - Negation
>       - Precision
> 
> We want to group signs based on related semantic/conceptual meaning - like all signs which relate to negation.

I am OK with both options. @Woseseltops @susanodd what do you think?

This is about RELATION_ROLE_CHOICES which are defined here:

https://github.com/Signbank/Global-signbank/blob/4ee6e3085f95a6e59045eb088a7d1aef912b9b88/signbank/dictionary/models.py#L2852-L2860

susanodd:
A similar approach is being used to introduce the field choice field as was done with the original introduction of field choices

**CODE IMPLEMENTATION**

<details>

- migration to create field role_fk in Relation as a FieldChoiceForeignKey
- population migration to store the constant choices into the new Field Choice RelationRole
- data migration of the role CharField into the new FieldChoice field
- retrieval methods combined into singular parameterised retrieval method
- updated gloss relations template to make use of dictionary lookup of relation types instead of hard coding
- CODE REVISION USING NEW FIELD TYPES, the original CharField has a character string as its domain, not an object, and the values now accessed using `role_fk.name`
- error checking on "add_relation" to prevent duplicates (plus checks that the role type exists in RoleRelation of FieldChoice, this should not happen because it's a model field of the form)
- CSV export and import Relations column revised to new type (checked on identical file being exported and imported without errors)
- relations added to gloss revision history, additions to revision history also for CSV update
- replaced redundant code with functions
- reverse role now a method of FieldChoice as well as Relation
- The usages of field role have been revised to use the new FieldChoice type
- Feedback shows the changes to other relations (reverse relations) in the overview of changes to do for the CSV (shown prior to the user performing "do changes")
- Tests on CSV update of relations, test on detection of erroneous input, and on creation of reverse relations

</details>

No deletion of relations performed during CSV update gloss.
[it's already implemented like this and it stays like this, the following example illustrates what happens.]

**EXAMPLE**

<details>

- row 1: gloss `TEMPORARYTESTGLOSS_1` updates its relations column to: `Synonym:TEMPORARYTESTGLOSS_2`

- row 2: gloss `TEMPORARYTESTGLOSS_2` updates its relations column to: `Hyponym:TEMPORARYTESTGLOSS_3`

1. when row 1 is processed, the state of gloss `TEMPORARYTESTGLOSS_2` has been updated with the reverse `Synonym` relation to `TEMPORARYTESTGLOSS_1`

2. when row 2 is processed, the initial state of gloss `TEMPORARYTESTGLOSS_2` now has the new synonym in it, but in the CSV the "new" state should be: `Hyponym:TEMPORARYTESTGLOSS_3`

The code shows the "side effects" to the user now in the "changes" (where the "do changes" button is offered prior to doing the changes).

</details>

OPEN

1. add method/model to set up whether a relation is symmetric, transitive, reversed with a different name in the other direction (this is partly hard-coded for different relation types)  (OPTIONAL, see issue description)
2. _**migration to remove the role CharField field and rename the new `role_fk` field to replace it**_

(2) must be done afterwards in a separate pull request


Woseseltops (commented):
Suprised to had so many consequences throughout the codebase! No need for another review after your processed this.

susanodd:
> It works!
> 
> One last UX (user experience) thing: would it be possible to hide/show the reverse_name_xx fields based on the selection at Reverse Relation Role?

That requires modification the the Django template itself. With the addition of jquery/javascript/trigger.

I wasn't able to do this in a straightforward way without that. Could this be a new issue?

vanlummelhuizen:
> Could this be a new issue?

Yes, let's make this an issue. Will you or shall I?

susanodd:
See #1727

Issue #1683, opened 2026-01-23
PhonologicalVariation


Woseseltops:
This is a large request by the VGT team, already discussed with @vanlummelhuizen a few months ago, to capture phonological variations within a sign. I propose these stages:

1. Adding and removing `PhonologicalVariations` to glosses

`PhonologicalVariation` is a Django object that contains all phonological fields glosses have. It belongs to a gloss, and can be added to this gloss in the UI. Importantly, it does not replace the existing phonological fields. This leads to these situations:

||Main phonological fields empty|Main phonogical fields filled|
|---|---|---|
|**0 PhonologicalVariations**|Similar to a new empty gloss right now|Similar to a filled out glos right now|
|**>0 PhonologicalVariations**|Somewhere in the phonology tab are arrow buttons to select which `PhonologicalVariation` you want to see|The main phonological fields are displayed as if the first `PhonologicalVariation`|

Perhaps this stage should be split into
a. Design UI (not trivial)
b. Implementation

2. Link videos to `PhonologicalVariation`

* `GlossVideo` stores to which `PhonologicalVariation` it belongs
* `phonological_variation` becomes a parameter in the API when uploading videos
* When looking at videos in the UI, only show the videos belonging to the `PhonologicalVariation` currently selected.
* When uploading videos in the UI, automatically link them to the `PhonologicalVariation` currently selected.

3. To discuss: remove the main phonology fields?

susanodd:
@Woseseltops for the Senses, the numbering was implicit. That allowed the user to reorder them.

Should the numbering here be implicit but numbered? (For the model in #1726)

susanodd:
Linguistics issue:

- the minimal pairs? what is the effect of this?

susanodd:
The phonological variations Phonology can be created and updated on `signbank-test.`

Perhaps @Woseseltops @vanlummelhuizen can consider where to "fit" these new videos...?

Or can they perhaps be created in the Gloss Videos  (navbar template) panel?

> * When looking at videos in the UI, only show the videos belonging to the `PhonologicalVariation` currently selected.

Are you sure about "hiding" the "other" variations and only showing the selected one?
It seems more likely the user will want to see them all next to each other to compare.

I suggest a hide/show button for the variations columns, so that the user can control what to see.

PR #1724, opened 2026-04-02
Lookahead instead of editable for Gloss update

### LINKED ISSUE #1714: Typing to call up value in all lists
Julie from ASL Signbank:

> In the tags I can type to call up a tag but not the rest of the lists, can we fix those so I can type to call up a value? Like for semantic field, word class, etc. And be able to hit “enter’ to commit changes?

Being able to hit enter would also solve:

> Sometimes the “cancel” and “ok” buttons are far on the side when screen is 100% and minimized. Can they move when screen is resized? But also hitting “enter” (like asked earlier in #6) would be good and make this a non-issue.

susanodd:
This pull request rewrites the Gloss Detail template by replacing the "editable" implementation of the fields.

This implementation is necessary to support  the phonological variations that will appear side by side in the Phonology panel (the panel is not wide enough using editables)

This is on the development server. It's impossible to describe what has been changed so far in so many words.

https://signbank-susan.cls.ru.nl/dictionary/gloss/3117

WHAT IT SOLVES

- offer lookaheads (as requested by this issue) as well as pull-down menus
- easier editing of multi-select Semantic Fields to resemble that of Tags (as requested by this issue)
- removes excess of OK / Cancel buttons used by editable (as requested by this issue)
- separate edit commands inside the various panels, no "global" edit button anymore
- the top area, including the videos panel on the left is turned on for editing via a "general" button (the various update areas needed to be encased in an element to limit the scoping context)

All of the edit panels are finished.

Implementation Set-up
<details>

- Separate save/update functions for different panels
- (per discussion with @Woseseltops) user profile setting to switch between original gloss detail edit and new (beta) gloss detail edit DOES NOT WORK, a TOGGLE is used in this template instead

- [ ] Revision History revision to store a lookup table mapping "fields" to human readable display (This exists, but it has become too unwieldly due to the increased number of variations of update methods in the code. It is now possible to update the related object fields "per object" - per row in the table. With editable, this was per table cell) The update methods work per related object and have the object id as part of the field name that is updated. This has been added to display method of the Revision table. (It's a bit haphazard.)

- [ ] This is to facilitate the development of #1741 to fit multiple variation columns under phonology. It's not possible to fit more columns under the current "editable" implementation. The development of phonological variations with editing inside the gloss detail, gloss edit phonology panel will build on this beta template. To avoid extremely large branches this is being developed first as a beta

- [ ] Separate update gloss functions that are not tied to "editable" when returning (switch to separate functions for specific update operations that make use of JsonResponse


susanodd:
To avoid excessive code duplication in different branches, it would be helpful if this branch could merge some of the forms, methods, and templates from #1726 (#1735) And VICE VERSE. As it goes, I've just been reusing some of the pieces of code, but included them elsewhere so as not to hinder merging.

susanodd:
I've got it so the single "edit" column is changing between black (disabled) and red (enabled).

Except the select fields are not changing to red. (This stopped when I added the disable to try to make use of a single column.)

I can't see yet how the select / option works to make it become red when edit is turned on. The lookaheads work.

https://signbank-susan.cls.ru.nl/dictionary/gloss/edit_phonology/42492

susanodd:
The development server `signbank-susan` is sync'd up to here.

susanodd:
The development server `signbank-susan` is sync'd up to here.

@Woseseltops try it out here:

https://signbank-susan.cls.ru.nl/dictionary/gloss/edit_phonology/42492

Next up:

- This is the one-column editing functionality that will also be used for phonological variants
- How to "subsitute" this code into the Gloss Detail template to replace the phonology and semantics? (and leave the other code the same) (NOT POSSIBLE. The entire template is being rewritten.)

Woseseltops:
Hey Susan, as you asked me to do a high level review of what you have so far: 

I see a lot of this

```js

var handedness_bloodhound = new Bloodhound({
      datumTokenizer: function(d) { return d.tokens; },
      queryTokenizer: Bloodhound.tokenizers.whitespace,
      remote: url+'/dictionary/ajax/fieldchoice/Handedness'+'/%QUERY'
    });

handedness_bloodhound.initialize();

function handednesstypeahead(target) {
     $(target).typeahead(null, {
          name: 'handedness',
          displayKey: 'name',
          source: handedness_bloodhound.ttAdapter(),
          templates: {
              suggestion: function(fc) {
                  return("<p><strong>" + fc.name +  "</strong></p>");
              }
          }
      });
}

var domhndsh_bloodhound = new Bloodhound({
      datumTokenizer: function(d) { return d.tokens; },
      queryTokenizer: Bloodhound.tokenizers.whitespace,
      remote: url+'/dictionary/ajax/handshape'+'/%QUERY'
    });

domhndsh_bloodhound.initialize();

function domhndshtypeahead(target) {
     $(target).typeahead(null, {
          name: 'domhndsh',
          displayKey: 'name',
          source: domhndsh_bloodhound.ttAdapter(),
          templates: {
              suggestion: function(fc) {
                  return("<p><strong>" + fc.name +  "</strong></p>");
              }
          }
      });
}
```

Why not

```js
// Define all typeahead configurations in one place
const typeaheadConfigs = [
    { name: 'handedness', endpoint: 'fieldchoice/Handedness' },
    { name: 'domhndsh', endpoint: 'handshape' },
    { name: 'subhndsh', endpoint: 'handshape' },
    { name: 'handCh', endpoint: 'fieldchoice/HandshapeChange' },
    { name: 'relatArtic', endpoint: 'fieldchoice/RelatArtic' },
    { name: 'locprim', endpoint: 'fieldchoice/Location' },
    { name: 'contType', endpoint: 'fieldchoice/ContactType' },
    { name: 'movSh', endpoint: 'fieldchoice/MovementShape' },
    { name: 'movDir', endpoint: 'fieldchoice/MovementDir' },
    { name: 'relOriMov', endpoint: 'fieldchoice/RelOriMov' },
    { name: 'relOriLoc', endpoint: 'fieldchoice/RelOriLoc' },
    { name: 'oriCh', endpoint: 'fieldchoice/OriChange' },
    { name: 'namEnt', endpoint: 'fieldchoice/NamedEntity' },
    { name: 'valence', endpoint: 'fieldchoice/Valence' }
];

// Factory function to create bloodhounds
function createTypeahead(config) {
    const bloodhound = new Bloodhound({
        datumTokenizer: function(d) { return d.tokens; },
        queryTokenizer: Bloodhound.tokenizers.whitespace,
        remote: `${url}/dictionary/ajax/${config.endpoint}/%QUERY`
    });
    
    bloodhound.initialize();
    
    return function(target) {
        $(target).typeahead(null, {
            name: config.name,
            displayKey: 'name',
            source: bloodhound.ttAdapter(),
            templates: {
                suggestion: function(fc) {
                    return `<p><strong>${fc.name}</strong></p>`;
                }
            }
        });
    };
}

// Initialize all typeaheads
typeaheadConfigs.forEach(config => {
    window[config.name + 'typeahead'] = createTypeahead(config);
});
```

Or perhaps the JS is not aware of the individual fieldnames at all, but it's just something it can take from html tag properties? Something along the lines of

```js
// Single factory function that reads from HTML data attributes
function initializeTypeaheadFromElement($element) {
    const name = $element.data('typeahead-name');
    const endpoint = $element.data('typeahead-endpoint');
    
    if (!name || !endpoint) return;  // Skip if missing attributes
    
    const bloodhound = new Bloodhound({
        datumTokenizer: function(d) { return d.tokens; },
        queryTokenizer: Bloodhound.tokenizers.whitespace,
        remote: `${url}/dictionary/ajax/${endpoint}/%QUERY`
    });
    
    bloodhound.initialize();
    
    $element.typeahead(null, {
        name: name,
        displayKey: 'name',
        source: bloodhound.ttAdapter(),
        templates: {
            suggestion: function(fc) {
                return `<p><strong>${fc.name}</strong></p>`;
            }
        }
    });
}

// Find all typeahead fields and initialize them
$(document).ready(function() {
    $('[data-typeahead-name]').each(function() {
        initializeTypeaheadFromElement($(this));
    });
});
```

Other examples of making the code more DRY:
* `renderSelectedSemField()` and `renderSelectedDerivHist()` are almost identical.
* Same for `toggle_edit_semantics` and `toggle_edit_phonology`

susanodd:
@Woseseltops I see the opportuntiies to make the code more concise.

But there's a major problem with the Gloss Edit template (see description of the problem in the issue).

<details>
(The entire setup of Gloss Edit replies on a "globallly scoped" single Edit button and the very small locally scoped field-update editable ajax calls. It does not work to "move the global edit button" because all of the contextual nesting that is made use of is gone and the inner elements have not outer scope. Moreover, everything is dynamic and it does not work to "attach" functions to things that end up disappearing. The "configure editable" function actually does this every single time that Edit is clicked on. So merely copying the functionality to make it more local does not work. There would need to be a configure function for every single smaller panel.

susanodd:
@Woseseltops this is not working at all.

I think the phonology and semantics needs to be in a separate template, not the Gloss Edit.

The Gloss Edit code is horrific.

SOLUTION

Totally rewrite the Gloss Detail / Gloss Edit template from scratch.

<details>
I'm working on this by step by step adding the other panels to the template.
This is necessary because of the (newly required to achieve modularity!) intricate hierarchy and hierarchical class- and type-based visibility when switching to edit mode for the separate (to become modular!) panels.

At the moment, none of the video panels nor the senses are being included.
Nothing is modular wrt editing. That is causing extreme problems because "all" of the diverse editing areas are not modular, they are globally turned on and off. Mainly by trickery.

susanodd:
@Woseseltops, as written above, I am totally rewriting the Gloss Edit template. This is necessary because it is not modular.

I shall make the improvements (hopefully) as you suggest. However, because of the extreme control flow and visibility (on / off edit functionality) and the need for this to work in a modular way, I can't do these improvements yet until all the code "learns to function independently". I've revised it by adding extra "category" indicators which are needed also by all the stupid "edit on/off" things.

<details>
To get this to actually "work" (I gave up editing the gloss_detail.html template yesterday), I have made a new template with the phonology and semantics, and am adding the "general" fields that appear above. Even getting that to work is time consuming!! It's all the toggles that need to be turned on and off. Let alone all the categories etc.) I excluded the videos (left panel) and the senses (1000 lines inside one table row!). Those will need to be made "modular" before inserting them into the rewritten template (right column)

Ideally, the lists of fields for the different panels will need to be separated so as to not be hard-coded. But this is impossible until the code (albeit hard-coded) actually does the control flow. It's extremely difficult to do this.

I would suggest making the videos be "editable" inside the separate Gloss Videos template, and only be visible in the left column. (The editing of the videos is also part of the giant editable scheme of turning stuff on and off, and class names are reused.)

susanodd:
The first version of the major rewrite of the gloss detail template has been pushed.

Only the "main" fields, the phonology and the semantics are shown now. (With lookaheads)

The Lemma editing has been revised to correct control flow bugs. (Some code was removed and replaced with hard-coded hide/show code per operation.)

==============

What is causing major problems is the loaded templates. These are causing the events assigned to buttons for Edit to disappear. If I remove these lines, then the code has button events, otherwise the events disappear.

But it does not work to just copy this code into the template instead of loading it. Something is messed up with this code. It could be that the guardian settings are loaded again and again. Also inside of Gloss Detail, the guardian tags keep being loaded, also with different variable names. (This is the original code, I did not cause these bugs. I suspect the original code kept using new variable names to not catch ones in outer scopes.)

https://github.com/Signbank/Global-signbank/blob/3dcf91e819cb837092475d3079d7d4448dce4cdd/signbank/dictionary/templates/dictionary/gloss_detail.html#L1350-L1351

susanodd:
SOLUTION TO SCOPING PROBLEMS (original Gloss Detail not being modular)

These two loaded files have been put into the code INLINE:

https://github.com/Signbank/Global-signbank/blob/3dcf91e819cb837092475d3079d7d4448dce4cdd/signbank/dictionary/templates/dictionary/gloss_detail.html#L1350-L1351

The "include" was causing the events on the element ids in the main template not to be recognised anymore on the buttons.
The loaded template filters inside these files have been moved to the top of the main template.

susanodd:
@Woseseltops I'm not entirely sure that it works to create the "constants" dynamically like that. (Don't they need to be in an outer scope so they exist after the function is applied, as a global declaration?)

There's some discussion about this issue below. I guess it's okay to try it out. Except that is a risk if it's a browser problem. I've been implementing all of this on Safari and on Firefox. Only some of the fields needed adjustment, but it was mainly for initialising the fields. (I made extra methods in the models for ensuring the internal representation can be compared to that in the choices. Those methods work like conversions, but are also able to be displayed, since we're using one column now....)

<details>
There were some of those in the original gloss edit javascript, but they were broken. There was additional code that was hard coded as well.

The problem is that some things don't exist when the page is generated by Django. The lookaheads need to be added dynamically. But some of it has to "exist". It can't be 100% dynamic. It's like the chicken or the egg. Something has to already exist to be attached to other things. Otherwise it's circular with non-existent dynamic stuff being attached to other non-existent dynamic stuff.

That's what was going on with the events being attached to elements. Sometimes the elements don't exist yet. It wasn't working to attach control flow events to buttons, for example. That's why those "include" files needed to be hard-coded.

</details>
======

<details>
Okay, most of it is implemented now. I've done additional things with CSS and classes for the various panels.
Lots of this is hardcoded.
The problem is that Django has to load the code before it exists. So I had problems getting it to apply things. I've implemented a lot of "toggling" type stuff based on the various panels, because they must work independently of each other.

There remains a single "save" url ajax call for phonology and semantics, because I didn't split up the python method (new for this issue). It seems that because unique ids are used everywhere that if the user has more than one panel open and clicks the save, then the panels that share the same save are all updated. (I haven't studied this yet. The ones that involve relations between different classes (like notes and updating videos) have their own buttons, since they don't just update gloss fields.

It is extremely tedious to make sure all of the panels actually operate. The code for turning stuff off and on cannot be put inside one method. It must find the exact elements and the classes. I'll see about the bloodhounds. But the question is when does Django initialise these things? I did the "disabling" partly by using css to disable the input fields. (with patterns of classes and matching ids).
I did this enabling disabling to make sure it's only in one column (!)

Django had huge problems with the tags and affiliations code being inside other html files. It can't work that way because no idea. But Django or the browser ended up removing all the events that were in the same area as those files. (It's not clear what happens when html is loaded from another file by Django. This has been the reason that Gloss Detail is so huge.) The bloodhounds have translations on all of the choices. So I'm not sure about that either. Need to investigate.

susanodd:
Development server `signbank-susan` is sync'd up to here.

susanodd:
@vanlummelhuizen I have added some code with a setting in the user profile to be able to switch between the new template and the original template.

I undid all of that because it was not working at all.

I think it's whatever was done to the STATIC setting...... #1749

It's not finding the media files anymore.

<details>

(The original gloss_detail.html template).

However, when it is supposed to be using the original template, it finds errors in this template. However, the errors it is reporting is nothing in this pull request. (It says that a javascript variable is not defined, except the code is the same as it was originally.)

Is it possible that "github" is not showing the "complete" differences? Perhaps something is changed in the template file, but github is not showing this?

I can't find any information by searching (also not from CoPilot that can explain why the browser is complaining about something that is not in the file.

If you try out the development server you will see what I mean. This is really annoying. Since github doesn't think anything has been changed in the file, it also doesn't work to fetch it again. (I want to get the one that is on master to make sure nothing has changed in it, but it thinks that nothing has changed so won't get the file)

vanlummelhuizen:
@susanodd I see you made big changes. I have some UI/UX issues:

The look-ahead menu is not always shown immediately. E.g. when you start editing an empty field, you have to type in a letter and then remove it (e.g. with backspace) to see the full list. 

When you can use Enter to commit, I would expect Esc to cancel. 

For e.g. Stronghand, in the old version, both letter and number could be either true or false. In the new version the user can select either letter or number. These are not the same thing.

susanodd:
> @susanodd I see you made big changes. I have some UI/UX issues:
> 
> The look-ahead menu is not always shown immediately. E.g. when you start editing an empty field, you have to type in a letter and then remove it (e.g. with backspace) to see the full list.
> 
> When you can use Enter to commit, I would expect Esc to cancel.
> 
> For e.g. Stronghand, in the old version, both letter and number could be either true or false. In the new version the user can select either letter or number. These are not the same thing.

Yes, I checked NGT to try to find if letter and number can both be true at the same time. (No glosses do this.)
It's still the same in the database, but I modified it in the form. I can restore it to what it was. @Woseseltops didn't like these fields, because it takes up lots of the horizontal space and will force the Variations to be too wide.

I am aware of the lookahead. 

(Resolved in subsequent code changes, original text below)

<details>
I modified the original typehead to change the limit, since at first it was only showing 10 things and it did not work to use the length setting in the typeahead initialisation. (I found a modification on stackexchange because it was something in the original code.)

I know it doesn't work to show the entire menu. It works in the ajax call to retrieve all the values, so it is something in the typeahead code itself. (Did not figure out how to solve this so far.)

It does not commit changes until you use the Save button for the respective panel. (So it undoes everything if you do cancel - don't save.) I will check out how to make use of ESC. (I had not idea that was something people do.)

The Save buttons need to be separated into separate methods for Semantics and Phonology. (Those two are still linked as that's how I was developing it at first in a different template. That ended up not being possible to "load". It all has to be in the same template. We also need all the translations inside one template.

susanodd:
UPDATE

After a lot of tweaking and trial and error, I got it so when you click in the lookahead field, the already selected choice is selected in the field. (This actually looks nice. It changes colour from black to red when you switch to edit mode.) But you actually need to click/select it and erase it in order to type something else. This seems as expected to me.

Still working on fields where nothing has been selected yet. It wants to select the first "-" choice. That is what is happening to it. The value is 0 and the placeholder is "-" (so it has put something there already, pre-selected). I guess I need to figure out how to make it choose nothing instead of 0 if it's empty. It's technically an input text field, so it should not be None (if the gloss field is None, then it sets it to the default of 0, but that is "-" again. I'll try the same approach to force it to be '' although it wants that to be the name of the 0 machine value.

susanodd:
ANOTHER UPDATE

Per review, the lookahead fields now show the already selected field when you're not looking.
They show up black when you're not editing, and when you switch to edit mode, they show up red. These are actually the selected choice that is showing up. In the previous version, this was the placeholder.

To edit the lookahead field, you select and backspace the shown field (which is the one that is already stored).

For this, the "-" that shows up is actually the "chosen" choice with machine value 0. This also needs to be backspaced and then you can choose a new one. (To get rid of the "-", we can modify the ajax field choice retrieval to just put "" for machine value 0. But that probably won't work correctly because that is empty string, not a choice.)

Clicking on the lookahead does not erase the chosen field (of the gloss). To select a different choice, you need to "erase" the one that is selected.

(This is how it works. This is also logical. It just doesn't work the same as the master branch with editable. Note that with editable, once, you choose something, you click OK. Here there is no OK button anymore. You can change what you want and then do Cancel to not save anything.

susanodd:
Got the lookaheads working now! (Newest code not on development server yet...)

- Ignore the excess code. I had problems with the lookaheads being behind other panels and tried to fix it. But it only happens if your display is too narrow. It was not possible to fix this with z-index because the browser overlaps things that have not been displayed yet, apparently.
- Also ignore the spurious extra pieces of code. The extra parameters seemed to be doing nothing at all for the lookahead. CoPilot makes some suggestions that don't work.
- I added a trigger if the machine value is 0. Although this is a bit disguised where the trigger needs to be. This needs to be done programmatically with jQuery, not as a lookahead setting. (See hidden comments above about the machine value 0 and '-' being a chosen value, that's why it is visible, and why it needed to be erased. The new trigger solves this.)
-  I also added code for setting up that the lookahead choice is set to the field choice when initialised. (See hidden comments)

vanlummelhuizen:
@susanodd It is still not working correctly/intuitively on my end. 

When I click for instance on `Edit Phonology` en then on the *empty* field for `Strong hand`, I don't see a list of options. After I type something, something may appear, if it matches one or more items in the list. When I remove everything from the field, a full list appears. 

----

Note that a lot of the desired functionality is covered by this Javascript package: [Select2](https://select2.org/getting-started/basic-usage/). There are Django packages that use Select2: https://django-select2.readthedocs.io/en/stable/ (used in [Django admin from version 2.0](https://docs.djangoproject.com/en/2.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.autocomplete_fields)) and https://django-autocomplete-light.readthedocs.io/en/master/alight.html#create-an-autocomplete-view. 

@Woseseltops, do you think Select2 is an option here?

susanodd:
@vanlummelhuizen are you sure?

I just updated it on `signbank-susan` and it works.

It also works locally. (Firefox on both.)

susanodd:
Select2 is a totally different thing. All the search templates that use it have been specially adjusted for Signbank. It's intended for multiple select.

susanodd:
This works on Safari, Firefox (Ubuntu), and Chromium (Ubuntu).

Is it possible you have an entirely different mouse?

I made it respond to the "focus" event. Perhaps this is what's not working, perhaps it needs to be "click" instead.

It's very weird that it works on 3 different browsers, but not on your computer.

I'll play the opposite to the Devil's Advocate, I can't duplicate your error. :)

vanlummelhuizen:
It is also not working in Chrome 149.0 on my Android 14.

> When I click for instance on `Edit Phonology` en then on the _empty_ field for `Strong hand`, I don't see a list of options. After I type something, something may appear, if it matches one or more items in the list. When I remove everything from the field, a full list appears.

After the procedure above, I do get a menu immediate if I blur and focus the field again.

vanlummelhuizen:
@Woseseltops and I found out that some glosses are working while other don't:

Working: https://signbank-susan.cls.ru.nl/dictionary/gloss/4483
Not working: https://signbank-susan.cls.ru.nl/dictionary/gloss/50206/

We don't know why...

susanodd:
Something went wrong with the css file. PyCharm all of a sudden was complaining about the syntax. I tried to fix it but it is somehow wrong now. It's been too long a day to fix it now.

(This is on this branch, not the server. I need to undo the most recent changes. The pulldown menus are appearing behind the panel headers and I tried to fix this again. z-index does not work at all. And now the menus have become really wide for some reason.)

The newer version of typeahead (the one right after the one we use) does not work!

I "pretty formatted" ours (v 0.10.0) and compared it to a "pretty formatted" one of v 0.10.1  
Just one bump in the version number makes it not work anymore.

This applies to ALL of the templates that use lookahead.

vanlummelhuizen:
I think that if a field has None as value, nothing happens on focus. If there is a dash (-) it works as intended.

vanlummelhuizen:
The following line is not working for the `String hand` field if there is no value attribute in the input element:

https://github.com/Signbank/Global-signbank/blob/340c4187a6014f8be33416b3037f976e6b7bbfb6/media/js/lookaheads.js#L1143

vanlummelhuizen:
> The newer version of typeahead (the one right after the one we use) does not work!
> 
> I "pretty formatted" ours (v 0.10.0) and compared it to a "pretty formatted" one of v 0.10.1 Just one bump in the version number makes it not work anymore.
> 
> This applies to ALL of the templates that use lookahead.

This could be a reason to abandon typeahead.js altogether and opt for something well maintained. Like Select2. Note that Select2 is not only for multiple select, as you mentioned earlier.

susanodd:
> The following line is not working for the `String hand` field if there is no value attribute in the input element:
> 
> https://github.com/Signbank/Global-signbank/blob/340c4187a6014f8be33416b3037f976e6b7bbfb6/media/js/lookaheads.js#L1143

I had switched the initial field values from '-' to '' previously. That is what was causing the error. I also had problems with an empty word class field. It's fixed now (on this branch, not on the server)

susanodd:
> > The newer version of typeahead (the one right after the one we use) does not work!
> > I "pretty formatted" ours (v 0.10.0) and compared it to a "pretty formatted" one of v 0.10.1 Just one bump in the version number makes it not work anymore.
> > This applies to ALL of the templates that use lookahead.
> 
> This could be a reason to abandon typeahead.js altogether and opt for something well maintained. Like Select2. Note that Select2 is not only for multiple select, as you mentioned earlier.

That needs to be a new issue since it affects numerous pages. I was thinking to do a simple replacement, but that isn't possible.

<details>
It's very weird that the version we are using seems not available online anymore. I suspect the Bloodhound class is inside of it, because we don't import that as far as I can tell. Perhaps the version we have was the initial input to all the numerous things available. (Akin to good old Berkeley Unix that was open source.)

We are only using Select2 for multiple select as far as I know. (Because the javascript has to do stuff to the choice lists that are fed into the Select2 code. From recollection. There is some code that searches out "ids" used inside the Select2 code itself. But this also applies to the twitter lookahead code and Bootstrap code. Some of this has to do with dynamic translations and colours. It could be remnants from before we used model translations and all the choice lists had to be generated dynamically, then replace the original choices. There is also code that adds frequencies to the lists. I ignored the frequencies and colours in this issue. Adding the frequencies would be an issue. Those actually need to obtain this from the database, not just the model.)

susanodd:
@Woseseltops @vanlummelhuizen 

I just set up a toggle between lists and lookaheads in the new Gloss Edit.

It's only set up for one field now, Word Class in the General panel.

It involves a toggle buttons, fetching this in the view, setting a session variable (this is all needed to set the context and initialise the form and the javascript properly.)

The form makes use of this to initialise the form with the appropriate field type. The javascript either sets up the lookaheads or uses the normal lists (in the "save" method it finds the appropriate field value)

This works!

(I'm happy)

Removed a bunch of redundant code.

susanodd:
UPDATE

I'm still working on the toggle between the two field access. On all sides, the forms, the template, and the javascript/jquery.

This also requires additional extra data for use in the Cancel operation. The typeahead fields require more data in order to reset them.

susanodd:
@Woseseltops @vanlummelhuizen 

I set up a toggle between lists and lookaheads in the new Gloss Edit. This is implemented for all of the fields.

- The Cancel buttons now do a proper reset of the fields. (It does not reload the page.)

- Toggling between Lookaheads and Lists reloads the page, since the forms need to be regenerated with the correct types and initial values (needed for reset/cancel).

- This branch is current. The development server is not yet up to date.

- The typeahead javascript (`typehead.bundle.min.js`)  is still as it was. (But this branch changes some length and limit fields that were hard-coded in the original source file.)

- The setup of things in the javascript has been simplified as per review @Woseseltops and applying the same to the activation in a different code.
- The original "gloss", "lemma", "sense" and "morpheme" lookaheads are still as they were originally (they also appear in other templates). 

- Some of the toggles have lots of duplication. I revised the element ids to create more patterns, but they have not all been abstracted into single commands yet. Some commands have been combined using patterns. (It is difficult to debug this because the goal is to keep the panels operating independently. Some of this is via panel-specific css combined with disable- attributes.)

- Q: It's sometimes unclear to me when something needs to happen in the "ready" or can be outside of it. The "cancel" button did not work properly when outside of the "ready" function.

susanodd:
Regarding typeahead, I was intending to revise the (original) code to a non-"min" version so it would at least be readable. I did a pretty print (not pushed) to try to spot the differences with subsequent versions. The code is extremely cryptic and unreadable. (However the "editable" code also started out as a "min" version and I modified that once upon a time to add the colours.)

vanlummelhuizen:
We probably should not waste time trying to use a slightly newer version of an already very outdated library that is Twitter's typeahead. I created an new issue for phasing out typeahead: #1751 .

susanodd:
The development server is up to date `signbank-susan`

Make sure to logout and login, since it didn't want to load the newest javascript for me at least.

I need to rewrite some of the tests for gloss update.... (Since the tests have been removed from deploy this is kind of hidden.)

LATER

I've been trying to update the layout since the edit fields are sometimes outlined in red and something not clear where to click since there are no outlines.

There are some glitches with the reset of the multiselect fields. (Semantic Fields).

I revised the code to make it not be hard-coded, but this is causing some problems since the left-and side can't be a variable. (I followed the instructions for being able to do this via functions. But now it seems to be erasing the "state" but only the stuff in javascript, since nothing is updated until you do a save. So it seems my clever solution is buggy. It worked fine when it was hard-coded.)

I have variables for both the "initial" state and the "working" state for the fields. The code is supposed to obtain the "initial" data (a list of dictionaries of the choices, name, machine_value). But something is going wrong and it is losing the data.
Probably the variables are sharing too much.

This is probably caused by the recursion....

vanlummelhuizen:
@susanodd I was looking at the PR and I must say, the number of changes is quite extreme. 

<details>
<summary>Diff summary</summary>

```
❯ git diff --stat master..HEAD
 media/css/style-twilight.css                                           |    4 +-
 media/css/style-twilight_bad.css                                       | 1463 ++++++++++++++++++++++++++++++++++++++++++++++++
 media/js/lookaheads.js                                                 | 1153 ++++++++++++++++++++++++++++++++++++++
 media/js/typeahead.bundle.min.js                                       |    2 +-
 signbank/dictionary/admin.py                                           |    2 +-
 signbank/dictionary/adminviews.py                                      |   95 +++-
 signbank/dictionary/bloodhounds.py                                     |   49 ++
 signbank/dictionary/context_data_gloss.py                              |    2 +-
 signbank/dictionary/display_functions.py                               |   54 ++
 signbank/dictionary/field_choices.py                                   |    1 -
 signbank/dictionary/forms.py                                           |  421 +++++++++++++-
 signbank/dictionary/gloss_revision.py                                  |   42 +-
 signbank/dictionary/models.py                                          |  200 ++++++-
 signbank/dictionary/templates/dictionary/admin_batch_edit_view.html    |    2 +-
 signbank/dictionary/templates/dictionary/edit_gloss.html               |  669 ++++++++++++++++++++++
 signbank/dictionary/templates/dictionary/gloss_detail.html             |    2 +-
 signbank/dictionary/templates/dictionary/gloss_detail_lookahead.html   | 4752 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 signbank/dictionary/templates/dictionary/minimalpairs_gloss_table.html |    8 -
 signbank/dictionary/templatetags/field_choice.py                       |    8 +
 signbank/dictionary/update.py                                          |  541 ++++++++++++++++--
 signbank/dictionary/urls.py                                            |   27 +-
 signbank/dictionary/views.py                                           |   63 ++-
 signbank/registration/templates/user_profile.html                      |    8 +-
 signbank/tools.py                                                      |    7 +
 24 files changed, 9467 insertions(+), 108 deletions(-)
```

</details>

I think some things are not necessary:

- `media/css/style-twilight_bad.css`: if not used, please do not commit.
- `signbank/dictionary/templates/dictionary/gloss_detail_lookahead.html`: a lot of the code is the same as `signbank/dictionary/templates/dictionary/gloss_detail.html`. Is it necessary to have both?
- `media/js/typeahead.bundle.min.js` is changed, but since it is a minimized version, it should not be changed.

susanodd:
(obsolete code removed)

I need to do something about the original "update_gloss" method. It is overused because there are so many calls that branch off and handle the relations with other models than Gloss. The new template makes use of all of those "branch off" methods. I didn't figure out a best way to organise this yet. It should be removed and replaced by all the "branched" methods directly.

The original method is probably okay, all of the endpoints return results not json, so they need to be handled differently. (I did not implement this part yet, since I was aiming to not completely rewrite the code.)

The "updates" end up reloading the page. Only the "Cancel" does not reload the page. There is some complicated handling of results by the original editable code. That has not been transcribed here yet. (The issue is getting too big.)

LOG MESSAGES VIA SETTING
<details>
I predicated the log messages on a setting at the top of lookaheads.js to facilitate debugging. It's off now but the log messages are still in the code. I got tired of adding and removing these. The control flow is complicated. I resolved it with an extra event on the parent (button) of the "x" delete inside the button. That makes a call to the global method. The "remove" button was somehow not causing the unselecting to trickle to other code. Maybe it was stuck with its context. No idea, hence all the logs. It's working now.)
</details>

RESOLVED:
<details>
There are some problems with the multi-select fields `dialect`, `semField`, and `derivHist`. In the start of the branch development, they were hard-coded. But then with the comments of @Woseseltops, I made it so they share code. But this caused other problems because it uses global variables to store the initial state and the editing state of the fields that are selected. When the different kinds of fields were abstracted into a general method this causes problems updating the global variables.
I've added an extra signal to the remove (field) button (edit mode) and some global methods on the global data.
But these require explicit reference to the global data (not a variable).

I know this could be a better setup with javascript classes, but this is a Django application and these are just there dynamically.

susanodd:
@vanlummelhuizen the branch is working correctly now. I removed unused code.

I need to continue other work starting from this, i.e., use this branch as the basis of the phonetic variations branch. They need to be placed into the Gloss Detail template, not as a separate template. But a first step is adding the new model to this branch. Then extra Phonology columns for the variations.

How to proceed? This branch is needed by the other branch, but the other branch will become extremely large. Do you want to review this branch and allow it to be merged? Or shall this branch become part of the Variations branch?

susanodd:
The development server `signbank-susan` is sync'd up to here.

susanodd:
The development server `signbank-susan` is sync'd up to here.

susanodd:
ALL OF THE TESTS PASS !!

susanodd:
The development server `signbank-susan` is sync'd up to here.

susanodd:
@Woseseltops 

- the `lookaheads.js` file needs a better name (you need to read that, it's all the javascript for editing and the method invocations (this is the code I wrote, it's not imported from libraries. The filename is probably confusing, but I had no idea how much would need to be revised when I started out, hence the name.)
- the "save" methods for the phonology, semantics, use the same python handler since they all operate on gloss fields (should these be split into the panels to separate them? as it works, the user can work in different panels and clicking save saves all the ones in edit mode. For the non-relation panels. The buttons are different for those, Dismiss instead of Cancel. Those have row-based edit buttons. (You'll see this yourself when you try it out.)
- the "`?edit`" url parameters have been revised to be done in the event handler for the ajax sucess via a storage variable
- the colors of the pull-downs has not been carried over (the editables have a hack in the code to achieve this) (this could be added to the lookaheads, and to the choices in the form initialisations when in list mode to carry over to the template)

signbank/dictionary/models.py
```diff
@@ -1166,6 +1260,16 @@ def ordered_senses(self):
...
+    def repeat_to_choice(self):
+        if self.repeat is None:
+            return '0'
+        return '1' if self.repeat else '0'
+
+    def altern_to_choice(self):
+        if self.altern is None:
+            return '0'
+        return '1' if self.altern else '0'
+
```
Woseseltops, line 1272: Having generic functions for all booleans (not just these two) is better

```python
def _bool_to_choice(self, value):
    if value is None:
        return '0'
    return '1' if value else '0'
```

Call it like

```python
self._bool_to_choice(self.altern)
```

media/js/lookaheads.js
```diff
@@ -0,0 +1,1185 @@
...
+    $('#enable_edit_othermedia').show();
+}
+
+$(".othermedia-edit-dismiss").on('click', function() {
+    $('.edit_only_othermedia').hide();
+    disable_editing('othermedia');
+    $('.read_only_othermedia').show();
+    $(this).hide();
+    $('#enable_edit_othermedia').show();
+});
```
Woseseltops, line 318: I'm seeing a lot of copying and pasting here. I think it should be possible to do something like this

```javascript

function setupEditToggle(section) {
  $(`#enable_edit_${section}`).on('click', function() {
    $(`.edit_only_${section}`).show();
    $(`.read_only_${section}`).hide();
    $(`#${section}_edit_dismiss`).show();
    $(this).hide();
    enable_editing(section);
  });

  $(`#${section}_edit_dismiss`).on('click', function() {
    $(`.edit_only_${section}`).hide();
    $(`.read_only_${section}`).show();
    disable_editing(section);
    $(this).hide();
    $(`#enable_edit_${section}`).show();
  });
}

['provenance', 'othermedia', 'annotated_sentences'].forEach(setupEditToggle);
```

Or an alternative idea, hide the values in the html, so you don't need to touch the JS every time a value is added:

```html
<button id="enable_edit_provenance" class="enable-edit-toggle" data-section="provenance">Edit</button>
<button id="provenance_edit_dismiss" class="edit-dismiss-toggle" data-section="provenance">Dismiss</button>

<button id="enable_edit_othermedia" class="enable-edit-toggle" data-section="othermedia">Edit</button>
<button id="othermedia_edit_dismiss" class="edit-dismiss-toggle" data-section="othermedia">Dismiss</button>

<button id="enable_edit_annotated_sentences" class="enable-edit-toggle" data-section="annotated_sentences">Edit</button>
<button id="annotated_sentences_edit_dismiss" class="edit-dismiss-toggle" data-section="annotated_sentences">Dismiss</button>
```

and the JS becomes

```javascript

$(document).on('click', '.enable-edit-toggle', function() {
    const section = $(this).data('section');

    $(`.edit_only_${section}`).show();
    $(`.read_only_${section}`).hide();
    $(`#${section}_edit_dismiss`).show();
    $(this).hide();

    enable_editing(section);
});

$(document).on('click', '.edit-dismiss-toggle', function() {
    const section = $(this).data('section');

    $(`.edit_only_${section}`).hide();
    $(`.read_only_${section}`).show();
    $(this).hide();
    $(`#enable_edit_${section}`).show();

    disable_editing(section);
});

```

media/js/lookaheads.js
```diff
@@ -0,0 +1,1185 @@
...
+         update['foreignrelation-loan_'+foreignrelationid] = $('#foreignrelation_loan_'+foreignrelationid+'_select_value').val();
+         update['foreignrelation-other-lang_'+foreignrelationid] = $('#foreignrelation_other_lang_'+foreignrelationid+'_text').val();
+         update['foreignrelation-other-lang-gloss_'+foreignrelationid] = $('#foreignrelation_other_lang_gloss_'+foreignrelationid+'_text').val();
+         $.ajax({
+            url : url + "/dictionary/update/update_gloss_foreignrelation/" + glossid + "/" + foreignrelationid,
+            type: 'POST',
+            data: update,
+            datatype: "json",
+            success : function(data) {
+                if (data.success) {
```
Woseseltops: Here too a lot of copypasting... example JS that uses values from the HTML:

```html
<button class="quick-save"
        data-glossid="{{ gloss.id }}"
        data-recordid="{{ othermedia.id }}"
        data-url-template="/dictionary/update/update_gloss_othermedia/{glossid}/{recordid}"
        data-fields='[
            {"key":"other-media-type_{recordid}","selector":"#othermedia-type_{recordid}"},
            {"key":"other-media-alternative-gloss_{recordid}","selector":"#othermedia-alternative-gloss_{recordid}_text"},
            {"key":"other-media-description_{recordid}","selector":"#othermedia-description_{recordid}"}
        ]'>
    Save
</button>
```

```javascript
function quickSaveAjax(url, data) {
    data['csrfmiddlewaretoken'] = csrf_token;
    $.ajax({
        url: url,
        type: 'POST',
        data: data,
        dataType: 'json',
        success: function(resp) {
            if (resp.success) {
                setTimeout(function() {
                    location.reload(true);
                }, 500);
            }
        },
        error: function(xhr, status, error) {
            alert("There was an error processing this change: " + xhr.responseText);
        }
    });
}

$(document).on('click', '.quick-save', function(e) {
    e.preventDefault();

    var $btn = $(this);
    var glossid = $btn.data('glossid');
    var recordid = $btn.data('recordid');

    var url = $btn.data('url-template')
        .replace('{glossid}', glossid)
        .replace('{recordid}', recordid);

    var fields = JSON.parse($btn.attr('data-fields'));
    var update = {};

    fields.forEach(function(f) {
        var key = f.key.replace('{recordid}', recordid);
        var selector = f.selector.replace('{recordid}', recordid);
        update[key] = $(selector).val();
    });

    quickSaveAjax(url, update);
});
```

Woseseltops (commented):
Thanks @susanodd , this must have been a lot of work! My comments are indeed all related to being DRY, as you already predicted.

I agree with @vanlummelhuizen  that after the PhonologicalVariants have been implemented, it would be good to switch to select2, as it is used by various other colleagues already so it is well known within our team.

> the colors of the pull-downs has not been carried over (the editables have a hack in the code to achieve this) (this could be added to the lookaheads, and to the choices in the form initialisations when in list mode to carry over to the template)

Perhaps leave this for now and do it after we switched to select2?

signbank/dictionary/models.py
susanodd: The booleans are not necessarily the same in the user interface.

I need to see what happens with the variants. These functions have been modified numerous times, also with the addition of the "Cancel". This requires the caller to use the field itself. This is different in the form, since the self is not the object but the form. I made the forms init methods be parameterised on the gloss or specific object so the fields could be instantiated. Then I started adding extra information to the form fields for use in cancel. Now I've added (again) the abstract class for Phonology and all the methods are there now instead of in Gloss. I tried to do this with minimal modifications in this branch so it could still merge into the variations branch. So basically I made additional methods with new names.

It worked to merge by hand (!) and the same migration as the original variations branch is right as rain.

I prefer to leave these as is for the purposes of debugging. Sometimes it gives errors that there is a lazy method that does not have an argument. Some of these similar methods return a translation human-readable value, used in the read-only templates. (There are lots of these. This was to reduce the amount of conditionals on values in the templates.)

The forms can probably use the gloss field (parameter) but in the methods when called in the template not. (The other duplicated methods are display methods in human readable form.)

susanodd:
@Woseseltops the patterns were not visible for a long time. They are now because I kept modifying the element ids to get them that way. I had to develop it with trial and error and had to modify lots of the ids to keep them unique.

I like the ideas for using the patterns now!

I will see about including the fields in the buttons. The problem is in the "reset" code which also depends on the types of the fields in how to reset it. The formatting on the values in the fields are hidden using css patterns on the element ids, based on the types of the fields. I'm not sure moving this into the buttons will break other stuff. I guess just duplicating the data there. Except it's not fetching the information in the "input" fields from the buttons.

Okay, I see what you did. Just the mechanics of pressing the button is moved into the button.

Did you browse the css that changes the format based on the enabled and disabled? The javascript / query is doing that, but the css has entirely different format (it removes all the formatting basically to make it look like plain text. But the patterns are repeated for each panel with the ids and classes there.)

I understand modifying the css dynamically using jquery, except the actual css formatting still has to be declared in css. There is a lot of css for this also.

susanodd:
I didn't investigate this enough to "assimilate" the JSON response type update methods into the page. They just reload the page.  The original the editable code inserts the results back into the page so it does not reload. But now there is kind of a lot going on in the "one column" setup that things would be reinserted after save method. (I will work on that later after the variants have been done. This was not supposed to be such a huge task, but I had to rewrite most of the code to get rid of the single "edit mode" for editables. Yes, this was a lot of work.)

media/js/lookaheads.js
susanodd: Okay, this is an improvement. For several weeks there was no visible pattern in these at all, because it was all derived from the original gloss detail  template and the buttons were added one by one in conjunction with debugging to make sure it still worked the same.

The buttons are also tied to many CSS classes to get this to happen correctly for using one column.

The original editable code had values hidden in each of the table cell elements where each value had to be reinserted in the cell when things were selected, as well as when the save returned. All of that was hidden because the editable forms per cell (!) were only visible when clicked upon. All of this code was obscuring what was going on with showing and hiding things.

susanodd:
URGENT it looks like I forgot to commit the bloodhound javascript file. It's on my computer at the uni but not on github...

Fixed using fetch from cdnjs (now that the updated version can be found there)

(Not on the server yet)

media/js/lookaheads.js
susanodd: Some of the extra parameters are already available in the url and keys, selectors since it's available in the template in the iteration. So it could be further simplified.

susanodd:
Okay, all of the requested revisions have been completed, with the exception of the "model method for Booleans". That has been delayed until the Variations have been added so as to keep the code mergeable since that revises the models.

(There remain a couple of very similar looking "edit toggles" but only a few lines of code. As I pointed out above, there is a lot of CSS tied to these events.)

vanlummelhuizen:
@susanodd 

Some UI/UX issues:

1. The lookahead menus are showing the background in between options. 
2. Some menus are cut off if they extend below the end of the collapsible panel. 
 
Below is a suggestion how to fix that:

<details>
<summary>suggestion for fixes</summary>

```diff
diff --git a/media/css/style-twilight.css b/media/css/style-twilight.css
index bd4f5583..80b0bf89 100644
--- a/media/css/style-twilight.css
+++ b/media/css/style-twilight.css
@@ -495,6 +495,7 @@ table {
   line-height: 20px;
   background-color: var(--typeahead-background);
   color: var(--typeahead-color);
+  margin: 0;
 }
 
 .tt-suggestion.tt-cursor {
@@ -507,6 +508,13 @@ table {
   margin: 0;
 }
 
+.tt-menu {
+    border: 1px solid black;
+    z-index: 10000 !important;
+    max-height: 25em;
+    overflow: auto;
+}
+
 .col-md-4
 {
        width: 100%


```

</details>

For issue 2, it would be best if the menus are above/outside the collapsible panel (while in the same position) so that the whole menu is shown even when the field for that menu is at the bottom of the panel.

susanodd:
@vanlummelhuizen I will try your fix.

But I have been trying to get them in front for a while. None of the solutions from online have worked.

I suspect it could be something to do with the outermost elements structure / nesting. For some there has to be a surrounding element so it can trace back. Most of the modals are embedded inside of table cells etc. I think the browser can't determine the nesting.

I will try the 10000, I didn't try that high yet.

We may have added something to the typeahead "min" file at one point to get the menus in front.

susanodd:
@vanlummelhuizen that added an outline to the menus. But the z-index did nothing. Something (unknown) is overriding it to be 100. Class tt-menu has z-index 100. Always. Regardless of where css is added to enforce this to be larger.

These are being ignored because of bootstrap3, which is found by the browser. That's where it's getting the 100 from.

Here, look how old the file are:

bootstrap-css-files

We seem to be upgrading django-bootstrap3 but not the actual bootstrap files

<details>
I didn't find "tt-menu" anyplace except where you just added it to the css.

But this line is inside of the (downloaded typeahead -- now replaced with a cdnjs):

`media/js/typeahead-bundle-0-11-1.js:            menu: "tt-menu",`

I tried a fix suggested to set the panel-group overflow to visible. But this does not work because in the debug for the browser this is being overridden to "hidden" by "bootstrap.min.css" which is nowhere to be found in our code. It was suggested that when things are hidden, the browser cannot tell what is behind what else.

I previously did a "clear fix". That works except it ends up adding a huge amount of visible space between the panel with the lookahead and the next panel. Some some "stuff" is ending up hidden because "somebody" makes the overflow "hidden" and it's Django Bootstrap apparently.

This is located under (static) bootstrap/css/bootstrap.css/bootstrap.min.css

It's dated 2016

I know you don't want me to hack this file. But we're kind of stuck using Bootsrap 3.
The "answer" I could find said to find out where this code is being "linked" and then make sure the overrides occur after that.
I don't see that we are doing this ourselves, so it must `be {% load bootstrap3 %}`
Except all of our overrides are after this.
Maybe because it's in `baselayout.html` as well as "all" the templates?

Nothing I have tried will override that stupid setting. It keeps getting this from the bootstrap3.min.css file.
For some reason out django bootstrap keeps getting this static file and not fetching it.

What to do, I cannot solve this problem. That's why the code (also the original code) has "vertical space" added to the template.

susanodd:
The issue #1683 is using this branch.

This is becoming a bit of a problem because that branch needs to add extra columns to phonology. Changes to this branch still need to be pulled into that one.

If this branch isn't getting approved, the development needs to continue in the newer branch.
Making it become an enormous pull request.

susanodd:
I need some help with a glitch, @vanlummelhuizen 

You can see the glitch on `signbank-susan`

If you set it to "use lookaheads"  "lookaheads" and then edit.

The field "Relatieve oriëntatie: beweging" (right below Boolean field "Alternerende beweging") does not show the choices.

However, if you use the Ajax call they are there. For some reason they are not showing up in the lookahead. They're not filled in.

Can you please check this? Why are they missing here?

The only thing I can think of is special characters in the name fields. But I would expect a syntax error in the page.

susanodd:
@vanlummelhuizen I attempted many things to determine what is going on.

I commented out the two "repeat" and "altern" fields so that no code is between the lookaheads.
That had no effect.

I suspect there is a hidden white space or control character  in one of the names of RelOriMov

It only happens with this lookahead, none of the others. Nothing is shown. But the ajax call on its own shows the records.

Please help!

UPDATE

I tried "not including" choices with "+" and ">" in the names.

Then it works!

It does not work so far to replace this character with anything. The ajax still works, but nothing shows in the menu.

Some other choice lists include ">" but don't include "+".

Removing the choices and the typeahead works correctly, so it's something specific with the data.

UPDATE

The closest thing that seems that the bloodhound tries to infer the "type" of whatever the data is.
Stack exchange suggests to use a "prepare" function in the "remote" source part (where the url to our ajax is)
Maybe because "+" and ">" appear it thinks it's some numerical data that doesn't parse?

vanlummelhuizen:
> The field "Relatieve oriëntatie: beweging" (right below Boolean field "Alternerende beweging") does not show the choices.

@susanodd I have looked into this, but I have not found the problem yet.

susanodd:
> > The field "Relatieve oriëntatie: beweging" (right below Boolean field "Alternerende beweging") does not show the choices.
> 
> @susanodd I have looked into this, but I have not found the problem yet.

Do you mean that you have not found a solution, or that it works okay for you?

(This is for the lookaheads, the normal choice list works okay.)

susanodd:
It might be that inside the bloodhound, a "config" is used now, as per @Woseseltops to condense the amount of code.

Reading about the backquote, it says that javascript evaluates expressions inside of it. Perhaps the ">" and "+" in the name field is causing it to be evaluated? (But somewhere inside of Bloodhound and not causing a visible error, just becoming empty.)

Okay, it's called a backtick.

> Important Considerations
> 
> - Backticks are part of ES6 (ECMAScript 2015) and may not work in older environments like Internet Explorer.
> - Use backslashes (```) to escape backticks inside template literals.

So something is going on with expressions inside of it.

So the ">" and "+" may need to be "backtick" escaped ?

susanodd:
I solved it, via a roundabout code adjustment. I'm not actually sure which part has fixed it. I kept tweaking and seeing what the bloodhound was seeing. At first it looked like it was stopping at particular machine values.

https://github.com/Signbank/Global-signbank/pull/1752/commits/5736520bb3d63bbfd8bb814054d25bec44fb9756

Since you guys were slow in approving this branch, I continued to work on this code in the followup branch #1752 

Now you'll have 11K lines of code to review.

susanodd:
> Thanks @susanodd , this must have been a lot of work! My comments are indeed all related to being DRY, as you already predicted.
> 
> I agree with @vanlummelhuizen that after the PhonologicalVariants have been implemented, it would be good to switch to select2, as it is used by various other colleagues already so it is well known within our team.
> 
> > the colors of the pull-downs has not been carried over (the editables have a hack in the code to achieve this) (this could be added to the lookaheads, and to the choices in the form initialisations when in list mode to carry over to the template)
> 
> Perhaps leave this for now and do it after we switched to select2?

Yes, I am the one who implemented the select2 stuff in Signbank. It's used in all the Search plates where you can multi-select the choices. I am one of the "various colleagues". :)

I continued implementation of this on the new combined branch #1752 

I solved the problem with the  RelOriMov typeahead on the new branch.

PR #1734, opened 2026-04-20
BUG FIX, allow to set dataset default language and acronym

### LINKED ISSUE #1733: How do you add new datasets?
In the admin, the acronym and the default language are read only... I'm guessing via script or fixture?

susanodd:
if appropriate permissions

### DIFF
--- a/signbank/dictionary/admin.py
+++ b/signbank/dictionary/admin.py
@@ -56,6 +56,16 @@ class DatasetAdmin(GuardedModelAdmin):
     def has_delete_permission(self, request, obj=None):
         return False
 
+    def get_readonly_fields(self, request, obj=None):
+        if obj is not None and request.user not in obj.owners.all() and not request.user.is_superuser:
+            return self.readonly_fields
+        return []
+
+    def has_change_permission(self, request, obj=None):
+        if obj is not None and request.user not in obj.owners.all() and not request.user.is_superuser:
+            return False
+        return True
+
 
 class KeywordAdmin(VersionAdmin):
     search_fields = ['^text']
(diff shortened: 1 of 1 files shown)


susanodd:
@Woseseltops @vanlummelhuizen 

I see CoPilot's suggestions for the various permissions.

What is not clear to me is what users are able to use Admin. As far as I can tell from experimentation, only superusers and staff can get into Admin.

But in many cases, if Django does not find a form and a form exists in Admin, it ends up using that form. (Observed on basis of code debugging and print statements in other isses.)

It is not clear to me what specific permissions have been granted to Dataset Owners. Whether this circumvents staff settings.

That's why the original code was checked in the order it was, because I tried to make assumptions about what users would be able to create datasets using the admin. And hence need access to the two fields that were previously read-only.

It was also intended that once a dataset already exists, it would be wise to keep those fields read-only, since modifying them entails moving all of the video files to a different folder setup. (Default language and Acronym are inherent to the video folders and filename of the video files.)

The fields were previously made read-only to prevent extreme side effects to the video storage.

I'm not sure whether the suggestions of CoPilot also works as intended.

Some tests are needed in order to create various test users with certain permissions.

I assumed that a dataset owner would be able to use the admin. But this is not certain.

Woseseltops:
Hey @susanodd , if you cannot verify what Copilot says, ignore it. I think it might not have the context that datasets can only be added in the admin anyway... but if we're wrong and Copilot's right, I don't see it as a huge security thing, as I'm not that worried current datasets owners will go wild adding new ones :P

susanodd:
I can go ahead and make some tests. That seems wise. When using it, it worked as I wanted it to work.

Woseseltops (approved):

PR #1735, opened 2026-04-20
Setup phonolocal variation model and admin support

### LINKED ISSUE #1726: PhonologicalVariation: model and display in UI
Part 1: model is just a copy of all phonogical fields in the Gloss
Part 2: display as a table in the detail view

susanodd:
TO DO
- hide/show the editing functionality in Phonological Variations template (only a user with Dataset Change permission has editing permission)
- for inclusion of the new template code inside of Gloss Detail, the editable code with all the individual field OK/Cancel buttons needs to be replaced with lookahead code and panel-specific buttons (#1724). This is to reduce the width of the columns so the new variations will fit. Because of the setup of editable, all of the fields in the entire template are linked together in edit mode. This needs to be totally rewritten to confine the scope of the element nesting and events.

This is a working implementation.

Implemented
- New model `Phonology` that is an abstract class, inherited by `Gloss` and `PhonologicalVariation`
- Table showing variations in panel of Gloss Detail View
- Edit button to switch to new Phonological Variations View
- Variations can be edited and created in the new view
- Admin to Edit already created variations
- Admin create is not allowed because the gloss is not known (!) (constraints on gloss lemmas, annotations, and senses prohibit creating and editing them in the admin)

Functionality to update existing variations in the table

- Ajax calls are added to the table cells by means of a form-per-variant setup with a save button per form. These make use of dynamic data and jquery to perform the appropriate update using the fields for the form associated with the desired variant.
- The display uses columns per variant and rows per field. The primary gloss phonology is in the first column.

The tabular editing of variants and creation of a new variant is done in this template, not in the (extremely crowded) Gloss Edit View. This is necessary because identifiers are needed to distinguish all of the different variants and their forms in the template and their associated ajax calls.


susanodd:
@Woseseltops based on #1740 what would the default variation number be? 1 instead? I'll need to generate the migration again.

Woseseltops:
Hey @susanodd , perhaps better to reserve pull requests to talk about code and reviews. To answer the question: the first variation number would be 2 (as 1 is the main gloss).

About this PR: I wrote down to review this today, but given our discussion about removing editable, shall I wait until these extra things are realized?

susanodd:
I suggest to implement the Variations template so the primary gloss can also be edited there. (It's just an additional form for that column.)

And then just use the template to replace Gloss Detail Phonology. (Not mix editable together. Really, it is impossible to follow the control flow because editable adds forms and input fields dynamically and they are only visible when you click on a cell. Those dynamic forms conflict with the "other" forms. There is even code for hiding other forms because at one time editable was waiting on a form to complete but would open another form at the same time. That code to make sure the form "disappears" is still in the gloss edit javascript.)

Perhaps a setting USE_PHONOLOGICAL_VARIANTS so the users could keep using the editable version if they don't care about having variants.

susanodd:
There is a new branch for this that is starting from branch #1724 and including this work there. The merging would otherwise be complicated because of the new models. This branch also has some trial and error commits before the abstract class for Phonology was introduced.

PR #1746, opened 2026-06-03
Fix two HTTP 500s on /dictionary/api_* endpoints

rem0g:
Two separate HTTP-500 bugs encountered while integrating the Signbank API into [signCollect-v2](https://github.com/rem0g/signCollect-v2).

### 1. `retrieve_language_code_from_header` → `TypeError` when `Accept-Language` omitted
`parse_accept_lang_header(http_accept_language_header)` (Django 4.2.29) raises `TypeError: object of type 'NoneType' has no len()` when the header is missing. Fix: coerce `None` to `''`, which Django handles cleanly.

### 2. `get_gloss_update_human_readable_value_dict` → `JSONDecodeError` on empty body
`json.loads(request.body.decode('utf-8'))` raises before the view's own validation can produce a meaningful 400. Fix: treat empty / invalid bodies as `{}` so the view's existing "Gloss operation not confirmed" 400 fires.

### Affected endpoints
- `api_update_gloss`, `api_create_gloss`, `api_delete_gloss` (and their `/nl/` variants)

### DIFF
--- a/signbank/abstract_machine.py
+++ b/signbank/abstract_machine.py
@@ -45,7 +45,7 @@ def retrieve_language_code_from_header(url_language_code, api_accept_language_he
         # the url language code matches the one in the Accept-Language request
         return url_language_code
     # parse the language codes to get the first one
-    parsed_language_codes = parse_accept_lang_header(http_accept_language_header)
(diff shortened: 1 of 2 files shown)


Woseseltops:
Rewrote comment to be shorter and about current state (not about the past)

PR #1747, opened 2026-06-03
Fix NOT NULL constraint failure on GlossRevision when previous value is NULL

rem0g:
`GlossRevision.old_value` / `new_value` are `NOT NULL CharField`s. When a gloss field was previously `NULL` and gets a new value, `original_human_value` could be `None`, which violated the constraint and rolled back the entire `api_update_gloss` transaction with a confusing "Transaction management error."

Fix: coerce both ends to `""` and normalise `field_name` to `str` so the constraint is satisfied.


susanodd:
Good catch! (Apologies that it was broken. It looks like coercions were removed at some point.)

Woseseltops:
Hmm seems like @susanodd is reviewing this at the same time ;). Agreed this is a good catch.

PR #1748, opened 2026-06-03
Detect WebM uploads from plain "WebM" file(1) output

rem0g:
`file(1)` on Ubuntu 22.04 returns just `"WebM"` for some `.webm` uploads (no `"video/webm"` prefix). The existing match missed those, causing the upload to be rejected as an unknown format. Fix: also accept the plain `"WebM"` form.


Woseseltops (approved):
Okay from my point of view, but I want an okay from @vanlummelhuizen as well because I remember him talking about this mechanic.

vanlummelhuizen:
I made the same change recently. Perhaps we get a merge conflict, but that should not be a problem. It is merged.

Issue #1520, opened 2025-03-20
Available directory listing for /static


Woseseltops:
See https://signbank.cls.ru.nl/static/ . Probably needs a different Apache config. Needs to be changed in the Ansible playbooks as well.

This came out of a vulnerability scan, reported to us by C&CZ

susanodd:
@Woseseltops I am working on this on signbank-susan.

But after modifying some things in the ansible script for apache, it now is giving 503 service unavailable.

`The server is temporarily unable to service your request due to maintenance downtime or capacity problems. Please try again later.`

Is this something I have caused, or have they removed this container's url? The virtual environment and all the tests pass, so the container is still operational.

susanodd:
@Woseseltops, in spite of a new empty container, and sucessfully installing ALL of the ansible scripts, the server is doing exactly what it was before. Server Error 500.

susanodd:
@Woseseltops I fixed the container. The problem was that the ansible script created the static folders, but they were root as owner and group. I changed them to be ubuntu and wwwsignbank.

susanodd:
SOLVED !!!

susanodd:
@Woseseltops I added these to the apache.yaml script on `signbank-susan`

I did:
```
root@signbank-susan:~/signbank-ansible# git add apache.yaml
root@signbank-susan:~/signbank-ansible# git commit -m "Apache enable CORS headers, disable directory listing"
```

This yields:
```
Author identity unknown

*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.

fatal: unable to auto-detect email address (got 'root@signbank-susan.(none)')
```

But I don't know how to commit/push it to the humanities github where ansible is.

susanodd:
I figured it out! It's not the normal git username, it's the ponyland username.

Woseseltops:
If the latest version of the Ansible is also pushed to https://gitlab.science.ru.nl/tsg-humlab/signbank-ansible/-/commit/18c9db9c55aabbb56bca2cfc888df660e0c5ee0b, feel free to close this issue!

susanodd:
Yes, it's been pushed. It was also pulled to the other servers and run there.

PR #1749, opened 2026-06-03
Let server_specific.py override STATIC_ROOT

rem0g:
`base.py` unconditionally set `STATIC_ROOT = PREFIX_URL` (a URL fragment like `/signbank_local`), so `collectstatic` would try to `mkdir` an absolute path under `/` and fail with `PermissionError` on any non-root install. Per-instance settings already override `URL` / `PREFIX_URL` etc; let them override `STATIC_ROOT` too with a real disk path.

No behaviour change unless `server_specific.py` defines `STATIC_ROOT`.


susanodd:
See #1520 for some previous problems with static on the servers. (Ter informatie)

PR #1750, opened 2026-06-03
API upload receipts, idempotency headers, JSON-500 middleware + role-based verbose-errors gate

rem0g:
Implements §1, §2, §5 from `docs/signbank-api-recommendations.md` plus the role-based error-detail gating discussed in @Woseseltops's reply on #1743.

### New
- **`signbank/api_receipts.py`** — `sha256_of_uploaded_file`, `file_summary`, RFC-9110 `If-Match` parser (`_parse_if_match`), `evaluate_idempotency_headers`, `upload_receipt`. Handles `"a", "b"`-style comma-separated quoted ETag lists and `W/`-prefixed weak validators per the review.
- **`signbank/api_middleware.py`** — `JsonErrorMiddleware`: catches uncaught exceptions on API endpoints and emits a JSON 500. Targets API endpoints by introspecting the resolved view's `__module__` (`signbank.api_*`, `gloss_update`, `zip_interface`) with `Accept: application/json` as a fallback — no more hard-coded URL prefixes.
- **`signbank/api_errors.py`** — `has_verbose_errors(user)` + `safe_error_message(exc, user)`. Raw Python exception text is only included in API responses for `settings.DEBUG`, superusers, or members of the `Verbose API Errors` group; everyone else gets `"Internal server error."`

### Modified
- **`signbank/api_interface.py`** — `api_add_video` / `api_add_image` compute the incoming sha256 before save, evaluate `If-Match` / `If-None-Match` against the current stored file, and return an `UploadReceipt` (sha256 + size of incoming and stored, `replaced` block for prior file).
- **`signbank/gloss_update.py`** — exception text in the `api_update_gloss` 400 response now goes through `safe_error_message()` instead of being surfaced verbatim.
- **`signbank/settings/base.py`** — `JsonErrorMiddleware` registered.
- **`docs/signbank-api.json`** — `UploadReceipt`, `FileSummary`, `ApiError` schemas; `If-Match` / `If-None-Match` parameters; 412 / 500 response shapes on `/dictionary/api_update_gloss/{glossid}/video` and `.../image`.

### Deployment note
The role-based gate is opt-in: enable verbose errors for an API user by adding them to a `Verbose API Errors` group (`Group.objects.get_or_create(name='Verbose API Errors')`). Until that group exists, only `DEBUG=True` and superusers get raw exception text.

### Replaces
#1743 — this is the substantive piece of the 5-way split requested in the review. The other four are #1746, #1747, #1748, #1749.

Cc @vanlummelhuizen @Woseseltops @susanodd — the top-level reply on #1743 has the per-thread responses; happy to also paste them inline here if preferred.

### DIFF
--- a/signbank/api_errors.py
+++ b/signbank/api_errors.py
@@ -1,0 +1,44 @@
+"""
+Helpers for shaping error responses returned by the API.
+
+Not all clients holding a valid API token can be trusted with raw Python
+exception text — internal paths, query fragments and other sensitive data
+may leak through ``str(exc)``. ``has_verbose_errors`` decides who gets the
+detail; ``safe_error_message`` returns the appropriate string for the
+caller. See PR #1743 (Wessel/Micha review thread) for the discussion.
+"""
+
+from django.conf import settings
+
+
+VERBOSE_ERRORS_GROUP = 'Verbose API Errors'
+
+GENERIC_ERROR_MESSAGE = 'Internal server error.'
+
+
+def has_verbose_errors(user):
+    """
+    True if ``user`` is allowed to see raw exception detail in API responses.
+
+    Superusers always qualify, and so do members of the
+    ``Verbose API Errors`` group (created by an admin per deployment).
+    ``settings.DEBUG`` also enables verbose errors globally — useful for
+    development and signbank-test, never for production.
+    """
+    if getattr(settings, 'DEBUG', False):
+        return True
+    if user is None or not getattr(user, 'is_authenticated', False):
+        return False
+    if getattr(user, 'is_superuser', False):
+        return True
+    return user.groups.filter(name=VERBOSE_ERRORS_GROUP).exists()
+
+
+def safe_error_message(exc, user, *, generic=GENERIC_ERROR_MESSAGE):
+    """
+    Return either the formatted exception or a generic message, depending
+    on whether ``user`` is allowed to see verbose errors.
+    """
+    if has_verbose_errors(user):
+        return f"{type(exc).__name__}: {exc}"
+    return generic
--- a/signbank/api_middleware.py
+++ b/signbank/api_middleware.py
@@ -1,0 +1,70 @@
+"""
+Turns uncaught exceptions on API endpoints into a JSON 500 instead of the
+default HTML error page. Browser hits the same URL still get HTML; JSON /
+API clients get something they can parse.
+"""
+
+import logging
+
+from django.http import JsonResponse
+
+from signbank.api_errors import has_verbose_errors, GENERIC_ERROR_MESSAGE
+
+
+log = logging.getLogger(__name__)
+
+
+API_VIEW_MODULES = (
+    'signbank.api_',
+    'signbank.gloss_update',
+    'signbank.zip_interface',
+)
+
+
+def _is_api_view(request):
+    """
+    Decide whether ``request`` is for an API endpoint without hard-coding
+    URL prefixes. Order of preference:
+
+      1. The view that resolved the URL lives in one of the API modules
+         listed above (introspected via ``request.resolver_match.func``).
+      2. The client explicitly asked for JSON (and not HTML) via the
+         ``Accept`` header — covers cases where URL resolution has not run
+         yet or the view lives outside ``signbank.api_*``.
+    """
+    rm = getattr(request, 'resolver_match', None)
+    view = getattr(rm, 'func', None) if rm else None
+    module = getattr(view, '__module__', '') or ''
+    if any(module.startswith(prefix) for prefix in API_VIEW_MODULES):
+        return True
+    accept = (request.META.get('HTTP_ACCEPT') or '').lower()
+    return 'json' in accept and 'html' not in accept
+
+
+class JsonErrorMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        return self.get_response(request)
+
+    def process_exception(self, request, exception):
+        if not _is_api_view(request):
+            return None
+        log.exception(
+            "api 500 on %s %s: %s", request.method, request.path, exception
+        )
+        verbose = has_verbose_errors(getattr(request, 'user', None))
+        body = {
+            'ok': False,
+            'status_code': 500,
+            'errors': [{
+                'field': None,
+                'code': 'internal_error',
+                'message': f"{type(exception).__name__}: {exception}" if verbose else GENERIC_ERROR_MESSAGE,
+            }],
+        }
+        if verbose:
+            import traceback
+            body['traceback'] = traceback.format_exc().splitlines()[-30:]
+        return JsonResponse(body, status=500)
--- a/signbank/api_receipts.py
+++ b/signbank/api_receipts.py
@@ -1,0 +1,222 @@
+"""
+Helpers for emitting upload receipts on the api_add_video / api_add_image
+endpoints:
+
+* every upload returns metadata the client can use to prove later *which*
+  file landed (sha256, size, uploaded_at, uploaded_by, stored path);
+* if the upload replaced an existing file, the prior file's metadata is
+  reported under ``replaced``;
+* clients can opt into refusal-on-overwrite by sending ``If-None-Match: *``
+  or refusal-on-mismatch via ``If-Match: "<sha256>"``.
+
+See ``docs/signbank-api-recommendations.md`` (#1 and #2) for the design
+rationale.
+"""
+
+import hashlib
+import os
+from datetime import datetime, timezone
+
+from django.http import JsonResponse
+
+from signbank.tools import get_two_letter_dir
+from signbank.settings.server_specific import WRITABLE_FOLDER, GLOSS_IMAGE_DIRECTORY
+
+
+ISO_8601_UTC = '%Y-%m-%dT%H:%M:%SZ'
+
+
+def _iso_utc(dt):
+    """Format a datetime as RFC 3339 / ISO 8601, UTC, second precision."""
+    return dt.strftime(ISO_8601_UTC)
+
+
+def now_iso():
+    return _iso_utc(datetime.now(tz=timezone.utc))
+
+
+def sha256_of_uploaded_file(uploaded_file):
+    """
+    Compute the SHA-256 of a Django UploadedFile *without* losing the read
+    cursor. Returns the hex digest and the file size in bytes.
+
+    Uses chunked iteration so we don't load large videos into memory.
+    """
+    h = hashlib.sha256()
+    size = 0
+    for chunk in uploaded_file.chunks():
+        size += len(chunk)
+        h.update(chunk)
+    try:
+        uploaded_file.seek(0)
+    except (AttributeError, OSError):
+        pass
+    return h.hexdigest(), size
+
+
+def sha256_of_path(path):
+    """Compute SHA-256 + size for an existing file on disk, or (None, None) if absent."""
+    if not path or not os.path.isfile(path):
+        return None, None
+    h = hashlib.sha256()
+    size = 0
+    with open(path, 'rb') as file:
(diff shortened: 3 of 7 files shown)


rem0g:
Calling out the role-based error gating for visibility — this implements @Woseseltops's [second proposal from #1743](https://github.com/Signbank/Global-signbank/pull/1743#issuecomment-) directly. Lives in `signbank/api_errors.py`:

```python
VERBOSE_ERRORS_GROUP = 'Verbose API Errors'
GENERIC_ERROR_MESSAGE = 'Internal server error.'

def has_verbose_errors(user):
    if getattr(settings, 'DEBUG', False):
        return True
    if user is None or not getattr(user, 'is_authenticated', False):
        return False
    if getattr(user, 'is_superuser', False):
        return True
    return user.groups.filter(name=VERBOSE_ERRORS_GROUP).exists()

def safe_error_message(exc, user, *, generic=GENERIC_ERROR_MESSAGE):
    if has_verbose_errors(user):
        return f"{type(exc).__name__}: {exc}"
    return generic
```

A caller gets the raw `TypeName: message` if **any** of these is true:

1. They belong to the `Verbose API Errors` Django group (the "special role" — created per deployment by an admin via `Group.objects.get_or_create(name='Verbose API Errors')` or just through the Django admin UI).
2. They are a superuser.
3. `settings.DEBUG` is on (dev / signbank-test convenience — never production).

Everyone else gets the literal string `"Internal server error."` — no exception text, no traceback. The middleware also only attaches the `traceback` field for verbose users.

Wired into two places:

- **`signbank/api_middleware.py`** — `JsonErrorMiddleware.process_exception` (any uncaught API 500).
- **`signbank/gloss_update.py`** — the `(TransactionManagementError, ValueError, IntegrityError)` 400 in `api_update_gloss` (previously surfaced `str(exc)` verbatim, which was the line @vanlummelhuizen flagged on #1743).

`docs/signbank-api.json` documents the contract under `components/responses/InternalError` (description: "The detailed Python exception text is only included for users with verbose errors enabled.").

@vanlummelhuizen @Woseseltops — happy to switch to any of these shapes if you prefer:

- A custom **permission** (e.g. `dictionary.view_verbose_api_errors`) instead of a group.
- A boolean on **`UserProfile`** if you want it editable per-user without group membership.
- A flag on the **API token** itself (so the same human can have a "verbose" token for debugging and a "normal" token for production scripts).

The current group choice was the lightest implementation that matched the "special role" wording — happy to refactor based on what fits the existing Signbank permission model best.

rem0g:
End-to-end smoke run against a local install of this branch — sharing the results so the implementation is grounded in actual behaviour rather than just code review.

### 1. `JsonErrorMiddleware` + role gate (`Django RequestFactory`, real `User` rows, real `Group`)

Forced a `ValueError("sample exception with /secret/path/leak.db in it")` through `JsonErrorMiddleware.process_exception` for four personas under both `DEBUG` modes:

| Caller | `DEBUG=True` (dev) | `DEBUG=False` (production) |
|---|---|---|
| anonymous | JSON 500, verbose + traceback | JSON 500, **generic, no traceback** |
| plain user | JSON 500, verbose + traceback | JSON 500, **generic, no traceback** |
| member of `Verbose API Errors` group | JSON 500, verbose + traceback | JSON 500, verbose + traceback |
| superuser | JSON 500, verbose + traceback | JSON 500, verbose + traceback |

Every response was `application/json` with the documented `{ok: false, status_code: 500, errors: [...]}` shape — no HTML 500 page. This matches @Woseseltops's "only provide the detailed error message if the status is high enough" exactly.

### 2. Real upload via `runserver` + `curl` (gloss `UPLOADTEST` in NGT, 2.2 KB mp4)

**2a — plain upload → HTTP 200**
```json
{
  "ok": true,
  "gloss_id": 1,
  "sha256": "6f2d87c0aa19719e18f2fc038be48155aea47d23b759311215be8f6574e0950a",
  "size_bytes": 2240,
  "stored_path": "/.../glossvideo/NGT/UP/UPLOADTEST-1.mp4",
  "stored_sha256": "6f2d87c0…",
  "stored_size_bytes": 2240,
  "replaced": {
    "filename":  "UPLOADTEST-1.mp4",
    "sha256":    "40bcc8fb…",
    "size_bytes": 1715,
    "uploaded_at":"2026-05-08T19:52:50Z"
  },
  "label": "center"
}
```

**2b — re-upload with `If-None-Match: *` → HTTP 412**
```json
{
  "ok": false, "status_code": 412,
  "errors": [{"code":"precondition_failed",
              "message":"A file already exists for this gloss; refused because of If-None-Match: *."}],
  "current": { "sha256": "6f2d87c0…", "size_bytes": 2240, … }
}
```

**2c — `If-Match: "deadbeef…"` (wrong sha) → HTTP 412**
> *"current sha256 is '6f2d87c0…' (expected one of ['deadbeef…'])."*

**2d — `If-Match: "deadbeef", "6f2d87c0…"` (comma-separated list, second matches) → HTTP 200**

That last case is the exact RFC-9110 scenario @vanlummelhuizen called out — `"foo", "bar"` style. The new `_parse_if_match` (api_receipts.py:99) splits on commas, strips quotes per ETag, handles weak validators (`W/`), and `evaluate_idempotency_headers` accepts the upload if any listed ETag matches the current sha256.

### Reproducing locally

If you want to run the same matrix on signbank-dev / signbank-test:

```python
# Django shell
from django.test import RequestFactory
from django.test.utils import override_settings
from signbank.api_middleware import JsonErrorMiddleware
mw  = JsonErrorMiddleware(get_response=lambda r: None)
req = RequestFactory().post('/dictionary/api_update_gloss/1/video', HTTP_ACCEPT='application/json')
req.user = your_user
with override_settings(DEBUG=False):
    resp = mw.process_exception(req, ValueError("test"))
print(resp.status_code, resp.content)
```

And for real uploads, any small mp4 + a `bin/develop.py runserver` is enough — auth via `Authorization: Bearer <token>` from a `SignbankAPIToken`.

Happy to add Django `TestCase`s covering these scenarios if you'd like them in the suite proper rather than just as smoke evidence. The new helpers (`_parse_if_match`, `has_verbose_errors`, `safe_error_message`) are pure-Python and don't need a DB, so unit tests for them are cheap.

signbank/api_middleware.py
```diff
@@ -0,0 +1,70 @@
...
+    Decide whether ``request`` is for an API endpoint without hard-coding
+    URL prefixes. Order of preference:
+
+      1. The view that resolved the URL lives in one of the API modules
+         listed above (introspected via ``request.resolver_match.func``).
+      2. The client explicitly asked for JSON (and not HTML) via the
+         ``Accept`` header — covers cases where URL resolution has not run
+         yet or the view lives outside ``signbank.api_*``.
+    """
+    rm = getattr(request, 'resolver_match', None)
```
vanlummelhuizen, line 35: Change `rm` to `resolver_match` for readability.

signbank/api_middleware.py
```diff
@@ -0,0 +1,70 @@
...
+from signbank.api_errors import has_verbose_errors, GENERIC_ERROR_MESSAGE
+
+
+log = logging.getLogger(__name__)
+
+
+API_VIEW_MODULES = (
+    'signbank.api_',
+    'signbank.gloss_update',
+    'signbank.zip_interface',
```
vanlummelhuizen, line 20: I don't think this is a view module. There is no urls path that points to any class or function.

signbank/api_middleware.py
```diff
@@ -0,0 +1,70 @@
...
+from django.http import JsonResponse
+
+from signbank.api_errors import has_verbose_errors, GENERIC_ERROR_MESSAGE
+
+
+log = logging.getLogger(__name__)
+
+
+API_VIEW_MODULES = (
+    'signbank.api_',
```
vanlummelhuizen, line 18: I think only `signbank.api_interface` contains actual views.

signbank/api_middleware.py
```diff
@@ -0,0 +1,70 @@
...
+    'signbank.gloss_update',
+    'signbank.zip_interface',
+)
```
vanlummelhuizen, line 21: `signbank.abstract_machine` also contains an API view (`api_create_gloss`).

signbank/api_middleware.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def _is_api_view(request):
```
vanlummelhuizen, line 24: Although I trust this setup will work, I think it is not ideal. It is easy to forget API endpoints, partly because they can be everywhere in the code.

What we want is a single source of truth. Would it be possible to use `signbank-api.json` for this? Also a question for you, @Woseseltops .

signbank/api_receipts.py
```diff
@@ -0,0 +1,222 @@
...
+    try:
+        gloss_video = gloss.glossvideo_set.filter(version=0).first()
+    except (AttributeError, ValueError):
+        gloss_video = None
+    if not gloss_video or not gloss_video.videofile:
+        return None
+    try:
+        path = gloss_video.videofile.path
+    except (AttributeError, ValueError):
+        return None
```
vanlummelhuizen, line 106: `Gloss` had method `get_video`: https://github.com/Signbank/Global-signbank/blob/45be4a654a513ea1e06e653ca3b16493f99b9af4/signbank/dictionary/models.py#L2379-L2390 Could that be used here?

signbank/api_receipts.py
```diff
@@ -0,0 +1,222 @@
...
+                                gloss.lemma.dataset.acronym, get_two_letter_dir(gloss.idgloss))
+    except (AttributeError, ValueError):
+        return None
+    if not os.path.isdir(base_dir):
+        return None
+    prefix = f"{gloss.idgloss}-{gloss.pk}."
+    candidates = [name for name in os.listdir(base_dir) if name.startswith(prefix)]
+    if not candidates:
+        return None
+    candidates.sort(key=lambda name: os.path.getmtime(os.path.join(base_dir, name)), reverse=True)
```
vanlummelhuizen, line 123: `Gloss` has method `get_image_path`: https://github.com/Signbank/Global-signbank/blob/45be4a654a513ea1e06e653ca3b16493f99b9af4/signbank/dictionary/models.py#L2343-L2359 Could that be used here?

signbank/api_receipts.py
```diff
@@ -0,0 +1,222 @@
...
+        'uploaded_by': uploaded_by,
+        'sha256':      incoming_sha256,
+        'size_bytes':  incoming_size,
+    }
+    if stored_summary:
+        receipt['stored_filename']   = stored_summary.get('filename')
+        receipt['stored_path']       = stored_summary.get('stored_path')
+        receipt['stored_size_bytes'] = stored_summary.get('size_bytes')
+        receipt['stored_sha256']     = stored_summary.get('sha256')
+    receipt['replaced'] = previous_summary
```
vanlummelhuizen, line 221: This could be in the definition of `receipt` above.

signbank/api_errors.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def safe_error_message(exc, user, *, generic=GENERIC_ERROR_MESSAGE):
+    """
+    Return either the formatted exception or a generic message, depending
+    on whether ``user`` is allowed to see verbose errors.
+    """
+    if has_verbose_errors(user):
+        return f"{type(exc).__name__}: {exc}"
+    return generic
```
vanlummelhuizen, line 44: Please write a test for these functions. It is essential these keep working correctly.

vanlummelhuizen (changes requested):
@rem0g Again, thanks for your work. I have some additional comments.

I have requested tests for the functions in `api_errors.py`. If you have time, tests for other parts you wrote are welcome.

vanlummelhuizen:
@Woseseltops there are some Python files that start with `api_`. I think we should consider moving them to a dedicated `api` folder/Django app. If we do so, we should determine whether the other API endpoints belong there too. What do you think?

Issue #1751, opened 2026-06-16
Phase out Twitter's typeahead and bloodhound Javascript


vanlummelhuizen:
Twitter's typeahead and bloodhound Javascript library is very outdated (https://github.com/twitter/typeahead.js/). We should therefore phase is out in favor of a modern and well maintained alternative. 

I recommend [Select2](https://select2.org/) which is well established in the Django world, e.g. in the [Django admin](https://docs.djangoproject.com/en/6.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.autocomplete_fields). See also https://github.com/codingjoe/django-select2. I have used it extensively.

susanodd:
There are several templates that use typeheads (the static file) that have nothing to do with the gloss edit  issues that led to the revisions in the pull requests:

```
admin_handshape_list.html
admin_dataset_manager.html
dataset_detail.html
```

These have all been successfully updated to use the cdnjs.cloudflare typeahead.

susanodd:
@vanlummelhuizen there is a problem with the user typeahead.

- I have revised the "typeahead" the same way all the other typeaheads have been revised.

- But when I attempt to use "Manage Datasets" the browser wants to "put my own username" into the form, not show the "signbank" user names.

- The browser is overriding the lookahead and thinks I want to "login" to this "username" field.

- It's on Manage Datasets where you can grant permissions. There are "user typeaheads" here.

There is a problem with this.

https://github.com/Signbank/Global-signbank/pull/1764/commits/9e9e4e082fbc2bde059bac247bbd928a7a0a0fc6

- It did not work to change `username` to `gebruiker` in the form.

- It did not work to add `autocomplete="off"` (both suggested by CoPilot)

susanodd:
- The user typeahead is working correctly. I had to revise the occurrences of "username" to "gebruiker" in the code so the browser would stop prompting for a login. The placeholder has also been modified to "Name of User".

- But, it is still broken. It allows to select from the list and shows the selection, but the value is not passed back in the request. I don't understand actually why the dataset forms are sent as get rather than post. Could this be related to old software versions (again).

[...]

PR #1752, opened 2026-06-18
Phonologicalvariations (including lookaheads branch)

### LINKED ISSUE #1726: PhonologicalVariation: model and display in UI

susanodd:
This branch adds the code in branch #1735 (issue #1726) into the branch #1724 (issue #1714)

The original variations branch was merged by hand into the lookahead branch in order to prevent conflicts.

This branch is huge. The branch it builds on was not approved in time and so that code has been further developed here.

- Split Python update methods into separate methods per panel? (Not necessary) (Three of the panels --- general, semantics, publication --- share one method --- update.py:edit_gloss_save --- at setup, only the fields of the panel are passed to the Python method. This is either an example of "too general methods" or "clever combining functionality into one method". Take your pick.


susanodd:
@Woseseltops the appearance of a button to add a variation is very confusing. There are too many buttons for phonology. It can't appear at the right because for all the other panels, that's where the save / dismiss buttons are.
I moved the save  button for phonology to the column when there are variations. But now the "add variant" has no easy place. There can't be an extra column for a header button because the table is getting very wide, it will end up off the screen.

susanodd:
For some strange reason, the relOriMov lookahead appears empty in the template. But the ajax call works.

See https://github.com/Signbank/Global-signbank/pull/1724#issuecomment-4795943926

susanodd:
@Woseseltops it's not just those text fields that cause the panel to become wide, the actual "verbose names" in the first column are ending up wrapped because they are long. (This is on a laptop when you have created 2+ variants)

It doesn't work to "force" the text field to be narrower using a "textarea" input and setting the dimensions.
Somehow Django keeps using the actual width of this field and won't let it be narrower unless you are actually inside of edit mode, then you can "pull" it t make it narrower. But when not in edit mode, it keeps making it 50 characters wide.

That is also causing the "field names" column to shrink and be really narrow on a laptop when you have 3 variations (see screenshot)

The width of the columns always needs to accommodate the widest content (also for the pulldown fields). I have some code that attempts to shorten the width when the "letter" "number" is there, but this does not seem to work.

Below, you can see the "pull" tab for the VirtObj field. The field is remaining 50 characters wide. Django is getting this from the model field. It ignores the resetting in forms and in the template.

Also if you try to be sneaky, because the field gets initialised from the actual data, the "template" input gets "associated with" the actual "data" that Django knows about.

crowded-columns

susanodd:
Caveat: If there are more than two variants (three including the primary gloss), the phonology is TOO WIDE for a laptop computer (13-inch with retina display)

Perhaps during editing, we need to allow to collapse the left column?

susanodd:
The development server is not up to date. Major revisions have been done since the last sync.

susanodd:
Okay, I'm repairing some bugs. The "combined" "single" update method for updating "all" of the variations and the primary gloss phonology at the same time has a few glitches.

I fixed the "hide" on empty fields (taking care of the variations too. It was only looking at the primary gloss, not the variations. So was ignoring the case that the user would have a field set for a variation but not for the primary gloss. (No idea if this happens)

susanodd:
development server `signbank-test` sync'd up to here

vanlummelhuizen:
@susanodd You solved it!

I hate to say it, but I think I found a new problem: only the first Phonological Variant gets the lookaheads. Every other variant gets the normal drop downs. Hopefully not too difficult.

media/js/gloss_detail_edit.js
```diff
@@ -0,0 +1,1080 @@
...
+         }
+         if (this_id.startsWith("othermedia")) {
+             $(this).attr('disabled', true);
+         }
+         if (this_id.endsWith("_multiselect")) {
+             $(this).attr('disabled', true);
+         }
+         if (this_id.startsWith("nmevideo")) {
+            $(this).attr('disabled', true);
+         }
```
vanlummelhuizen: This could be shorter/DRY-er. See https://stackoverflow.com/questions/63881200/how-to-use-endswith-with-multiple-values for a nice way to do so.

susanodd:
> @susanodd You solved it!
> 
> I hate to say it, but I think I found a new problem: only the first Phonological Variant gets the lookaheads. Every other variant gets the normal drop downs. Hopefully not too difficult.

Hahaha.

Actually, I implemented it as such, only the primary gloss (first column) gets the lookaheads.

When created, the "variations" get initialised to the same phonology as the primary gloss.

Shall I do the lookaheads for all of the columns? (It makes a lot of extra code)

vanlummelhuizen:
Why would it make a lot of extra code? I would think it would be possible to just say something like "apply this lookahead thing to every variant instead of just the first", e.g. a simple for loop going over all variant.

media/js/gloss_detail_edit.js
```diff
@@ -0,0 +1,1080 @@
...
+         if (this_id.startsWith("othermedia")) {
+             $(this).removeAttr('disabled');
+         }
+         if (this_id.endsWith("_multiselect")) {
+             $(this).removeAttr('disabled');
+         }
+         if (this_id.startsWith("nmevideo")) {
+            $(this).removeAttr('disabled');
+         }
```
vanlummelhuizen: This could be shorter/DRY-er. See https://stackoverflow.com/questions/63881200/how-to-use-endswith-with-multiple-values for a nice way to do so.

media/js/gloss_detail_edit.js
```diff
@@ -0,0 +1,1080 @@
...
+    $('.empty_row_'+category).hide();
+    $('.button-'+category+'-to-appear-in-edit-mode').hide();
+    $('#enable_edit_'+category).removeClass('edit_enabled').show();
+    disable_editing(category);
+    busy_editing = false;
+}
+
+function toggle_edit_panel(category) {
+    if ($('#enable_edit_'+category).hasClass('edit_enabled'))
+    {
```
vanlummelhuizen: Move curly bracket to end of previous line for consistent coding style.

media/js/gloss_detail_edit.js
```diff
@@ -0,0 +1,1080 @@
...
+        $('.read_only_publication').hide();
+        $('.edit_only_publication').show();
+    }
+    $('.empty_row_'+category).show();
+    $('#enable_edit_'+category).addClass('edit_enabled').hide();
+    $('.button-'+category+'-to-appear-in-edit-mode').show();
+    enable_editing(category);
+}
+
+function hide_edit_panel(category) {
```
vanlummelhuizen, line 374: This function looks very similar to `show_edit_panel`; most things are toggled (hide or show). The functions could be merged if you use jQuery's `.toggle(bool)`, see https://api.jquery.com/toggle/#toggle-display.

media/js/lookaheads.js
vanlummelhuizen, line 1: Is it correct that `lookahead.js` is not used? If so please delete.

media/js/typeahead-bundle-0-11-1.js
vanlummelhuizen, line 1: Is it correct this file is not used? If so, please delete?

susanodd:
> Why would it make a lot of extra code? I would think it would be possible to just say something like "apply this lookahead thing to every variant instead of just the first", e.g. a simple for loop going over all variant.

No, there are conditionals in the template for each because the "init" form is different and there is extra stuff in the form for the fields. (The lookaheads have a typeahead and the normal ones are selects. It just adds more code.)

It's straightforward to add it.  (Just don't complain that the pull request gets longer) :)

I added the code to the template, that was simple copy paste and modify the element ids.

Now the typeahead setup needs to be modified because it has to be attached to each of the fields in each column inside the config code.
(The page is reloaded whenever the user switches between lists and lookaheads, as well as when a new variation is created or deleted. As well as when committed.)

vanlummelhuizen:
Then leave it for now. Perhaps after a first review we can decide if this is still desirable.

signbank/dictionary/templates/dictionary/phonological_variations.html
vanlummelhuizen, line 1: I cannot find a reference to this template. Please delete if it is not used.

media/js/gloss_detail_edit.js
```diff
@@ -0,0 +1,1080 @@
...
+
+function selectField(field, machine_value, value) {
+    if (console_log) { console.log('selectField: '+field+' '+machine_value+' '+value+' busy editing '+busy_editing); }
+    if (field === 'dialect') {
+        selected_dialect.push({'name': value, 'machine_value': machine_value})
+    } else if (field === 'semField') {
+        selected_semField.push({'name': value, 'machine_value': machine_value});
+    } else if (field === 'derivHist') {
+        selected_derivHist.push({'name': value, 'machine_value': machine_value});
+    }
```
vanlummelhuizen, line 105: You could do this more dynamic/generic with an array of field names, this: https://www.w3schools.com/Jsref/jsref_includes_array.asp and  `var selected_fields = window["selected_"+field];`.

media/js/gloss_detail_edit.js
```diff
@@ -0,0 +1,1080 @@
...
+        selected_semField = selected_semField.filter(item => {
+            return !(item['machine_value'] === machine_value);
+        });
+        if (console_log) { console.log('unselect '+JSON.stringify(selected_semField)); }
+    } else if (field === 'derivHist') {
+        selected_derivHist = selected_derivHist.filter(item => {
+            return !(item['machine_value'] === machine_value);
+        });
+        if (console_log) { console.log('unselect '+JSON.stringify(selected_derivHist)); }
+    }
```
vanlummelhuizen, line 122: See previous comment.

media/js/gloss_detail_edit.js
```diff
@@ -0,0 +1,1080 @@
...
+    busy_editing = true;
+    renderMultiSelected(field);  // stay in edit mode
+}
+
+// dynamically sets up the editable buttons in the left column during edit mode
+function renderMultiSelected(field) {
+    if (console_log) { console.log('call to renderMultiSelecte: '+field+' busy is '+busy_editing); }
```
vanlummelhuizen: There a lot of these lines. Perhaps it helps readability if you abstract the if-clause and call to 'console.log' away in a small separate function.

media/js/gloss_detail_edit.js
```diff
@@ -0,0 +1,1080 @@
...
+    var selected_fields = window["selected_"+field];  // read-only in this function
+    if (console_log) { console.log('call to render multiselected '+field+' '+JSON.stringify(selected_fields)); }
+    var container = $('#multiselect_value_'+field);
+    container.empty();
+    var values_input = $('#'+field+'_hidden_input_values');
+    values_input.empty();
+    var placeholder_lookahead = $('#'+field+'_multiselect');
```
vanlummelhuizen, line 137: Move this line closer to where it is actually used.

media/js/gloss_detail_edit.js
```diff
@@ -0,0 +1,1080 @@
...
+            $(this).parent().trigger('removeClicked');
+        });
+        tag.append(removeBtn);
+        tag.append('</button>').on('removeClicked', function() {
+            if (console_log) { console.log('button received event removeClicked'); }
+            unselectField(field, item.machine_value);  // destructive
+        });
+        container.append(tag);
+        values_input.append(input_value);
+    });
```
vanlummelhuizen, line 152: Add blank lines before and after for readability.

media/js/gloss_detail_edit.js
```diff
@@ -0,0 +1,1080 @@
...
+            source: bloodhound.ttAdapter(),
+            autoSelect: false,
+            templates: {
+                suggestion: function(fc) {
+                    return `<p><strong>${fc[config.displayKey]}</strong></p>`;
+                }
+            }
+        });
+    };
+}
```
vanlummelhuizen: This chunk is very similar to lines 31-71. If possible, merge the two.

media/js/gloss_detail_edit.js
```diff
@@ -0,0 +1,1080 @@
...
+
+const multiselectConfig = [
+    { name: 'dialect', element: '#dialect_multiselect', lookup: '.dialecttypeahead', selected_fields: 'selected_dialect' },
+    { name: 'semField', element: '#semField_multiselect', lookup: '.semFieldtypeahead', selected_fields: 'selected_semField' },
+    { name: 'derivHist', element: '#derivHist_multiselect', lookup: '.derivHisttypeahead', selected_fields: 'selected_derivHist' }
+]
+
+function readyMultiselect(config) {
+    let selected_fields = window[config.selected_fields];
```
vanlummelhuizen: This variable seems unused. Please remove.

susanodd:
Okay, the lookaheads have been added for the extra variations columns.

This is not on the server yet.

Some of the formatting seems to have a different background colour. (My screen is too small to check the dynamic html / css inheritance)

There is some javascript / jQuery code that attempts to dynamically adjust the width of the phonology fields with "extra" stuff. I'm not sure whether this actually works or not. It sometimes seems like it's working, then not working. It may not be in a location in the code where it is invoked at the correct moment. This code may also be trying to modify how those cells are displayed. This branch has been under development for months.

susanodd:
> Then leave it for now. Perhaps after a first review we can decide if this is still desirable.

Already done.

media/js/gloss_detail_edit.js
```diff
@@ -0,0 +1,1080 @@
...
+$('[class^="select_"]').each(function() {
+    $(this).on('change', function() {
+        busy_editing = true;
+    });
+});
+
+$('[class^="text_"]').each(function() {
+    $(this).on('change', function() {
+        busy_editing = true;
+    });
```
vanlummelhuizen: This can be done in one go with jQuery multi-selector: https://api.jquery.com/multiple-selector/
i.e.`$('[class^="multiselect_"], [class^="select_"], [class^="text_"]')`

media/js/gloss_detail_edit.js
```diff
@@ -0,0 +1,1080 @@
...
+            });
+        }
+        busy_editing = false;
+        if (['general', 'phonology', 'semantics', 'publication'].includes(panel_category)) {
+            hide_edit_panel(panel_category);
+        }
+    });
+
+     $('.quick_save_publication').click(function(e)
+	 {
```
vanlummelhuizen: Move curly bracket to end of previous line.

media/js/gloss_detail_edit.js
```diff
@@ -0,0 +1,1080 @@
...
+         for (var i=0; i < gloss_fields.length; i++) {
+            var field = gloss_fields[i];
+            if (['inWeb', 'isNew', 'excludeFromEcv'].includes(field)) {
+                var field_lookup = '#'+field+'_select_value';
+                var field_key = $(field_lookup).attr("name");
+                var field_value = $(field_lookup).val();
+                update[field_key] = field_value;
+            }
+         }
+         alert('update: '+JSON.stringify(update));
```
vanlummelhuizen: Seems here for the purpose of debugging. Delete please.

media/js/phonology_edit.js
vanlummelhuizen, line 1: Why is this in a separate file and not in `gloss-detail_edit.js`?

signbank/dictionary/templates/dictionary/gloss_detail.html
```diff
@@ -621,6 +565,317 @@
...
+input[type="text"] {
+    background-color: inherit;
+    width: inherit !important;
+}
+input[type="text"]:focus {
+    width: inherit !important;
+    color: red;
+}
+.tt-input {
+    background-color: rgba(50,50,50,15%) !important;
```
vanlummelhuizen: The alpha (a) in rgba should be decimal value not a percentage (https://www.w3schools.com/Css/css_colors_rgb.asp).

signbank/dictionary/templates/dictionary/gloss_detail.html
```diff
@@ -621,6 +565,317 @@
   outline: 2px solid red;
 }
 </style>
+<style>
```
vanlummelhuizen: Very big <style> element. Consider a separate CSS file.

signbank/dictionary/templates/dictionary/gloss_detail.html
```diff
@@ -621,6 +565,317 @@
...
+    opacity: 1;
+}
+.editform-semantics input:disabled {
+    color: inherit !important;
+    appearance: none;
+    -webkit-appearance: none;
+    -moz-appearance: none;
+    border: none;
+    opacity: 1;
+    outline: none;
```
vanlummelhuizen: This code is very similar to or even exactly the same as other pieces of code. Please make this whole CSS section DRY-er.

signbank/dictionary/templates/dictionary/gloss_detail.html
```diff
@@ -2386,236 +2931,1450 @@ <h2 id='modalTitleBlend'>{% trans "Delete This Blend" %}</h2>
...
+                </td>
+                {% endwith %}
+                {% endwith %}
+                {% endwith %}
+
+                 {% for gv in gloss_variations %}
+                 {% with variation_forms|get_item:gv.id as variant_form %}
+                 {% with variant_form.fields|get_item:'handedness' as handedness_field %}
+                 {% with variant_form.fields|get_item:'weakdrop' as weakdrop_field %}
+                 {% with variant_form.fields|get_item:'weakprop' as weakprop_field %}
```
vanlummelhuizen, line 2633: You could combine these assignments in one `with`, see https://docs.djangoproject.com/en/6.0/ref/templates/builtins/#with.

signbank/dictionary/templates/dictionary/gloss_detail.html
```diff
@@ -2386,236 +2931,1450 @@ <h2 id='modalTitleBlend'>{% trans "Delete This Blend" %}</h2>
...
+                    </select>
+                    </div>
+                </td>
+                {% endwith %}
+                {% endwith %}
+
+                 {% for gv in gloss_variations %}
+                 {% with variation_forms|get_item:gv.id as variant_form %}
+                 {% with variant_form.fields|get_item:'domhndsh' as domhndsh_field %}
+                 {% with variant_form.fields|get_item:'domhndsh_letter_or_number' as domhndsh_letter_or_number_field %}
```
vanlummelhuizen, line 2743: Combine the `with` assignments.

signbank/dictionary/templates/dictionary/gloss_detail.html
```diff
@@ -2386,236 +2931,1450 @@ <h2 id='modalTitleBlend'>{% trans "Delete This Blend" %}</h2>
...
+            <tr class="empty_row_phonology">
+            {% else %}
+            <tr>
+            {% endif %}
+            {% endwith %}
+                <th style="width:300px;">{% trans "Handedness" %}</th>
+                {% with phonologyform.fields|get_item:'handedness' as handedness_field %}
+                {% with phonologyform.fields|get_item:'weakdrop' as weakdrop_field %}
+                {% with phonologyform.fields|get_item:'weakprop' as weakprop_field %}
+                <td class="editform-phonology" style="width:150px !important;">
```
vanlummelhuizen: In the Phonology block, from about this point, there are many pieces of code that are copied and have only minor changes, see also your latest commit https://github.com/Signbank/Global-signbank/commit/f29ac6b0864f01564e54e42da2b67b27c95d3d27. You have got to find a more generic/DRY-er way to do this. E.g. with `{% include ... %}', see https://docs.djangoproject.com/en/6.0/ref/templates/builtins/#include.

signbank/dictionary/display_functions.py
```diff
@@ -0,0 +1,84 @@
...
+            show_field[field] = [semantic_fields.count() > 0]
+            continue
+        if field in ['derivHist']:
+            derivation_history_fields = gloss.derivHist.all()
+            show_field[field] = [derivation_history_fields.count() > 0]
+            continue
+        if field in ['dialect']:
+            dialect_fields = gloss.dialect.all()
+            show_field[field] = [dialect_fields.count() > 0]
+            continue
```
vanlummelhuizen: This could be one `if` clause when using `getattr(gloss, field)`.

signbank/dictionary/display_functions.py
```diff
@@ -0,0 +1,84 @@
...
+        if field in ['domhndsh_letter_or_number']:
+            letter_value = getattr(gloss, 'domhndsh_letter')
+            number_value = getattr(gloss, 'domhndsh_number')
+            show_field[field] = [not letter_value and not number_value]
+            continue
+        if field in ['subhndsh_letter_or_number']:
+            letter_value = getattr(gloss, 'subhndsh_letter')
+            number_value = getattr(gloss, 'subhndsh_number')
+            show_field[field] = [not letter_value and not number_value]
+            continue
```
vanlummelhuizen: Could also be one if clause.

signbank/dictionary/display_functions.py
```diff
@@ -0,0 +1,84 @@
...
+            if field in ['domhndsh_letter_or_number']:
+                letter_value = getattr(gv, 'domhndsh_letter')
+                number_value = getattr(gv, 'domhndsh_number')
+                show_field[field].append(not letter_value and not number_value)
+                continue
+            if field in ['subhndsh_letter_or_number']:
+                letter_value = getattr(gv, 'subhndsh_letter')
+                number_value = getattr(gv, 'subhndsh_number')
+                show_field[field].append(not letter_value and not number_value)
+                continue
```
vanlummelhuizen: Could be one if clause.

signbank/dictionary/forms.py
```diff
@@ -1785,6 +2058,343 @@ def __init__(self, *args, **kwargs):
...
+                                                         choices=choicelist_queryset_to_translated_dict(
+                                                                 list(FieldChoice.objects.filter(field='WordClass').order_by(
+                                                                     'machine_value')),
+                                                         ordered=False, id_prefix='', shortlist=False
+                                                         ),
+                                                         widget=forms.Select(attrs=ATTRS_FOR_FORMS))
+            self.fields['wordClass'].initial = self.gloss.wordClass.machine_value if self.gloss.wordClass else 0
+
+
+class PhonologyForm(forms.Form):
```
vanlummelhuizen, line 1866: This class is very similar to `PhonologicalVariationUpdateForm`. Please make it DRY-er.

vanlummelhuizen (changes requested):
@susanodd This is a huge PR. You obviously worked hard on it, so thank you for that.

There are a few general remarks. 

In the future we have make sure PRs are not growing this big. It is very hard to cope with for a reviewer but also for the Github PR interface. Typing this is very slow and I hear the fan of my computer trying to cool it.

Related to the size of this PR: there quite a few places where you copied code, only to make minor changes. Please try to avoid repeated code. If you would like a simple guideline:

1. If you copy more than 3 lines, make it more generic.
2. If you copy something more than once, make it more generic.

It is nice if you do this while coding, but you could also make it work first and refactor just before committing.

This last thing is that there are files that seem to not be in use. These add quite a number of lines to the total.

signbank/dictionary/display_functions.py
susanodd: We probably don't need this since it's on the same row as `domhndsh` and `subhndsh`

It's not clean to just remove these, since the PHONOLOGY is iterated over and they end up missing then.

Or we need to remove stuff dynamically from the "setting" that is used. (That looks bad.)

signbank/dictionary/display_functions.py
susanodd: I don't trust combining things in this case because the types/classes of the objects are totally different.

It ends up a "fake" polymorphic function. (Like those old C macro substitutions to reduce code.)

I will think about it.

signbank/dictionary/templates/dictionary/gloss_detail.html
susanodd: The "include" does not work if there are translations. ALL of the field choices use translations.

<details>
It's true, I was surprised the modification to add the lookaheads to the other columns was quite simple.

I kept distinct "ids" for the normal select versus the "lookahead" because it's impossible to debug code if the ids are reused. PyCharm complains. The original Gloss Edit had an extreme amount of code being "shared" between cases. It was murder trying to rewrite that code. (It was perhaps "concise" but unreadable. I moved all the "empty" row to a separate method. That was previously embedded all of the template.

I see you are complaining, but the code is much more readable that the original code.

media/js/lookaheads.js
susanodd: I renamed it. I will remove it now. I kept it in case you guys would finish review the other branch so I could modify that branch and pull it to this one.

media/js/gloss_detail_edit.js
susanodd: @Woseseltops helped on the original version.

It's true there is an extreme amount of toggling in order to "make pretty" versus "allow editing".
It's basically totally stupid CSS. I hate CSS. I did all the classes by hand. There are some patterns now.
I had to make sure the panels were "separate" from each other.
In the original Gloss Edit, ALL the stuff went into edit mode there was only one button.

There are slight differences based on what is in the panels. I will check this again.
The original version had the "python update" method attached to the "gloss object fields". They are been split into the panels now. Only the phonology has a different save method.

Those toggling things are doing a lot of contextual stuff. But the context has been changing during development.

I will check this.

signbank/dictionary/forms.py
susanodd: I already removed several forms from what it was originally. I was trying to keep all the class names different in different branches to allow merging.

The only difference is in the "object". I guess I can just call it "object". :)

media/js/gloss_detail_edit.js
susanodd: Oh, yes, that one kept getting longer and longer. (Trial and error.) It's a mess.

They basically all reply on the "form control" class being there in combination with others. It was not immediately obvious during development just "what" was to be disabled/enabled. The html things are all different.

media/js/typeahead-bundle-0-11-1.js
susanodd: It's not used in this branch. The other templates that have typeaheads should be modified to use the newer version.

media/js/gloss_detail_edit.js
susanodd: I can probably remove them. I got tired of adding all the console logs and then removing them

media/js/gloss_detail_edit.js
susanodd: This is the chunk that should be used in the "other" templates that use there. I purposely made it be separate so it can be copied to the other templates. (Or put in a file.) But that is not this PR.

media/js/gloss_detail_edit.js
susanodd: Hmmm.

media/js/phonology_edit.js
susanodd: Oh, it can be in the same as the other. That was the "variations" part originally. Then @Woseseltops wanted a single "save" for all the phonology, not per column. (that was the major rewrite the past week.)

(I was trying to keep the lookaheads code separate from the other branch)

signbank/dictionary/templates/dictionary/gloss_detail.html
susanodd: I was having lots of trouble with stuff "not being loaded" yet by Django. The css stuff got out of control. (But the original Gloss Edit template was already out of control.)
There are some weird things going on in some code because Jetske also put inline formatting css in the Senses area. There is also special formatting for video recording. Then there are special css that are imported. But then the nesting structure has been changed because of the "panel" edit buttons.
(Some assumptions about the overall code structure is not longer true.)

I will try to move the CSS.

I kept looking at the browser debugger area to see "what" css was being used. Sometimes it was doing really weird stuff. It would get all the way "down" to "body" elements. I think that because the dynamic stuff is sometimes at the end of the file, there isn't any "surrounding" element for it, so it can't inherit the css.

signbank/dictionary/display_functions.py
susanodd: Originally I had these fields in new rows. Now I'm wondering if this is not part of the Hand field..

Except the user could modify these but do nothing with the strong hand weak hand. Then if those are empty there could be stuff in these fields. (This seems kind of like a bug for the model?)

signbank/dictionary/display_functions.py
susanodd: I'm wondering if just the field actually retrieves all the objects. I'll see what it retrieves....

No, it does not work. Although the retrieved value is None, it does not work to merely compare it to "None", when it's empty, it's not None.

susanodd:
Well, something broke when I made some of the suggested changes. It's now stuck in edit mode.

media/js/gloss_detail_edit.js
susanodd: I had lots of trouble with these global state variables. It did not work unless they are explicit like that.

It works as read-only to combine, but not when they need to be updated. The function calling was "updating" them and then they would end up "not updated". (It has to have side effects on them.)

They are used inside of "stored unevaluated functions". (Forgot the technical same at the moment) Oh, closures.

media/js/gloss_detail_edit.js
susanodd: Hmmm. There was some weird stuff going on with those "global state variables". 

They need to be updated in a specific order here because of side effects.

media/js/gloss_detail_edit.js
susanodd: Okay, cool. I think something goes wrong with the syntax when I attempt to do these things.

media/js/gloss_detail_edit.js
susanodd: Oh, I see I was probably thinking about moving the arguments into the button using quicl-save.
It looks like I just stopped working on it. The alert is really stupid.

(This has been resolved, including next comment about alert.)

signbank/dictionary/templates/dictionary/gloss_detail.html
susanodd: Good catch! I just searched, there are many of these in other templates.....

There's even one in main.css (!!)

What to do, what to do. I can't change a bunch more files in this issue....

susanodd:
signbank-test sync'd up to here

Next up making one form instead of two.

signbank/dictionary/forms.py
susanodd: Okay, it's working now.

I had modified the "model" of the form to be Phonology.

And the object is now just object

susanodd:
`signbank-test` is NOT sync'd yet to reflect newest changes

susanodd:
Okay, the biggest problem that remains is the CSS.

I hate CSS coding. I find it difficult to use the syntax to abstract patterns. All the different panels have hard-coded specific CSS. I could not tell anymore what is superfluous. Since they are evaluated top down, and the patterns seem to overlap with each other. I have been watching the "browser" css to see what it's "cancelling out". But sometimes it seems there are missing "containing" elements.

Recall that the original Gloss Edit template was one gigantic edit block with editable on every cell.

I abused the "form-control" class for identifying the "single column" "disable" strategy for switching between edit and view modes. The actual "things that get edited" share the space and are the same things that are displayed. The original "editable" had separate elements / attributes for this that's why there were "OK" buttons everywhere.

Ideas on how to fix this are welcome.

I fear that if the "duplicated" form fields code is "condensed" that the CSS will totally choke. I have the free version of PyCharm and it does not like any "django" "conditional" element ids. It can't parse these.
I like having all the form elements explicitly written out. It makes is obvious what the structure is.

signbank/dictionary/templates/dictionary/gloss_detail.html
susanodd: It's been moved to a separate gigantic CSS file.

I have trouble parsing it myself.

media/js/gloss_detail_edit.js
susanodd: I left the third one separate because it also has a "reset" event.

(It's either three things as one, plus an extra one, or two things as one, plus an extra with two things.)

susanodd:
signbank-test is sync'd to here

media/js/gloss_detail_edit.js
susanodd: See previous answer.

media/js/gloss_detail_edit.js
susanodd: I undid that yesterday because it was having trouble with the multiselect.

signbank/dictionary/templates/dictionary/gloss_detail.html
susanodd: I did not succeed to remove stuff that seemed "unnecessary". It stopped formatting. (Could this be browser specific how it actually processes CSS ?)

signbank/dictionary/templates/dictionary/gloss_detail.html
susanodd: I guess I don't entirely trust the software. These are in fact already shortcuts. (The example with the patterns in the javascript for that long conditional was a huge improvement. I'm not sure this actually makes the code shorter, just too wide for the screen.)

signbank/dictionary/display_functions.py
Woseseltops: Not sure if I understand your reply @susanodd , but this could be how the suggestion by Micha works:

```python
for prefix in ['domhndsh', 'subhndsh']:
    if field == prefix + '_letter_or_number':
        letter_value = getattr(gv, prefix + '_letter')
        number_value = getattr(gv, prefix + '_number')
        show_field[field].append(not letter_value and not number_value)
        break
else:
    continue
````

File: signbank/dictionary/display_functions.py (hunk 22 above)
Comment (Woseseltops) [reply signbank/dictionary/display_functions.py]: Not sure if I understand your reply @susanodd , but this could be how the suggestion by Micha works:

```python
for prefix in ['domhndsh', 'subhndsh']:
    if field == prefix + '_letter_or_number':
        letter_value = getattr(gloss, prefix + '_letter')
        number_value = getattr(gloss, prefix + '_number')
        show_field[field] = [not letter_value and not number_value]
        break
else:
    continue
```

signbank/dictionary/templates/dictionary/gloss_detail.html
Woseseltops: Do not view the comments as complaints @susanodd . They're suggestion so we can collaborate on making our own code base more easy to read and extend :) . 

From what I read, I think it should be possible to make `{% include %}` work with translations, can you experiment with it a bit and attempt to make it work?

signbank/dictionary/templates/dictionary/gloss_detail.html
Woseseltops: So 

```python
{% with variation_forms|get_item:gv.id as variant_form %}
{% with handedness_field=variant_form.fields|get_item:'handedness' weakdrop_field=variant_form.fields|get_item:'weakdrop' weakprop_field=variant_form.fields|get_item:'weakprop' %}
```

media/js/gloss_detail_edit.js
Woseseltops: Example of what @vanlummelhuizen has in mind:

```javascript
const fieldArrays = {
    dialect: selected_dialect,
    semField: selected_semField,
    derivHist: selected_derivHist
};

if (fieldArrays[field]) {
    fieldArrays[field].push({'name': value, 'machine_value': machine_value});
}
```

media/js/gloss_detail_edit.js
Woseseltops:
```javascript
const selected = {
    dialect: selected_dialect,
    semField: selected_semField,
    derivHist: selected_derivHist
};

selected[field] = selected[field].filter(item => !(item['machine_value'] === machine_value));
if (console_log) { console.log('unselect ' + JSON.stringify(selected[field])); }
```

signbank/dictionary/display_functions.py
susanodd: This has just been removed. There are no rows for these fields, although they may be set.

Should the users be able to set these if the handshape isn't filled in? (Model allows this.)

signbank/dictionary/templates/dictionary/gloss_detail.html
susanodd: I already experimented with "include" in the past (several years ago) because I wanted the "query parameters" to be loaded since that is also quite long. That did not work at all.

The "documentation" has you doing the "with" assignments as part of the "include" for all the translations, so they get processed to "feed into" the included part. (It's obvious there are too many translations for that.)

The Signbank code that makes use of additional templates, like "gloss row" (called by ajax) are then "processed" and replaced into the template.
(The examples here have nothing to do with ajax calls. The code in question is "forms" not pre-processed data.)

I previously made an ISSUE for this to figure out how to employ the "include". This ought to be done for the 1000 lines of code that is Senses.

media/js/gloss_detail_edit.js
susanodd: I don't see that that gains anything.

The stackexchange stuff I could find related to this said that code was working on a copy of the global thing not the actual global thing.

signbank/dictionary/templates/dictionary/gloss_detail.html
susanodd: Hmpf. It just makes it really really long. (I'm working on a laptop. It's too hard for my eyes to focus on the same row scrolling so far along off the edge of the window. PyCharm actually starts line-breaking stuff if you make stuff so long....)

susanodd:
Global comment about the above, 

- PyCharm complains about really long lines in the editor
- E.g. the import statements have parentheses around them if they end up on multiple lines
- PyCharm runs the code in the Run Signbank tool.
- Regarding the suggestion to use "`{% include`":  it does not work to use "include" because of the translations and model translations inside the "included" file.
- the Javascript global state variables need to stay global state variables. They can't be aliased for "dry" code. 

<details>
Not using global variables explicitly did not work. Originally they were "dry". But with the introduction of "config" to reduce the amount of code, they became "aliases" and did not update the "global" state anymore. Using "const" is not implemented properly (apparently in the javascript interpreter) because the variables are not constants. They end up getting reset back to the original value.

susanodd:
You guys have written too many comments inside of the code.

It's not possible for me to view "just the code" anymore. It keeps clouding it up with all of your comments.

(My GitHub also does not show the files under the Project in the left column. This is probably a GitHub bug on Safari.)

media/js/gloss_detail_edit.js
vanlummelhuizen: @Woseseltops 's solution is also nice. It gains that there is single place that does the `push` while having a very readable list of what field may need pushing. What I actually meant was this even shorter solution:

```javascript
const fields = ['dialect', 'semField', 'derivHist'];
if(fields.includes(field)) {
    window['selected_'+field].push({'name': value, 'machine_value': machine_value});
}
```

Given the little experiment you can see in the screenshot below, I think this should work.

<details>
image
</details>

media/js/gloss_detail_edit.js
vanlummelhuizen:
```javascript
const fields = ['dialect', 'semField', 'derivHist'];  // USE THE ONE DEFINED ABOVE
if(fields.includes(field)) {
    var selected_field = window['selected_'+field];
    selected_field = selected_field.filter(item => !(item['machine_value'] === machine_value));
    if (console_log) { console.log('unselect '+JSON.stringify(selected_field)); }
}
```

media/js/gloss_detail_edit.js
vanlummelhuizen: > it can be copied to the other templates

That is not DRY and therefore not a good idea. 

> (Or put in a file.)

This sounds like a better idea. If there is a generic solution, like a small separate file containing this code, that would be the best thing to do. So, please try to end up with one generic version of this code.

susanodd:
`signbank-test` is sync'd to here

media/js/gloss_detail_edit.js
vanlummelhuizen: > I undid that yesterday because it was having trouble with the multiselect.

What problems exactly? I made the following changes and it seems to work fine:

```diff
diff --git a/media/js/gloss_detail_edit.js b/media/js/gloss_detail_edit.js
index a3b98417..9497ffad 100644
--- a/media/js/gloss_detail_edit.js
+++ b/media/js/gloss_detail_edit.js
@@ -548,22 +548,13 @@ function readyMultiselect(config) {
     });
 }
 
-$('[class^="multiselect_"]').each(function() {
-    $(this).on('change', function() {
-        busy_editing = true;
-    });
-});
-
-$('[class^="select_"]').each(function() {
+$('[class^="multiselect_"], [class^="select_"], [class^="text_"]').each(function() {
     $(this).on('change', function() {
         busy_editing = true;
     });
 });
 
 $('[class^="text_"]').each(function() {
-    $(this).on('change', function() {
-        busy_editing = true;
-    });
     $(this).on('reset', function() {
         var initial_value = $(this).attr('data-initial');
         $(this).val(initial_value);
```

signbank/dictionary/templates/dictionary/gloss_detail.html
vanlummelhuizen: > What to do, what to do. I can't change a bunch more files in this issue....

Make a separate issue.

signbank/dictionary/templates/dictionary/gloss_detail.html
vanlummelhuizen: I would have expected that you could make a multiline `with`. Apparently not. There is a hack to do so, but let's not go there. Perhaps we could solve this when building the context in the view, but that is something for another time.

signbank/dictionary/templates/dictionary/gloss_detail.html
vanlummelhuizen: Please point to a specific piece of code where translations don't work if a template include is used.

signbank/dictionary/display_functions.py
vanlummelhuizen: > We probably don't need this since it's on the same row as domhndsh and subhndsh

Then remove it, please.

signbank/dictionary/templates/dictionary/gloss_detail.html
susanodd: Yes, the "gigantic query parameters modal"

https://github.com/Signbank/Global-signbank/blob/0c3bf2311082307a5e51364615b814ec83bedde1/signbank/dictionary/templates/dictionary/gloss_detail.html#L650-L765

susanodd: I had to remove the "includes" for the tags and affiliations. Their code is now inside of this template.

(When "include" was used, the browser/Django render refused to put them in the correct place and "undid" the subsequent "nesting" of elements so that the jQuery could not locate the elements anymore.)

(master)
https://github.com/Signbank/Global-signbank/blob/0c3bf2311082307a5e51364615b814ec83bedde1/signbank/dictionary/templates/dictionary/gloss_detail.html#L1350-L1351

(this branch)

https://github.com/Signbank/Global-signbank/blob/78d981728ec60f0eb506e3fab0ecaa7d4d4bb40c/signbank/dictionary/templates/dictionary/gloss_detail.html#L1178-L1267

media/js/gloss_detail_edit.js
susanodd: This may have to do with that it is a list of dictionary.

I tried having a nice structure for it. But the actual "things" seemed to be copies of the "things".

The code must have (must change, needs to cause) a side effect on the global variable.

The actual "operations" occur during the "lookahead selection" inside of an event handler. It does not have the "context" when the event is created. (These are like closures.)
I had to remove that code from the event and move it to methods outside of the "config".

This all worked perfectly before there was a "config" setting it up. It used to be an elegant solution as you show. But the "config" broke this.
It also seems that trying to label those variables as "const" also ends up having them be reset, not actually updated.

(I can't debug the javascript interpreter)

media/js/gloss_detail_edit.js
susanodd: Okay, I prefer to put it in a file. Not combine it with the other fields.
(I added back the colours to the field choices, very simple. These don't have colours. I saw that they are wanting colours on the tags, then noticed there was a bug because of obsolete Django code and fixed it.)

media/js/gloss_detail_edit.js
susanodd: I have no idea. It just didn't find them anymore or something.

I can try again. I suspect maybe the entire code isn't being loaded properly by the browser. (This actually happens with ETC PHP stuff. The old code in the browser is sometimes impossible to get reloaded.)

media/js/gloss_detail_edit.js
susanodd: The problem is that "const". They are not constants.

I had console logs inside of all the events and the "updates" and "reads" of these variables, also when the "typeahead:select" events happened. I could see that the variables were being updated and that this was then undone.
Maybe because "all" of them are inside the same "mega" structure they all end up being reset or something.

(I can't debug the javascript interpreter)

Why is the original version with the (removed) console logs  being shown here?

media/js/gloss_detail_edit.js
susanodd: I see now why I did not combine them. There are two places where the gloss typeahead is attached to in the template. It's not clear how to do this in a config. Maybe it needs a list of element ids to attach itself. (It should be a separate "typeahead" instance for each appearance, correct?)

susanodd: This is not working. The typeahead (via a "config" in a different file) is not a function. Maybe there are too many of them being configured?

```
jQuery.Deferred exception: typeahead is not a function. (In 'typeahead($(config.lookup))', 'typeahead' is undefined) 
"readyLanguageLookahead@http://localhost:8000/static/js/language_typeaheads.js:63:14
```

Because there are multiple occurrences of gloss typeahead I have added to the "ready" config

```
const lookaheadLanguageConfig = [
    { name: 'lemma', target: '#new_lemma_pk', lookup: '.lemmatypeahead', displayKey: 'lemma' },
    { name: 'gloss', target: '#morpheme_target_id', lookup: '.glosstypeahead', displayKey: 'annotation_idgloss' },
    { name: 'gloss', target: '#morpheme_blend_id', lookup: '.glosstypeahead', displayKey: 'annotation_idgloss' },
    { name: 'gloss', target: '#relation_target_id', lookup: '.glosstypeahead', displayKey: 'annotation_idgloss' },
    { name: 'morph', target: '#morpheme_gloss_id', lookup: '.morphtypeahead', displayKey: 'annotation_idgloss' }
    ]
```

susanodd: Is it a problem because there are multiple occurrences? Without the "config" (to avoid duplication) it does not have any problems. But there are multiple occurrences of the gloss typeahead in the code.
Do these need "artificial" distinct names to force them to be "unique" when the "config" operates?

On one of the configs, I used the element id of the typeahead bloodhound input element. I guess that was significant.

I guess it's having trouble finding it in "window" from the "name". (But how many of these things are there? Are there multiple bloodhounds for gloss lookahead?) I'm having trouble figuring out how many "things" there should be.

susanodd:
I made a new issue for this #1763 

[BELOW OUTLINES WHAT WENT WRONG]
<details>

See above. Putting the "language typeaheads" in a separate file is causing problems because there are multiple occurrences of "gloss typeahead" in different fields.

This is delaying the completion of this issue and its merging.

> 6. Avoiding common pitfalls
> • Global pollution – Avoid attaching too many properties to window.
> • let/const do NOT go on window – Only var and function declarations do.
> • Always check existence before calling:


> Yes, you can attach one typeahead/autocomplete configuration to multiple input fields, but you must initialize the plugin separately for each element. This is because most autocomplete/typeahead plugins bind internally to a single element at a time.


Yes, the problem is that there are multiple gloss typeaheads. It fails on the second one.

I figured out how to solve it because I actually already solved this for the "variation columns" and their lookaheads.

Except for the solution, it will make the "javascript file" less generic because the config needs to know about occurrences of the gloss typeahead by their ids.


This isn't working. Also not with help from CoPilot.

The template code seems to be too recursive if trying to do this inside a "config".
The "window" isn't finding the right elements. The original code makes use of the typeahead (typed) class.

I'm supposed to be working on the videos. Now this has cost me half a day trying to put this code in a separate javascript file. This should be a separate issue. The other templates that use these exact same lookaheads may or may not need to also be edited to use shared code. All of the templates that use these lookaheads have their own explicit code. Some of this code is 10 years old.

media/js/gloss_detail_edit.js
susanodd: I made a new issue for this #1763

media/js/gloss_detail_edit.js
susanodd: This has been put back to the shorter version.

susanodd:
`signbank-test` is sync'd to here

susanodd:
Something weird happened to the Morpheme search panel of Gloss List View. (In spite of that that template was not edited.)

It would seem that the various templates that use the "original" typeahead are somehow finding the "new" typeahead instead (in the browser).

susanodd:
This branch will need to include branch #1764 because of issues with other templates that also use typeaheads. (The javascript is interfering with each other since there is still the static file for typeahead. The browser has problems with this for the other templates that use that file.)

Probably just continue on the new branch with the language typeaheads in the separate file.
I'm ending up revising ALL of the other templates that use lookaheads. There are some problems with using the "static" "old" typeahead javascript and the "new" from cloudfare. The browser seems to want to use the "new" one. This caused the other templates to break.

One template left to go. Then I'll (try to) pull that code into this branch. And into the variations videos branch.

So much for keeping the pull requests small. The variants are a major changed.

The editables had to be removed because they were taking up too much width in the phonology.
Then because it was a global Edit button, all of the panels had to be made independent of each other with separate save/dismiss buttons per panel and separate python update handlers, not the global update_gloss method with httpresponse (which had its own javascript to update the "editable" fields)
This could not be done in separate pull requests because everything is combined.
The old typeaheads were woven into the editable implementation (also in the javascript).

susanodd:
`signbank-test` is sync'd to here

susanodd:
`signbank-test` is sync'd to here

The typeheads for the language fields are now in a separate file that is shared by the other templates that use the language typeaheads. Those templates have been updated and templates that imported the typeahead javascript for no reason have it removed from their code.

There are three remaining "typeahead" templates (for user) see #1751  (Those are totally off topic for this issue)

susanodd:
`signbank-test` is sync'd to here

PR #1760, opened 2026-07-06
Phonologicalvariations videos

susanodd:
This branch builds on branch #1752 (issue #1726) which builds on branch #1724.

In an effort to complete this before the vacation, a simple implementation has been chosen

- simple add video removes existing videos and leaves just the new video (simple)
- the simple filename for the video is unique in the system 
--`<objectid>-<glossid>-<phonologicalvariationid>.<ext>`
--where `object` is a `PhonologicalVariationVideo`
- admin shows the variation videos in a new table (simple)
- variation videos upload is in the column of the phonology section (simple)
- video paths shown in Gloss Videos navigation panel


susanodd:
`signbank-susan` is sync'd up to here

susanodd:
`signbank-susan` is sync'd up to here

susanodd:
`signbank-susan` is sync'd up to here

susanodd:
`signbank-susan` is sync'd up to here

signbank/dictionary/models.py
```diff
@@ -4538,3 +4538,55 @@ class PhonologicalVariation(Phonology):
...
+        video.save()
+
+        # see if there is a file with the correct path that is not referred to by an object
+        relative_path = get_phonologicalvariation_video_file_path(video, str(videofile))
+
+        video.videofile.save(relative_path, videofile)
+
+        self.save()
+
+        return video
```
Woseseltops, line 4563: Minor thing: too much whitespace here :P

signbank/dictionary/update.py
```diff
@@ -347,6 +347,29 @@ def delete_phonological_variation(request, variationid):
...
+    elif 'videofile' in request.POST.keys():
+        # for unit tests, the file is found in the POST data
+        vfile = request.POST['videofile']
+    else:
+        error_message = gettext("A video file is required.")
+        # return JsonResponse({'error': error_message}, status=400)
+        messages.add_message(request, messages.ERROR, error_message)
+        return HttpResponseRedirect(reverse('dictionary:admin_gloss_view', kwargs={'pk': gloss.pk}))
+
+    variation.add_video(request.user, vfile)
```
Woseseltops, line 361:
```python
    variation = get_object_or_404(PhonologicalVariation, id=variationid)
    gloss = variation.gloss
    vfile = request.FILES.get('videofile') or request.POST.get('videofile')
    if not vfile:
        messages.add_message(request, messages.ERROR, gettext("A video file is required."))
        return HttpResponseRedirect(reverse('dictionary:admin_gloss_view', kwargs={'pk': gloss.pk}))

    variation.add_video(request.user, vfile)
```

Woseseltops (changes requested):
Functionally this is different from what I proposed, but I think better, so let's do your way :). I also like it code wise, well done. There's only one thing I'm unsure about, which is the new video file name format:
* The original plan was to only have the video pk there, but in the end we decided to also add the human readable gloss as a helper to quickly identify what file it is. I notice you are adding more information to the filename, though, which I think kind of defeats the purpose of moving to a new system. 
* Reviewing this code made me realize I forgot about the fact that not all videos are in the same table and thus will have overlapping pks. Perhaps there should be a code to indicate if something is a normal video (`v-`?), NME video (`nme-`?), perspective videos (`p-`?) and phonological variation videos (`pv-`?). So a full filename would be `pv-2344-BAL.mp4`. This does tell you it's a phonological variation video for the gloss BAL, but no longer which variation exactly
* Also good to keep in mind that we want to switch over all videos to this new system, so all code related to figuring out filenames would be better on a generic place

signbank/dictionary/update.py
```diff
@@ -347,6 +347,29 @@ def delete_phonological_variation(request, variationid):
...
+    if 'videofile' in request.FILES.keys():
+        vfile = request.FILES['videofile']
+    elif 'videofile' in request.POST.keys():
+        # for unit tests, the file is found in the POST data
+        vfile = request.POST['videofile']
+    else:
+        error_message = gettext("A video file is required.")
+        # return JsonResponse({'error': error_message}, status=400)
+        messages.add_message(request, messages.ERROR, error_message)
+        return HttpResponseRedirect(reverse('dictionary:admin_gloss_view', kwargs={'pk': gloss.pk}))
```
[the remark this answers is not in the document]
susanodd: Hmm. This string case was specific for the "tests" that this case is in the code.

signbank/dictionary/update.py
susanodd: Yeah, some of this code was borrowed from similar methods.

I just realised something about how I tend to make separate declarations for variables rather than embed them (as for the `error_message` in the original code. Once upon a time in compiler development, you tried to optimize things so that the values could be reused later, like only evaluate things once. I gather Django does some kind of library of all the "translated text".

susanodd:
> Functionally this is different from what I proposed, but I think better, so let's do your way :). I also like it code wise, well done. There's only one thing I'm unsure about, which is the new video file name format:
> 
> * The original plan was to only have the video pk there, but in the end we decided to also add the human readable gloss as a helper to quickly identify what file it is. I notice you are adding more information to the filename, though, which I think kind of defeats the purpose of moving to a new system.
> * Reviewing this code made me realize I forgot about the fact that not all videos are in the same table and thus will have overlapping pks. Perhaps there should be a code to indicate if something is a normal video (`v-`?), NME video (`nme-`?), perspective videos (`p-`?) and phonological variation videos (`pv-`?). So a full filename would be `pv-2344-BAL.mp4`. This does tell you it's a phonological variation video for the gloss BAL, but no longer which variation exactly
> * Also good to keep in mind that we want to switch over all videos to this new system, so all code related to figuring out filenames would be better on a generic place

About the naming convention, I wanted to avoid months of revising the video storage system AGAIN. (!!!!!)

I know it's a separate table. I did this on purpose. All of the other video classes have subtyping, so all of the methods need to work on all of the objects. This is an extreme headaches. We want to code to get finished, don't we?

To be in the same table, the videos need to be in the same (super) class. Then all the methods apply.
This is far too exhausting to achieve this at an acceptable completion time.

I put the variant pk in the video filename because it can't be the "number". Those are not static. If the user removes variants, they get renumbered. Then if you use the "number" in the filename, it may refer to a different object.

I think it's a bad idea to put them into the GlossVideo class with just a different filename. Of all the things that would need to change, the admin filters and queries would not work. The variants are not glosses, they need extra dereferencing to get the gloss from the object. But then in traversing, what do you do? Keep track of whether you got to the "gloss" field via the variant? (This seems susceptible to a lot of errors.)

susanodd:
`signbank-susan` is sync'd up to here

susanodd:
@Woseseltops 

> - Reviewing this code made me realize I forgot about the fact that not all videos are in the same table and thus will have overlapping pks. Perhaps there should be a code to indicate if something is a normal video (v-?), NME video (nme-?), perspective videos (p-?) and phonological variation videos (pv-?). So a full filename would be pv-2344-BAL.mp4. This does tell you it's a phonological variation video for the gloss BAL, but no longer which variation exactly

Why will these have overlapping pks? The created objects are unique and are in different classes. They should be referenced via Django via the object. The pk of the phonological variation object is also in the filename.

What classes actually reuse the pk's ? The new retrieval of the videos uses the pk of the video object. Are the subclasses also reusing these pk's?

Users should be accessing the videos  by way of the gloss or variation object, not by way of the file system.

Perhaps we adjust the optional "download" link shown for the files?

(Not related, but sort of) When you make screenshots on Apple in the most recent OS on the various devices, it puts the timestamp in the filename. (There is no other kind of information, it's just "screenshot" (in Dutch) with a timestamp.) Those filenames are totally useless to the user.

susanodd:
@Woseseltops 

> - Also good to keep in mind that we want to switch over all videos to this new system, so all code related to figuring out filenames would be better on a generic place

But a counter argument, all of the "new kinds of videos" are in a separate table. These can thus also be easily "moved" or "renamed" to a new system.

susanodd:
`signbank-susan` is sync'd up to here

signbank/dictionary/templates/dictionary/gloss_videos.html
```diff
@@ -138,6 +138,28 @@ <h3>{% trans "Citation Form Image" %}</h3>
...
+                <table class='table table-condensed table-condensed-light' style="margin-right:5px;margin-top:10px;margin-bottom:10px;">
+                    <tr>
+                        <td>#1</td>
+                        <td>{{gloss.get_video_url}}</td>
+                    </tr>
+                    {% for gv in gloss_variations %}
+                        {% if gv.has_video %}
+                        <tr>
+                            <td>#{{gv.variation}}</td>
+                            <td>{{gv.get_video_url}}</td>
```
vanlummelhuizen: In other parts of the code base `gv` means gloss_video. Here it means gloss_variation. It would be nice if both uses of `gv` are expanded to their respective full names to avoid confusion and improve readability.

signbank/dictionary/models.py
```diff
@@ -4538,3 +4538,51 @@ class PhonologicalVariation(Phonology):
...
+                                           get_phonologicalvariation_video_file_path)
+
+        if not isinstance(videofile, File):
+            msg = gettext("No video file supplied for video upload of variation video {variationid}.").format(variationid=self.pk)
+            raise ValidationError(msg)
+
+        # get existing video objects for this variation and delete them
+        existing_videos = PhonologicalVariationVideo.objects.filter(variation=self).order_by('pk')
+        for ev in existing_videos:
+            ev.delete()
```
vanlummelhuizen: Avoid very short variable names like `ev`. You could change it to `video` here.

signbank/dictionary/models.py
```diff
@@ -4538,3 +4538,51 @@ class PhonologicalVariation(Phonology):
...
+        video_path = str(existing_video.videofile)
+        filepath = os.path.join(settings.WRITABLE_FOLDER, video_path)
+
+        if not os.path.exists(filepath):
+            return ''
+
+        return video_path
+
+    def has_video(self):
+        return self.get_video() not in ['', None]
```
vanlummelhuizen: Use QuerySet.exists() (https://www.djangotricks.com/blog/2024/12/https-for-django-development-environment/) if you only need to know if there are videos or not.

signbank/video/admin.py
```diff
@@ -820,7 +820,53 @@ class ExampleVideoHistoryAdmin(admin.ModelAdmin):
...
+            return ""
+        video_file_full_path = os.path.join(WRITABLE_FOLDER, str(obj.videofile))
+
+        return video_file_full_path
+
+    def variation(self, obj=None):
+        """
+        column VIDEO FILE
+        this will display the full path in the list view, also for non-existent files
+        this allows to browse the file paths also on the development servers
```
vanlummelhuizen: Is this doc correct?

signbank/video/admin.py
```diff
@@ -820,7 +820,53 @@ class ExampleVideoHistoryAdmin(admin.ModelAdmin):
...
+        if obj is None:
+            return ""
+        return obj.variation.variation
+
+    def gloss(self, obj=None):
+        """
+        column VIDEO FILE
+        this will display the full path in the list view, also for non-existent files
+        this allows to browse the file paths also on the development servers
+        """
```
vanlummelhuizen: Is this doc correct?

signbank/video/models.py
```diff
@@ -1528,3 +1528,75 @@ def delete_files(sender, instance, **kwargs):
...
+
+
+def get_phonologicalvariation_video_file_path(instance, original_filename):
+
+    ext = extension_on_filename(original_filename)
+
+    idgloss = instance.variation.gloss.idgloss
+    video_folder = os.path.join(GLOSS_VIDEO_DIRECTORY,
+                                instance.variation.gloss.lemma.dataset.acronym,
+                                get_two_letter_dir(idgloss))
```
vanlummelhuizen:
```
    gloss = instance.variation.gloss
    video_folder = os.path.join(GLOSS_VIDEO_DIRECTORY, gloss.lemma.dataset.acronym, get_two_letter_dir(gloss.idgloss))
```

signbank/video/models.py
```diff
@@ -1528,3 +1528,75 @@ def delete_files(sender, instance, **kwargs):
         status = instance.delete_files()
     else:
         status = instance.delete_files()
+
+
+def get_phonologicalvariation_video_file_path(instance, original_filename):
```
vanlummelhuizen: When reading this function, it is nice to know what kind of object `instance` is. Apparently, it is a PhonologicalVariationVideo. Instead of `instance` you could name it `phonological_variation_video`. If you want is a bit shorter, perhaps `phon_variation_vid`.

signbank/video/models.py
```diff
@@ -1528,3 +1528,75 @@ def delete_files(sender, instance, **kwargs):
...
+    # video version, version = 0 is always the one that will be displayed
+    # we will increment the version (via reversion) if a new video is added
+    # for this gloss
+    version = models.IntegerField("Version", default=0, null=False)
+
+    def __init__(self, *args, **kwargs):
+        if 'upload_to' in kwargs:
+            self.upload_to = kwargs.pop('upload_to')
+        else:
+            self.upload_to = get_phonologicalvariation_video_file_path
```
vanlummelhuizen:
```
self.upload_to = kwargs.pop('upload_to', get_phonologicalvariation_video_file_path)
```

signbank/video/models.py
```diff
@@ -1528,3 +1528,75 @@ def delete_files(sender, instance, **kwargs):
...
+
+        if os.path.exists(self.videofile.path):
+            try:
+                os.remove(self.videofile.path)
+            except (OSError, PermissionError, IOError, KeyError, ValueError):
+                self.videofile.name = ''
+                return
+        else:
+            self.videofile.name = ''
+        return
```
vanlummelhuizen: You could use a guard clause here, i.e. `if not os.path.exists(...):` etc.

signbank/video/models.py
```diff
@@ -1528,3 +1528,75 @@ def delete_files(sender, instance, **kwargs):
...
+    :param instance:
+    :param kwargs:
+    :return:
+    """
+    if DEBUG_VIDEOS:
+        print('delete_files pre_delete: ', sender, str(instance))
+    if not instance.videofile or not instance.videofile.name:
+        return
+
+    instance.delete_files()
```
vanlummelhuizen: ❗ `PhonologicalVariationVideo` does not have a `delete_files` method.

vanlummelhuizen (changes requested):
My comments are mostly about minor things.

When testing the functionality, I got an error when deleting a variation (url: /dictionary/update/delete_phonological_variation/2/). Apparently, PhonologicalVariantionVideo.delete_files does not exist. See [my comment](https://github.com/Signbank/Global-signbank/pull/1760/changes#r3579680507)

Another thing: when a variation has a video, it is not shown when a user is editing. (S)He might thing the video disappeared. Just add some text, a link to the video, or change the button text to something like "Replace video" (or do more than one of these suggestions).

susanodd:
Yes, CoPilot had me change the name of delete_files. Looks like I missed one.

I'll show the video during editing. I thought it was confusing so I hid it. I'll make it visible again.

signbank/dictionary/models.py
susanodd: That's a good point. I think it was that way because the method used to be returning a path or something (the comparable methods, there are numerous methods for these things. like we just kept making more methods for some reason. good point, I will fix it.) (I made it be the same as the other similar methods, there are too many methods.)

This is checking for '' or None because the function it calls also checks whether the file exits.

signbank/video/models.py
susanodd: Yeah, that was CoPilot. I didn't push that to the server. I also missed one.

signbank/video/models.py
susanodd: Hmmm. I always thought instance was a reserved word like self. Oh, but this is a totally separate function. Daah

susanodd:
`signbank-susan` is sync'd up to here

PR #1764, opened 2026-07-07
Replace static typeahead javascript with cdnjs.cloudflare retrieved

susanodd:
This issue has been renamed for convenience. Github did not want to show me the new branch I created yesterday.

The uses of the STATIC typeahead have been replaced with the linked version on cdnjs.cloudflare.

- usertypeahead
- handshapetypeahead

used in

- admin_dataset_manager.html
- dataset_detail.html
- admin_handshape_list.html

Original issue about sharing the language typeaheads
<details>
Put the language typeaheads in a separate javascript file so they can be shared by multiple templates

This branch builds on #1752

<details>
Problems with applying the original config setup here

The "window[.... config...]" constructor has problems

`window.styleMedia is deprecated draft version of window.matchMedia API that is not implemented in Firefox and will be removed from the web platform in future.
`
</details>
Resolved

The code is working now. Each of the typeahead class names must be unique or it doesn't work.

I modified the repetitions of glosstypehead to have new class names. The other templates will need to be modified in order to share the typeahead code.

@vanlummelhuizen Question:

How to exclude "morpheme annotation texts" from "Lemma" ?

https://github.com/Signbank/Global-signbank/blob/0c3bf2311082307a5e51364615b814ec83bedde1/signbank/dictionary/adminviews.py#L5807-L5809

I implemented it, but perhaps it's a bit clumsy.

https://github.com/Signbank/Global-signbank/blob/573d59ca421a6df04b593ba8dcfb049f62b049f1/signbank/dictionary/adminviews.py#L5846-L5854

Btw, it was not clear that the gloss and lemma lookaheads were returning also text of morphemes. It's only visible now because the lookaheads have been modified to show everything if the user has not typed anything yet. For the test dataset, there are some "test" morphemes, so you see them there.

Another issue, it would seem in `morpheme_detail.html` allows the user to select an other "lemma" for the morpheme, but these lemma's are using the lemmatypeahead, which is linked to the ajax method that retrieves lemma's. Thus also lemma's that belong to glosses. (This is a bug actually.) (@vanlummelhuizen kindly check this.) (Maybe the user should not be selecting "an other" lemma here for a morpheme?


susanodd:
Hmmm. OMG, the code of `lemma_typeahead.js` uses `editable`. It's for the `add_gloss.html` template

Any suggestions in how to make use of the new "language typeaheads" javascript file?

The whole template needs surgery.

Actually, the "config" table needs to be adjusted and "unused rows" ignored for this other template, which only has a lemmatypeahead

susanodd:
One more to go, add new gloss

susanodd:
The changes in this branch have been merged into the #1760 and #1752 branches

susanodd:
This branch has been merged into the branches that use this code.

Issue #1755, opened 2026-07-02
Assign colors to labels/tags


Woseseltops:
Coming out of talks with Caro Brosens from VGT Signbank, add an extra model field 'color', to be displayed in the search results, not unlike how it worked in the old version of VGT Sigbank:

<img width="1222" height="793" alt="Image" src="https://github.com/user-attachments/assets/3d6f1112-7c95-4a74-b357-965acd957484" />

but now linked to labels/tags, and configurable.

susanodd:
This needs further clarification. The Tags are prefab from a library. The Affiliations have colours in the admin. I implemented these, derived from Tags. Is this what you mean?

@Woseseltops can you obtain some examples of colour X for attribute Y ?  What are the row colours in the example above indicating?

I found where these original coloured row tags were introduced (these were not set by the user):

https://github.com/Signbank/Global-signbank/commit/ff896d850232b04201beb722ba12cf28cc51f07f

susanodd:
I found where these coloured row tags were introduced:

https://github.com/Signbank/Global-signbank/commit/ff896d850232b04201beb722ba12cf28cc51f07f

PR #1794, opened 2026-09-11
Remove django tagging

### LINKED ISSUE #1775: Upgrade Python dependencies other than Django
This is a separate issue, make it a separte PR as well

Woseseltops:
A while ago @vanlummelhuizen and myself discovered that the `django-tagging` module is no longer supported. A more detailed investigation reveals that it's not that complicated to move all of the functionality we use into the Signbank repo as `signbank.tags`. This has the important advantage that we can modernize the code, and even extend it if we want (#1755) . 

I've set it up in such a way that all calls to `django-tagging` continue to work, which minimizes the number of code changes required. This does have the important downside that `django-tagging` and the new `signbank.tags` cannot be active at the same time (would result in name clashes), which is a problem for migrating tags from the old to the new system. I'm now using a temporary class to solve this.


signbank/tags/templatetags/tagging_tags.py
```diff
@@ -0,0 +1,10 @@
+from django import template
+from signbank.tags.models import Tag
+
+register = template.Library()
+
+
+@register.simple_tag
+def tags_for_object(obj):
+    """Get all tags for an object (replaces django-tagging template tag)"""
```
vanlummelhuizen, line 9: I would remove any reference to django-tagging in the code except for the migration.

signbank/tags/migrations/0002_migrate_from_tagging.py
```diff
@@ -0,0 +1,106 @@
+# Data migration to migrate from django-tagging to custom tags
```
vanlummelhuizen, line 1: Just a question, no change expected: could this migration be avoided if you would set the 'db_table' of the new models meta class to the existing table names from django-tagging? See https://docs.djangoproject.com/en/6.1/ref/models/options/#db-table

vanlummelhuizen (commented):
How are the tables from django-tagging removed from the database? Something like `bin/develop.py migrate tagging zero` does not work because `tagging` is already remove from INSTALLED_APPS.

signbank/tags/migrations/0002_migrate_from_tagging.py
vanlummelhuizen: More theoretic fantasy: What would happen if you would call the new app `tagging` and putting a copy of models.py from django-tagging in it and then just uninstall `django-tagging`?

vanlummelhuizen: I could not help myself and tried it out. The answer: nothing. Ergo, this is also possible. See 
[tagging.zip](https://github.com/user-attachments/files/32188028/tagging.zip)

Woseseltops: Ah great, that approach requires even fewer code changes and allows us to skip the whole ugly migration. Will take that one.

susanodd:
I'm a  bit confused if this branch has not been replaced by the version of @vanlummelhuizen ?

Woseseltops:
Okay I've implemented and tested the approached suggested by @vanlummelhuizen , and can confirm it works. Two notes:
* It took me a while to figure out the `tagging` folder should not live next to `dictionary` etc, but a level higher. I feel as this is part of our Signbank now (and not something generic), it should be a sibiling of the other Signbank modules, but this would require code changes after all. A matter of taste.
* This requires `django-tagging` to be uninstalled and will give errors otherwise, which `pip install -r requirements.txt` does not support (it only adds stuff)... perhaps we should rebuild the env as part of the deploy process, or switch to `pip-sync`?

susanodd:
The feedback is part of Signbank but lives above dictionary. Oh, my bad. I flipped the iPad horizontally and now I see it. I guess I’m the one who put something above Signbank, the Communication module with the texts for emails. No, that’s correct, they’re above dictionary, but under Signbank. Okay, I see what you mean. But media is above Signbank. The original tagging is so that you can add tags to arbitrary objects.

Are you sure pip install requirements doesn’t just reinstall everything? It removes the old version when a new is installed. Is that maybe a setting? Probably the ansible script can do this?

PR #1797, opened 2026-09-16
1775 upgrade python dependencies other than django

### LINKED ISSUE #1775: Upgrade Python dependencies other than Django

vanlummelhuizen:
Dependencies are upgraded or removed from `requirements.txt` if not directly used (if another dependency needs it, it is installed anyway).


requirements.txt
```diff
@@ -1,24 +1,18 @@
-Django==4.2.30
-django-bootstrap3==26.1
+Django==5.2
```
Woseseltops, line 1: I'm confused the Django upgrade is also here, as I thought this was about dependencies other than Django. Still, wouldn't upgrading to a new Django not entail various changes throughout the code base?

templates/global-templates/menu.html
```diff
@@ -116,7 +116,13 @@ <h1>Global <br/><span id="second-title-part">signbank</span></h1>
...
         {% if user.is_authenticated %}
             <div class='navbar-text navbar-text-light'>
-                <a id="username_link" href="{% url 'logout' %}">{% trans "Logout" %} ({{ user.first_name }} {{ user.last_name }})</a></div>
+                <form action="{% url 'logout' %}" method="post">
+                    {% csrf_token %}
+                    <button type="submit" name="your_name" value="your_value" class="btn-link" style="margin: 0px; border: 0px">
+                        {% trans "Logout" %} ({{ user.first_name }} {{ user.last_name }})
+                    </button>
+                </form>
+            </div>
```
Woseseltops, line 125: Why is this here? Does the new Django version require this?

Woseseltops (commented):
Given my growing hate for dependencies, really happy to see so many of them dropped :) .

PR #1800, opened 2026-09-17
Efficient trimmed down CSV-export when more than 100 results

### LINKED ISSUE #1796: Make exporting whole datasets to CSV more convenient
Coming out of a brainstorm with Galya Stern (UvA): atm exporting a whole dataset, like NGT 7k glosses, takes long and might even crash. There's probably something we can do to make the user experience smoother here. 

I think the first question is what exactly is taking so long:
* If the problem is in the SQL, perhaps an optimized query is possible
* If we are simply waiting for SQLite to return the data, perhaps #673 is enough
* If the problem is not the loading of the glosses, but the creation of the CSV, perhaps a separate parallel job which the user can return to later might work

If all of the above fails, we can track which datasets change during the day, and generate the CSV for these during the night? The dataset page would then have an 'download nightly CSV button'.

susanodd:
Help feedback shown to user should the exported lemma fields of the CSV, which have been stripped, differ from those stored in the database, should the user import the file again using import update glosses mode

When there are more than 100 results in the query, only the Signbank ID, Dataset, Annotations, Lemmas, and phonology fields are exported.

For the entire NGT dataset, this takes less than five minutes now. (Or thereabouts.)

on export, text fields are trimmed of white space. On importing the newly exported file as an “update”, the update command now reports discrepancies in the Lemma fields if the database has white spaces. There were 6 of these glosses in NGT.


signbank/dictionary/views.py
```diff
@@ -1147,10 +1147,24 @@ def import_csv_update(request):
...
+                help1 = gettext("Row {row}: Attempt to update Lemma translations for Signbank ID {glossid}.").format(row=str(nl+2), glossid=str(pk))
+                error.append(help1)
+                help2 = gettext("The stored lemma translations differ from those in the CSV: {lemmaidglosstranslations}.").format(lemmaidglosstranslations=current_lemmaidglosstranslations)
+                error.append(help2)
+                if differences_lemmas:
+                    help3 = gettext("There are white space characters at the start or end of the text in the database.")
+                else:
+                    help3 = gettext("Use Import CSV Lemma Update instead.")
+                error.append(help3)
                 continue
```
Woseseltops: I feel like all the conditions and nesting make this hard to read. How about something along these lines:

```python
def get_current_lemma_translations(lemma):
    """Return {language: text} for every translation language of the lemma's dataset."""
    languages = lemma.dataset.translation_languages.all()
    stored = {t.language_id: t.text for t in lemma.lemmaidglosstranslation_set.all()}
    return {language: stored.get(language.pk, '') for language in languages}


def differs_only_by_whitespace(csv_translations, current_translations):
    """True if some stored text equals the CSV text once its surrounding whitespace is stripped."""
    for language, csv_text in csv_translations.items():
        stored_text = current_translations.get(language)
        if stored_text is not None and stored_text != csv_text and stored_text.strip() == csv_text:
            return True
    return False

current_lemmaidglosstranslations = get_current_lemma_translations(gloss.lemma)

if lemmaidglosstranslations and lemmaidglosstranslations != current_lemmaidglosstranslations:
    if differs_only_by_whitespace(lemmaidglosstranslations, current_lemmaidglosstranslations):
        hint = gettext("There are white space characters at the start or end of the text in the database.")
    else:
        hint = gettext("Use Import CSV Lemma Update instead.")

    error.append(gettext("Row {row}: Attempt to update Lemma translations for Signbank ID {glossid}.")
                 .format(row=nl + 2, glossid=pk))
    error.append(gettext("The stored lemma translations differ from those in the CSV: {lemmaidglosstranslations}.")
                 .format(lemmaidglosstranslations=current_lemmaidglosstranslations))
    error.append(hint)
    continue
```

Woseseltops (commented):
Seem like a good solution to me. Only one request regarding readability of conditions.

